PR #6, opened 2021-08-19
DeferredOperationQueue: add RunContext-supplier for flush job

fschinkel:


org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/util/concurrent/DeferredOperationQueue.java
```diff
@@ -105,10 +113,18 @@ protected long getMaxDelayMillis() {
     return m_batchOperation;
   }
 
+  protected Supplier<RunContext> getRunContextSupplier() {
+    return m_runContextSupplier;
+  }
+
+  protected RunContext createRunContext() {
```
andibur: The method `getRunContextSupplier` is sufficient. I would inline this one.

PR #11, opened 2021-08-24
Use the rowHeight of the GroupBox for a FormFields status

fschinkel:
Do not only use the rowHeight of the HtmlEnvironment, but also the
rowHeight of the parent group box's bodyLayoutConfig if present.


cguglielmo (commented):
Why is this necessary? What about other composite fields like SequenceBox and RadioButtonGroup?
The title of the change seems to have some grammatical errors.

fschinkel:
The FormFieldLayout previously used the rowHeight of the HTMLEnvironment
for the height of the status div. However, if the FormField is in a
GroupBox whose rowHeight has been changed and thus no longer equals the
rowHeight of the HTMLEnvironment, the FormField and status no longer fit
together. Therefore use for the height of the status div also the
rowHeight of the GroupBox if available.

PR #12, opened 2021-08-24
Tile: fix async loading

fschinkel:
A tile should always load its data in the context of the form it is part
of. If there is no parent of type IForm use IForm.CURRENT.get() instead.


cguglielmo (commented):
Please describe, what problem this fix solves (in general, it is more important to describe why you have created a change than describing what you have done).

fschinkel:
For example, if a Tile dynamically adds FormFields depending on the
loaded data, the Form of these FormFields is always IForm.CURRENT.get()
evaluated in the current RunContext. If this RunContext is not
explicitly created with the Form in which the Tile is located, then
IForm.CURRENT.get() is not necessarily the Form of the Tile but possibly
a different one. This happens for example when the data is already
loaded while the Form is still being created and has not yet been set.
In this case, operations like addField() fail because the Form of the
Tile and the dynamically created FormFields do not match.
Therefore, the data must always be loaded in the context of the Form in
which the Tile is located. Use IForm.CURRENT.get() only if there is no
parent of type IForm.

PR #20, opened 2021-09-14
BrushUp: add scroll shadow

cguglielmo:
The scroll shadow visualizes whether there is scrollable content.
It is installed automatically when scrollbars are installed.
The shadow is a div that is placed after the scrollable.
The div has 4 box shadows, one for each side. Since it is placed
after the scrollable, it has to track the style, visibility and position
of the scrollable and adjust itself accordingly. The size and position
are mainly updated by the already existing scrollbars.update calls by
the layouts. Additionally, dom mutation and intersection observers are
used to track dom movements and visibility changes.

Some alternative solutions that did not work well:
- Using a box shadow inside the scrollable container: The box shadow
  lays under the elements, so if they have a background color the shadow
  won't be visible completely.
- Using position sticky: only works with relative positioning and does
  not work well for horizontal scrolling
- Using position absolute and update position while scrolling as done
  for the scrollbars itself: does not work well with native scrolling
  because scroll events are not fired immediately -> shadow jumps
  around.

292522

### DIFF
--- a/eclipse-scout-core/src/form/fields/groupbox/GroupBox.js
+++ b/eclipse-scout-core/src/form/fields/groupbox/GroupBox.js
@@ -31,6 +31,7 @@ import {
   TabBox,
   TabItemKeyStroke,
   tooltips,
+  widgets,
   WrappedFormField
 } from '../../../index';
 import $ from 'jquery';
@@ -72,6 +73,7 @@ export default class GroupBox extends CompositeField {
     this.$body = null;
     this.$title = null;
     this.$subLabel = null;
+    this._statusPositionOrig = null;
   }
 
   static BorderDecoration = {
@@ -99,6 +101,7 @@ export default class GroupBox extends CompositeField {
       menuOrder: new GroupBoxMenuItemsOrder(),
       ellipsisPosition: this.menuBarEllipsisPosition
     });
+    this.menuBar.on('propertyChange:visible', () => this._updateMenuBarStyle());
     this._setFields(this.fields);
     this._setMainBox(this.mainBox);
     this._updateMenuBar();
@@ -218,6 +221,7 @@ export default class GroupBox extends CompositeField {
     this.addContainer(this.$parent, this.mainBox ? 'root-group-box' : 'group-box');
 
     this.$header = this.$container.appendDiv('group-box-header');
+    HtmlComponent.install(this.$header, this.session); // Complete layout chain for elements inside header (e.g. allow top status to invalidate layout when visibility changes)
     this.$title = this.$header.appendDiv('title');
     this.$borderBottom = this.$header.appendDiv('bottom-border');
     this.addLabel();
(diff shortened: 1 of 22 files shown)


cguglielmo:
run tests

avandorp:
run tests

cguglielmo:
run tests

avandorp:
run tests

eclipse-scout-core/src/form/fields/groupbox/GroupBox.js
```diff
@@ -360,6 +364,36 @@ export default class GroupBox extends CompositeField {
...
+  _updateScrollShadow() {
+    if (!this.rendered) {
+      return;
+    }
+    let hasScrollShadowTop = this.hasScrollShadow('top');
+    let hasScrollShadowBottom = this.hasScrollShadow('bottom');
+    let oldHasScrollShadowTop = this.$container.hasClass('has-scroll-shadow-top');
+    this.$container.toggleClass('has-scroll-shadow-top', hasScrollShadowTop);
+    this.$container.toggleClass('has-scroll-shadow-bottom', hasScrollShadowBottom);
+    if (oldHasScrollShadowTop !== hasScrollShadowTop) {
```
mvilliger, line 381: There is no need to compare the bottom scroll-shadow as well?

eclipse-scout-core/src/scrollbar/Scrollbar.js
```diff
@@ -76,6 +76,9 @@ export default class Scrollbar extends Widget {
     this._dirReverse = this.axis === 'x' ? 'right' : 'bottom';
     this._scrollDir = this.axis === 'x' ? 'scrollLeft' : 'scrollTop';
 
+    // this.$parent.appendDiv('invisible scroll-shadow ' + this._dir);
+    // this.$parent.appendDiv('invisible scroll-shadow ' + this._dirReverse);
```
mvilliger, line 80: still required?

eclipse-scout-core/src/scrollbar/scrollbars.js
```diff
@@ -127,6 +140,190 @@ export function _installNativeInternal($container, options) {
...
+  _installIntersectionObserver();
+  intersectionObserver.observe($container[0]);
+}
+
+function uninstallScrollShadow($container, session) {
+  let $shadow = $container.data('scroll-shadow');
+  if ($shadow) {
+    $shadow.remove();
+  }
+  intersectionObserver.unobserve($container[0]);
```
mvilliger, line 166: remove 'scroll-shadow' and 'scroll-shadow-style' from $container data?
And remove 'scroll' event listener from container?

eclipse-scout-core/src/widget/widgets.js
```diff
@@ -67,10 +67,26 @@ export function findFirstFocusableWidget(widgets, container, checkTabbable) {
   return arrays.find(widgets, widget => widget.isFocusable(checkTabbable));
 }
 
+export function preserveAndSetProperty(setter, getter, preserver, preserverName) {
```
mvilliger, line 70: Add jsdoc

PR #21, opened 2021-09-15
Table: don't show broken font icon in tooltip

cguglielmo:
Tooltip does not know how to display a font icon -> don't include
it when generating the plain text.

302034

### DIFF
--- a/eclipse-scout-core/src/encoder/PlainTextEncoder.js
+++ b/eclipse-scout-core/src/encoder/PlainTextEncoder.js
@@ -36,6 +36,10 @@ export default class PlainTextEncoder {
     // Separate td with ' '
     text = text.replace(/<\/td>/gi, ' ');
 
+    if (options.removeFontIcons) {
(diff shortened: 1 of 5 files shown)


eclipse-scout-core/src/encoder/PlainTextEncoder.js
```diff
@@ -36,6 +36,10 @@ export default class PlainTextEncoder {
     // Separate td with ' '
     text = text.replace(/<\/td>/gi, ' ');
 
+    if (options.removeFontIcons) {
+      text = text.replace(/<span>*[ \n]+class="[^"]*font-icon[^"]*">[^<]*<\/span>/gmi, '');
```
mvilliger: Is the first ">" correct?
E.g. use \s instead of [ \n]?

cguglielmo: No, it is not correct😁 thx. \s would be better as well

PR #24, opened 2021-09-22
Limit text length on smart field and tag field

cyrill-wyss:
Currently smart field and tag field allow an infinite text to be entered
by the user. this may cause two problems:
1. the website inside the browser may crash due to an out of memory
   exception.
2. the ui server thread handling the submitted text may crash due to an
   out of memory exception. This may also cause a security risk.

To solve this, the smart field and tag field will now have a max length
property and the ui server truncates all received text longer than the
max length value.
The .js fields will only allow to type as many characters as the max
length value, including text pasting.
The logic has been extracted from StringField.js and moved into a new
MaxLengthHandler.js which allows all fields to use the same logic.
In addition the local lookup call is made more resilient against regex DoS.

295423

Signed-off-by: Cyrill Wyss <cyrill.wyss@bsi-software.com>

### DIFF
--- a/eclipse-scout-core/src/util/MaxLengthHandler.js
+++ b/eclipse-scout-core/src/util/MaxLengthHandler.js
@@ -1,0 +1,130 @@
+/*
+ * Copyright (c) 2010-2021 BSI Business Systems Integration AG.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     BSI Business Systems Integration AG - initial API and implementation
+ */
+import {arrays, objects, scout, Status} from '../index';
+import $ from 'jquery';
+import {assertValue} from '../scout';
+
+export default class MaxLengthHandler {
+
+  constructor(options) {
+    options = options || {};
+    assertValue(options.target, 'target is mandatory');
+
+    $.extend(this, options);
+  }
+
+  install($textInputField) {
+    this.uninstall();
+    if (!$textInputField || (!$textInputField.is('input:text') && !$textInputField.is('textArea'))) {
+      return;
+    }
+    if ($textInputField) {
+      this.$textInputField = $textInputField;
+      this.$textInputField.on('paste', this._onInputFieldPaste.bind(this));
+    }
+  }
+
+  uninstall() {
+    if (this.$textInputField) {
+      this.$textInputField.off('paste', this._onInputFieldPaste.bind(this));
+    }
+  }
+
+  render() {
+    if (!this.$textInputField || objects.isNullOrUndefined(this.target.maxLength)) {
+      return;
+    }
+    // Check if "maxLength" attribute is supported by browser
+    if (this.$textInputField[0].maxLength) {
+      this.$textInputField.attr('maxlength', this.target.maxLength);
+    } else {
+      // Fallback for IE9
+      this.$textInputField.on('keyup paste', e => {
+        setTimeout(truncate.bind(this), 0);
+      });
+    }
+
+    // Make sure current text does not exceed max length
+    truncate.call(this);
+    if (!this.target.rendering) {
+      this.target.parseAndSetValue(this.target._readDisplayText());
+    }
+
+    function truncate() {
+      let text = this.$textInputField.val();
+      if (text.length > this.target.maxLength) {
+        this.$textInputField.val(text.slice(0, this.target.maxLength));
+      }
+    }
+  }
+
+  _onInputFieldPaste(event) {
+    if (!this.$textInputField || objects.isNullOrUndefined(this.target.maxLength)) {
+      return;
+    }
+    // must read out the text and selection size now because when the callback is executed, the clipboard content has already been applied to the input field
+    let textSize = this.$textInputField.val().length - this._getSelectionSize();
+
+    this._getClipboardData(event, pastedText => {
+      if (!pastedText) {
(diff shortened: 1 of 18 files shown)


avandorp:
Jenkins, please run tests.

eclipse-scout-core/src/util/MaxLengthHandler.js
```diff
@@ -0,0 +1,130 @@
...
+    $.extend(this, options);
+  }
+
+  install($textInputField) {
+    this.uninstall();
+    if (!$textInputField || (!$textInputField.is('input:text') && !$textInputField.is('textArea'))) {
+      return;
+    }
+    if ($textInputField) {
+      this.$textInputField = $textInputField;
```
awegmueller, line 30: member variables should be defined in the CTOR

eclipse-scout-core/src/util/MaxLengthHandler.js
```diff
@@ -0,0 +1,130 @@
...
+  constructor(options) {
+    options = options || {};
+    assertValue(options.target, 'target is mandatory');
+
+    $.extend(this, options);
+  }
+
+  install($textInputField) {
+    this.uninstall();
+    if (!$textInputField || (!$textInputField.is('input:text') && !$textInputField.is('textArea'))) {
```
awegmueller, line 26: I guess it doesn't matter, but DOM element names never us camel-case, thus --> textarea instead of textArea

eclipse-scout-core/src/util/MaxLengthHandler.js
```diff
@@ -0,0 +1,130 @@
...
+      return;
+    }
+    if ($textInputField) {
+      this.$textInputField = $textInputField;
+      this.$textInputField.on('paste', this._onInputFieldPaste.bind(this));
```
awegmueller, line 31: bind() always creates a new function, does the function you register here is not the same as the one you unregister on L37 - you must store the function returned by bind in a member variable!

eclipse-scout-core/src/util/MaxLengthHandler.js
```diff
@@ -0,0 +1,130 @@
...
+
+  render() {
+    if (!this.$textInputField || objects.isNullOrUndefined(this.target.maxLength)) {
+      return;
+    }
+    // Check if "maxLength" attribute is supported by browser
+    if (this.$textInputField[0].maxLength) {
+      this.$textInputField.attr('maxlength', this.target.maxLength);
+    } else {
+      // Fallback for IE9
```
awegmueller, line 49: Remove that in Scout 22.0

eclipse-scout-core/src/util/MaxLengthHandler.js
```diff
@@ -0,0 +1,130 @@
...
+      this.$textInputField.on('keyup paste', e => {
+        setTimeout(truncate.bind(this), 0);
```
awegmueller, line 51: SmartField has its own paste handling, is the order in which the paste handlers are executed important?

eclipse-scout-core/src/util/MaxLengthHandler.js
```diff
@@ -0,0 +1,130 @@
...
+      });
+    }
+
+    // Make sure current text does not exceed max length
+    truncate.call(this);
```
awegmueller, line 56: why use call() here and not simply this.truncate()?

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/form/fields/smartfield/AbstractSmartField.java
```diff
@@ -838,6 +865,9 @@ public void lookupByText(String text) {
   }
 
   protected void lookupByTextInternal(String text, boolean synchronous) {
+    if (StringUtility.length(text) > getMaxLength()) {
```
awegmueller, line 868: I guess the If here is not required because the substring method deals with the case, that the string length <= max-length. So you could always call substring

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/services/lookup/LocalLookupCall.java
```diff
@@ -86,15 +86,15 @@ protected Pattern createSearchPattern(String s) {
...
 
     // replace repeating wildcards to prevent regex DoS
     String duplicateWildcards = WILDCARD_PLACEHOLDER.concat(WILDCARD_PLACEHOLDER);
     while (s.contains(duplicateWildcards)) {
       s = s.replace(duplicateWildcards, WILDCARD_PLACEHOLDER);
     }
+    s = StringUtility.escapeRegexMetachars(s);
 
-    if (!s.contains(WILDCARD_PLACEHOLDER)) {
+    if (!s.endsWith(WILDCARD_PLACEHOLDER)) {
```
awegmueller, line 97: What has this change to do with the max-length change?

eclipse-scout-core/src/util/MaxLengthHandler.js
```diff
@@ -0,0 +1,130 @@
...
+        this._showNotification('ui.PastedTextTooLong');
+      }
+    });
+  }
+
+  _getSelectionSize() {
+    let start = scout.nvl(this.$textInputField[0].selectionStart, null);
+    let end = scout.nvl(this.$textInputField[0].selectionEnd, null);
+    if (start === null || end === null) {
+      start = 0;
```
awegmueller, line 90: simply return 0; here?

awegmueller (commented):
See comments in review.

PR #26, opened 2021-09-22
Data object: add exist check for convenience methods

stephan-merkli:
By using an exist check before, the node/collection is not created when these methods are used instead the ones on the collection itself.


paolobazzi (commented):
Change looks good but unit tests for changed behavior are missing.

PR #27, opened 2021-09-23
Simplify usage of DragAndDropHandler.js

cyrill-wyss:
The current drag & drop handler needs too much boilerplate code to use
and is not well documented.
With this change the handler will now provide a simple install and
uninstall function where all parameters are documented and can be set
within an options object.


avandorp:
Jenkins, please run tests.

PR #28, opened 2021-09-23
Features/cgu/queue microtask

cguglielmo:


eclipse-scout-core/src/layout/LayoutValidator.js
```diff
@@ -67,18 +67,30 @@ export default class LayoutValidator {
...
-        this.validate();
+    if (this._validateTimeoutId !== null) {
+      // Task already scheduled
+      return;
+    }
+    if (Device.get().supportsMicrotask()) {
+      this._validateTimeoutId = true; // boolean is sufficient, an id is only necessary for legacy purpose when setTimeout is used
+      queueMicrotask(() => {
+        // Validate, but only if still required
+        if (this._validateTimeoutId) {
```
mvilliger, line 78: Is there a risk, that as part of the call to validate() the layout is invalidated again?
Because this could lead to the situation that within the microtask new microtasks are scheduled which lets the browser process the microtasks endlessly. Can this happen? Maybe we should guard it to prevent endless processing? Whats your opinion?

cguglielmo: This is possible in theory, but this was already possible with the current implementation as well. I don't think it is necessary to guard it because calling invalidateLayoutTree inside a layout operation must be avoided anyway.

mvilliger: Ok, fine 👍

PR #34, opened 2021-10-02
BrushUp: ModeSelector

mvilliger:
Please review the ModeSelector brush up.
I think we have to talk about colors, especially for dark mode :)
Thanks


eclipse-scout-core/src/modeselector/Mode.less
```diff
@@ -31,45 +32,35 @@
...
 
-  &:hover {
-    color: @mode-hover-text-color;
-  }
-
-  &:not(.disabled):not(.selected):active {
-    color: @button-active-color;
-    background-color: @button-active-background-color;
+  &.selected {
+    color: @default-button-color; // override button.selected
```
cguglielmo: Please use @selected-color instead of default-button-color.

eclipse-scout-core/src/modeselector/Mode.less
```diff
@@ -31,45 +32,35 @@
...
   }
 
-  &.selected {
-    color: @mode-selected-text-color;
-    background-color: @mode-selected-background-color;
-    border: none;
+  &.selected:hover {
+    color: @mode-selected-text-color; // override action:hover
```
cguglielmo: Overriding of action rules is currently necessary because main.less contains the button rules which comes after the Action.less containing the action rules. I already stumbled upon that in another case and I think we should improve it. Suggestion: Move the rules for the .button class (not the button tag) to a file called Button.less inside the action folder. Add the Button.less after the Action.less to index.less.

eclipse-scout-core/src/modeselector/Mode.less
```diff
@@ -79,11 +70,33 @@
...
+    color: @mode-text-color;
+    border: @mode-alternative-border-width transparent solid;
+    background-color: transparent;
+    border-radius: @button-border-radius + @mode-alternative-border-width;
+
+    &.disabled {
+      color: @button-disabled-color;
+    }
+
+    &:hover:not(.selected):not(.disabled) {
```
cguglielmo: if possible please limit the usage of :not. The longer the rule is the harder it is to customize because the whole rule needs to be copied. Mostly it can be simplified by adjusting the order -> disabled should be at the end (hover, active, selected, focus, disabled)

eclipse-scout-core/src/modeselector/Mode.less
```diff
@@ -79,11 +70,33 @@
...
+      background-color: rgba(0, 0, 0, 0.1);
+    }
 
     &.selected:not(.disabled) {
-      background-color: inherit;
-      border: 1px solid @mode-alternative-selected-color;
-      color: @mode-alternative-selected-color;
+      color: @button-color;
+
+      -webkit-transition: color 500ms ease;
```
cguglielmo: no need to use the browser prefix, browser support is good enough

eclipse-scout-core/src/modeselector/Mode.less
```diff
@@ -9,12 +9,13 @@
  *     BSI Business Systems Integration AG - initial API and implementation
  */
 .mode {
-  display: table-cell;
   vertical-align: middle;
```
cguglielmo: vertical-align, text-align and height should be obsolete

eclipse-scout-core/src/style/colors.less
```diff
@@ -286,10 +286,11 @@
 @menu-item-disabled-color: @disabled-color;
 @messagebox-background-color: @popup-background-color;
 @messagebox-body-color: @palette-black;
-@mode-hover-text-color: @hover-color;
-@mode-selected-background-color: @selected-background-color;
-@mode-selected-text-color: @selected-color;
-@mode-alternative-selected-color: @mode-selected-background-color;
+@mode-text-color: @palette-gray-7;
```
cguglielmo: please remove -text- , just write @mode-color.

eclipse-scout-core/src/style/colors.less
```diff
@@ -286,10 +286,11 @@
...
+@mode-hover-text-color: @mode-text-color;
+@mode-selected-text-color: @default-button-color;
```
cguglielmo: Use @selected-color instead of default-button-color

eclipse-scout-core/src/style/colors.less
```diff
@@ -286,10 +286,11 @@
...
+@mode-alternative-selected-background-color: @button-background-color;
+@mode-alternative-background-color: #ebebeb;
```
cguglielmo: Please use a color from the palette or add a todo marker so that I can adjust it once the color palette is final

eclipse-scout-core/src/modeselector/Mode.less
```diff
@@ -31,45 +32,35 @@
...
-    border: none;
+  &.selected:hover {
+    color: @mode-selected-text-color; // override action:hover
   }
 
-  &:not(.disabled) {
-    cursor: pointer;
+  &:not(.disabled):not(.selected):hover {
+    color: @button-active-color; // override action:hover
   }
```
cguglielmo, line 54: currently it looks like the selected button can be clicked (hover and active states are applied). If clicking the selected button does not trigger an event, background-color should not be changed on hover/active

eclipse-scout-core/src/modeselector/ModeSelector.js
```diff
@@ -57,6 +63,11 @@ export default class ModeSelector extends Widget {
     }, this);
   }
 
+  _renderSlider() {
+    this.$slider = this.$container.appendDiv('slider');
```
cguglielmo: Maybe call css class mode-slider to avoid a conflict with the slider from slider.less

eclipse-scout-core/src/modeselector/ModeSelector.js
```diff
@@ -108,6 +125,27 @@ export default class ModeSelector extends Widget {
         visibleModes[selectedModeIndex + 1].$container.addClass('after-selected');
       }
     }
+    this._updateSlider();
+  }
+
+  _updateSlider() {
```
cguglielmo, line 128: It would be nice if the slider could be moved by dragging. Especially on touch devices the user could want to enable it this way. It think it is not necessary to visualize the dragging itself, just move the slider when the finger / mouse is on top of the new mode.

cguglielmo (commented):
Looks good. Nevertheless, I have some comments :)

PR #35, opened 2021-10-05
Data object contributions: lazy init of internal representation

stephan-merkli:
Most data objects won't have contributions. Always initializing m_contributions with an empty list isn't necessary therefore. m_contributions is now initialized on first access via IDoEntity#getContributions.

Providing an additional IDoEntity#hasContributions method that will not create the internal representation. IDoEntity#getContributions must always the create internal representation because a mutable list is returned.

### DIFF
--- a/org.eclipse.scout.rt.dataobject.test/src/test/java/org/eclipse/scout/rt/dataobject/DataObjectContributionTest.java
+++ b/org.eclipse.scout.rt.dataobject.test/src/test/java/org/eclipse/scout/rt/dataobject/DataObjectContributionTest.java
@@ -25,6 +25,14 @@ import org.junit.runner.RunWith;
 public class DataObjectContributionTest {
 
   @Test
+  public void testHasContributions() {
+    SimpleFixtureDo doEntity = BEANS.get(SimpleFixtureDo.class);
+    assertFalse(doEntity.hasContributions());
+    doEntity.putContribution(BEANS.get(FirstSimpleContributionFixtureDo.class));
+    assertTrue(doEntity.hasContributions());
+  }
+
+  @Test
   public void testHasGetContribution() {
     SimpleFixtureDo doEntity = BEANS.get(SimpleFixtureDo.class);
     assertTrue(doEntity.getContributions().isEmpty());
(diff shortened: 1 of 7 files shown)


org.eclipse.scout.rt.dataobject.test/src/test/java/org/eclipse/scout/rt/dataobject/DataObjectContributionTest.java
```diff
@@ -103,5 +111,17 @@ public void testEquality() {
...
+    assertEquals(doEntity1.hashCode(), doEntity2.hashCode());
+  }
+
+  @Test
+  public void testEqualityHashCodeWithLazyInit() {
+    SimpleFixtureDo doEntity1 = BEANS.get(SimpleFixtureDo.class);
+    SimpleFixtureDo doEntity2 = BEANS.get(SimpleFixtureDo.class);
+    doEntity2.getContributions(); // internal contribution list is created
+
+    // Order of contributions must not be relevant for comparison
```
paolobazzi: Update comment

org.eclipse.scout.rt.jackson/src/main/java/org/eclipse/scout/rt/jackson/dataobject/DefaultDoEntityDeserializerTypeStrategy.java
```diff
@@ -48,7 +49,9 @@ public String resolveTypeVersion(Class<? extends IDoEntity> entityClass) {
   @Override
   public void putContributions(IDoEntity doEntity, String attributeName, Collection<?> contributions) {
     // add contributions to corresponding list in do entity
-    //noinspection unchecked
-    doEntity.getContributions().addAll((Collection<? extends IDoEntityContribution>) contributions);
+    if (!CollectionUtility.isEmpty(contributions)) {
```
paolobazzi, line 52: Add this check also in raw case?
org.eclipse.scout.rt.jackson.dataobject.RawDoEntityDeserializerTypeStrategy#putContributions

paolobazzi (approved):

PR #37, opened 2021-10-06
Simplify usage of DragAndDropHandler.js

cyrill-wyss:
The current drag & drop handler needs too much boilerplate code to use
and is not well documented.
With this change the handler will now provide a simple install and
uninstall function where all parameters are documented and can be set
within an options object.

### DIFF
--- a/eclipse-scout-core/src/form/fields/clipboardfield/ClipboardField.js
+++ b/eclipse-scout-core/src/form/fields/clipboardfield/ClipboardField.js
@@ -8,7 +8,7 @@
  * Contributors:
  *     BSI Business Systems Integration AG - initial API and implementation
  */
-import {arrays, Device, dragAndDrop, InputFieldKeyStrokeContext, keys, mimeTypes, scout, Session, strings, ValueField} from '../../../index';
+import {arrays, Device, InputFieldKeyStrokeContext, keys, mimeTypes, scout, Session, strings, ValueField} from '../../../index';
 import $ from 'jquery';
 
 export default class ClipboardField extends ValueField {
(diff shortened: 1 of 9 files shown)


avandorp:
Jenkins, please run tests.

eclipse-scout-core/src/form/fields/clipboardfield/ClipboardField.js
```diff
@@ -17,7 +17,6 @@ export default class ClipboardField extends ValueField {
     super();
 
     this.allowedMimeTypes = null;
-    this.maximumSize = null;
```
cguglielmo: AbstractClipboardField still has a maximumSize so removing it will break the Java code. The "renaming" from maximumSize to dropMaximumSize will also break existing Scout JS code so you need to provide migration notes.

Two properties are indeed not ideal, either combine them and use the lower value or you convert the maximumSize in ClipboardFieldAdapter to dropMaximumSize by implementing _initProperties and _syncMaximumSize and add migration nodes to MigrationGuide_22_0.adoc

eclipse-scout-core/src/util/dragAndDrop.js
```diff
@@ -116,19 +116,101 @@ export function dataTransferTypesContains(dataTransfer, needleArray) {
...
   options = options || {};
-  options.target = target;
+  if (!options.target) {
+    return null;
+  }
   return new DragAndDropHandler(options);
 }
 
+/**
+ * installs or uninstalls a drag & drop handler on the target.
```
cguglielmo: Please use js doc data types for every parameter and option

cguglielmo:
Thank you! I rebased the branch locally and pushed it directly.

PR #38, opened 2021-10-06
Improve MessageBoxes helper

cyrill-wyss:
The helper to create delete confirmation messages only offers limited
functionality. For customized messages such as setting the error level
or tuning the header text, the helpers logic has to be duplicated.

To solve this, the creation of the messagebox is split into a separate
function.
The show function still exists to ensure backwards compatibility.
Also the item list has been tuned to show how many items are not shown
in the confirmation message.

288246


avandorp:
Jenkins, please run tests.

PR #43, opened 2021-10-11
Mode selector: mode text is not centered in alternative mode

cguglielmo:
Use transparent border instead of a padding. This way no compensation
is necessary which makes it easier.
Also add padding to mode selector itself instead of using mode borders.
This makes the hover effect look better.


bschwarzent (approved):
I tested both styles in normal and dense mode. Looks good!

PR #49, opened 2021-10-14
Fix tooltip-position if it overlaps the left edge

egloff:
Move tooltip to this.windowPaddingX if it overlaps the left edge.


egloff:
Hey Jenkins, please rerun tests

PR #61, opened 2021-11-03
Limit text length on smart field and tag field

cyrill-wyss:
Currently smart field and tag field allow an infinite text to be entered
by the user. this may cause two problems:
1. the website inside the browser may crash due to an out of memory
   exception.
2. the ui server thread handling the submitted text may crash due to an
   out of memory exception. This may also cause a security risk.

To solve this, the smart field and tag field will now have a max length
property and the ui server truncates all received text longer than the
max length value.
The .js fields will only allow to type as many characters as the max
length value, including text pasting.
The logic has been extracted from StringField.js and moved into a new
MaxLengthHandler.js which allows all fields to use the same logic.

In addition the code lookup call will also be resistent against regex
DoS.

295423

Signed-off-by: Cyrill Wyss <cyrill.wyss@bsi-software.com>

### DIFF
--- a/org.eclipse.scout.rt.shared.test/src/test/java/org/eclipse/scout/rt/shared/services/lookup/CodeLookupCallTest.java
+++ b/org.eclipse.scout.rt.shared.test/src/test/java/org/eclipse/scout/rt/shared/services/lookup/CodeLookupCallTest.java
@@ -71,7 +71,6 @@ public class CodeLookupCallTest {
   public void testGetDataByTextFiltered() {
     runGetDataByTextFiltered(null);
     runGetDataByTextFiltered("*or*");
-    runGetDataByTextFiltered("*or");
     runGetDataByTextFiltered("ip");
     runGetDataByTextFiltered("foo");
     runGetDataByTextFiltered("ipi");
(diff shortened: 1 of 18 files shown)


awegmueller (approved):

org.eclipse.scout.rt.shared.test/src/test/java/org/eclipse/scout/rt/shared/services/lookup/CodeLookupCallTest.java
```diff
@@ -69,18 +69,24 @@ public void testGetDataByText() {
 
   @Test
   public void testGetDataByTextFiltered() {
+    // test text filters which yield the same result as the legacy call.
     runGetDataByTextFiltered(null);
     runGetDataByTextFiltered("*or*");
-    runGetDataByTextFiltered("*or");
```
awegmueller, line 74: I'd prefer if you'd simply delete this test(line), instead of adding more complexity just for this odd legacy lookup call case. The rest looks Ok.

awegmueller (approved):

PR #64, opened 2021-11-09
Add filter for fields to store in the xml export

tgygax:
Before all fields of a form would be stored in the xml export. This
filter allows the form to customize which field end up in the xml
export.

### DIFF
--- a/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/form/AbstractForm.java
+++ b/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/form/AbstractForm.java
@@ -36,6 +36,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Function;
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/form/AbstractForm.java
```diff
@@ -209,6 +210,9 @@
   private Map<Class<?>, Class<? extends IFormField>> m_fieldReplacements;
   private IContributionOwner m_contributionHolder;
 
+  // filter for fields to store in xml export
+  private Predicate<IFormField> m_fieldXmlFilter;
```
andibur: Should we call it `m_storeToXmlFieldFilter` to make its purpose a bit clearer (i.e. that the filter is not used when loading from XML).

andibur (commented):
Thank you for the change. Could you extend an existing or add a new test covering the new flexibility (check org.eclipse.scout.rt.client.ui.form.StoreAndLoadXml1FormTest and other tests in the same package)

andibur (approved):
Thank you!

PR #66, opened 2021-11-15
RestLookupCall: fix restriction precedence

bschwarzent:
When creating clones of the template lookup call, values predefined in
the model of a RestLookupCall ('restriction' property) should take
precedence over values that are automatically added to
this._restriction. This behavior was not well defined before, which
could have lead to unexpected behavior.
One example was the 'maxRowCount' property which was always taken from
the lookup call, even when a different value was predefined in the
'restriction' object.

Documentation on these precedence rules was added to RestLookupCall.js.
Additionally, a Jasmine test was added asserting these rules.

306914

### DIFF
--- a/eclipse-scout-core/src/lookup/RestLookupCall.js
+++ b/eclipse-scout-core/src/lookup/RestLookupCall.js
@@ -8,9 +8,38 @@
  * Contributors:
  *     BSI Business Systems Integration AG - initial API and implementation
  */
-import {arrays, LookupCall, scout} from '../index';
+import {arrays, LookupCall, objects, scout} from '../index';
 import $ from 'jquery';
 
+/**
+ * A lookup call that can load lookup rows from a REST service.
+ *
+ * API:
+ * ----
+ * By default, the REST service is expected to listen for POST requests at the URL defined by
+ * this.resourceUrl. It receives a restriction object and must return a list of matching lookup rows.
+ * The serialization format is JSON.
+ *
+ * Lookup rows:
+ * ------------
+ * The standard lookup row properties defined by Scout are usually sufficient (see AbstractLookupRowDo.java).
+ *
+ * Restriction:
+ * ------------
+ * The restriction object consists of a number of 'well-known' properties (e.g. 'text' in QueryBy.TEXT
+ * mode, see AbstractLookupRestrictionDo.java for details) and additional, service-dependent properties
+ * that can either be predefined in the model or added programmatically at runtime. Since all of those
+ * properties are sent in the same restriction object, some care must be taken to prevent accidental
+ * overwriting of properties.
+ *
+ * Order of precedence (lowest to highest):
+ * 1. Restrictions automatically applied to all clones after their creation in the respective cloneFor method.
+ *    These are: 'active' (ALL, TEXT, REC) and 'maxRowCount' (ALL, TEXT, REC)
+ * 2. Restrictions predefined in the model property 'restriction', shared by all clones.
+ * 3. Restrictions applied to clones programmatically, e.g. during a 'prepareLookupCall' event.
+ * 4. Hard-coded properties that are fundamental to the respective queryBy mode (cannot be overridden).
+ *    These are: 'ids' (KEY, KEYS) and 'text' (TEXT)
+ */
 export default class RestLookupCall extends LookupCall {
 
   constructor() {
@@ -42,6 +71,17 @@ export default class RestLookupCall extends LookupCall {
     this._restriction[key] = value;
   }
 
+  /**
+   * Adds the given key-value pair to 'this._restriction', but only if there is no predefined
+   * value for this key in 'this.restriction'. This prevents unintentional overriding of
+   * user-defined model restrictions.
+   */
+  _addRestrictionIfAbsent(key, value) {
+    if (!this.restriction || objects.isNullOrUndefined(this.restriction[key])) {
+      this.addRestriction(key, value);
+    }
+  }
+
   _getAll() {
     return this._call();
   }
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/src/lookup/RestLookupCall.js
```diff
@@ -42,6 +71,17 @@ export default class RestLookupCall extends LookupCall {
     this._restriction[key] = value;
   }
 
+  /**
+   * Adds the given key-value pair to 'this._restriction', but only if there is no predefined
+   * value for this this key in 'this.restriction'. This prevents unintentional overriding of
```
cyrill-wyss: Remove one 'this'

eclipse-scout-core/test/lookup/RestLookupCallSpec.js
```diff
@@ -0,0 +1,327 @@
...
+
+    const expectedDefaultRestriction = {
+      active: true,
+      maxRowCount: 777
+    };
+
+    expect(lookupCall.restriction).toBeNull();
+    expect(lookupCall._restriction).toBeNull();
+
+    const cloneAll = lookupCall.cloneForAll();
```
cyrill-wyss, line 38: The checks from the different test functions look quite similar. Maybe it would be better to extract them into separate functions?

eclipse-scout-core/test/lookup/RestLookupCallSpec.js
```diff
@@ -0,0 +1,327 @@
...
+      myProperty: 'someValue'
+    });
+
+    // Check template again (must not have been altered)
+    expect(lookupCall.restriction).toBeNull();
+    expect(lookupCall._restriction).toBeNull();
+  });
+
+  it('applies restrictions in the expected order', () => {
+    // This test asserts that the following precedence rules are applies then creating the final restriction object:
```
cyrill-wyss, line 101: The list does not match the one in the lookup call comment.

eclipse-scout-core/test/lookup/RestLookupCallSpec.js
```diff
@@ -0,0 +1,327 @@
...
+    // 2. predefined restriction
+    // 3. dynamic restrictions added by cloneForX()
+    // 4. dynamic restrictions added by user code (e.g. during 'prepareLookupCall' event)
+
+    const lookupCall = scout.create('RestLookupCall', {
+      session: session,
+      resourceUrl: 'test-api/dummy',
+      maxRowCount: 777,
+      active: true, // will be ignored
+      myProperty: 'one',
```
cyrill-wyss: I guess this should be ignored too?

eclipse-scout-core/test/lookup/RestLookupCallSpec.js
```diff
@@ -0,0 +1,327 @@
...
+    spyOn(lookupCall, '_call'); // <-- disable the ajax call in this test
+
+    const expectedModelRestriction = {
+      maxRowCount: 888,
+      active: false,
+      myProperty: 'two',
+      myDefaultProperty: 'three'
+    };
+
+    const expectedDefaultRestriction = {
```
cyrill-wyss: looks a bit strange to have two semantic identical objects => Add a comment to describe the position change in the ajax call restriction

eclipse-scout-core/src/lookup/RestLookupCall.js
```diff
@@ -8,9 +8,38 @@
...
+ * ------------
+ * The restriction object consists of a number of 'well-known' properties (e.g. 'text' in QueryBy.TEXT
+ * mode, see AbstractLookupRestrictionDo.java for details) and additional, service-dependent properties
+ * that can either be predefined in the model or added programmatically at runtime. Since all of those
+ * properties are sent in the same restriction object, some care must be taken to prevent accidental
+ * overwriting of properties.
+ *
+ * Order of precedence (lowest to highest):
+ * 1. Restrictions automatically applied to all clones after their creation.
+ *    These are: 'active' (ALL, TEXT, REC) and 'maxRowCount' (ALL, TEXT, REC)
```
paolobazzi, line 37: maybe add:  
"1. Restrictions automatically applied to all clones after their creation **by their corresponding cloneFor method**"

paolobazzi (approved):

eclipse-scout-core/test/lookup/RestLookupCallSpec.js
bschwarzent: 2 of the 3 test methods were almost identical, I merged them into one. The third has many small differences. It may be a bit verbose, but at least the test cases are clearly identifiable.

eclipse-scout-core/test/lookup/RestLookupCallSpec.js
bschwarzent: Correct, I updated the comment

eclipse-scout-core/test/lookup/RestLookupCallSpec.js
bschwarzent: Yes, comment was added

eclipse-scout-core/test/lookup/RestLookupCallSpec.js
bschwarzent: Semantically, they are not exactly the same, but in this particular case, they contain exactly the same properties (the order is not important). There is no default property left that could be used to show that difference, therefore I removed one of these objects.

eclipse-scout-core/src/lookup/RestLookupCall.js
bschwarzent: OK

PR #68, opened 2021-11-23
#62 File Drop on a table row no more working

### LINKED ISSUE #62: File Drop on a table row no more working
Hello,

I just upgraded my Scout application from Scout 10 to Scout 11. I use the Java environment. I use the file drop feature on Table Pages, both on rows and on empty space.

When I drop the file on a row, I expect to get the target row in my application. It used to work with Scout 10.

After the upgrade to Scout 11, it doesn't work anymore. I did some debugging, firstly in the Java application, then in the browser. What I see is that the 'rowId' property is empty in the uploadProperties variable in JsonTable.consumeBinaryResource.

I checked what happens in Table.js, I see this:
```
  _createDragAndDropHandler() {
    return dragAndDrop.handler(this, {
      supportedScoutTypes: dragAndDrop.SCOUT_TYPES.FILE_TRANSFER,
      onDrop: event => this.trigger('drop', event),
      dropType: () => this.dropType,
      dropMaximumSize: () => this.dropMaximumSize,
      additionalDropProperties: event => {
        let $target = $(event.currentTarget);
        let properties = {
          rowId: ''
        };
       if ($target.hasClass('table-row')) {    // <=== issue here
          let row = $target.data('row');
          properties.rowId = row.id;
        }
        return properties;
      }
    });
  }
```
The indicated line is where it fails. The target has no 'table-row' class, because at this point it is the table itself, not the row. So the rowId property remains an empty string. Anyway, since it is not a row, the target has no 'row' property. Here is the data as seen under the debugger:
```
additionalDropProperties
  <this>: {...}
  $target: {...}
    0: div.table.detail-table.parts-page.view
    length: 1
    <prototype>: {...}
```

Thank you.
Michael

cyrill-wyss:
supply the correct event parameter when calling additionalDropProperties() to allow matching the event source with the corresponding element of the ui server.

307416

### DIFF
--- a/org.eclipse.scout.rt.ui.html/src/main/js/scout/form/fields/FormFieldAdapter.js
+++ b/org.eclipse.scout.rt.ui.html/src/main/js/scout/form/fields/FormFieldAdapter.js
@@ -38,7 +38,7 @@ scout.FormFieldAdapter.prototype._goOnline = function() {
 
 scout.FormFieldAdapter.prototype._onWidgetEvent = function(event) {
   if (event.type === 'drop' && this.widget.dragAndDropHandler) {
-    this.widget.dragAndDropHandler.uploadFiles(event.files);
+    this.widget.dragAndDropHandler.uploadFiles(event);
   } else {
     scout.FormFieldAdapter.parent.prototype._onWidgetEvent.call(this, event);
   }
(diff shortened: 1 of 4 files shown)


cguglielmo (approved):
Looks good, thank you!

PR #77, opened 2021-12-08
Test MR Build: Adds new function _prepareDisplayTest to SmartField

egloff:
Split preparing of display-text from _renderDisplayText into separate
function _prepareDisplayText to allow simple customization.


egloff:
run test

PR #86, opened 2022-01-06
Add utility function objects#isPromise that checks for thenables

michaelze:
This change adds a new function to the objects utility that checks a provided
value for being a thenable or not, according to the Promises/A+ spec as
mentioned here: https://promisesaplus.com/#terminology.


michaelze:
@bschwarzent Can you please have a look?
Is the target branch correct? Should be no problem to port back to older releases...

bschwarzent:
Thanks! 22.0 is fine for the moment, we can backport it later when needed.

PR #88, opened 2022-01-17
Add UuidHelper

jwinkens:
Adds URL-safe support for base64 encoding/decoding of Uuids


bschwarzent:
Thanks for your contribution!

PR #97, opened 2022-01-28
FilterSupport: move icon slightly and add transparent background

fschinkel:
Add a transparent background color for all browsers that do not support
blur (e.g. Firefox).

312135


fschinkel:
run tests

PR #99, opened 2022-01-28
Browser recognition: add support for Chrome on iOS

fschinkel:
The Chrome on iOS behaves the same way Safari does (different UI but the
same engine) but has a different version scheme. Therefore, consider the
AppleWebKit-version for all Safari-like browser in order to determine
whether a browser is supported or not.


mvilliger (approved):
+1

fschinkel:
run tests

fschinkel:
Run tests

fschinkel:
run tests

fschinkel:
run tests

PR #102, opened 2022-02-04
Create own modules for datamodel related classes

mrudolf9954:
Moved all datamodel related classes to their new modules.
That also includes the IComposerField which has also been relocated.


mrudolf9954:
run tests

PR #105, opened 2022-02-15
Remove unused class MethodState

csk-bsi:
314041


org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/AbstractExtensionChain.java
```diff
@@ -105,23 +105,19 @@ private void computePrevious(boolean forceRewind) {
     }
   }
 
-  protected void callChain(MethodInvocation<?> methodInvocation, Object... arguments) {
+  protected void callChain(MethodInvocation<?> methodInvocation) {
```
andibur, line 114: Keep old method, mark as deprecated and delegate to new one without varargs.

andibur (approved):
Thanks

PR #112, opened 2022-02-22
RestLookupCall: more specific error message defined

torstentrompler:
A generic error message ('Failed to load the data.') is returned if the
RestLookupCall runs into an error. The same message is displayed if the
user does not have sufficient permissions for the lookup call. To
distinguish this situation, a more specific message ('You are not
authorized to read this data.') is now returned.


torstentrompler:
No longer required

PR #114, opened 2022-02-24
Refactor fix for silent page reloading to IDesktop

jwinkens:
* Added outline class ID to IBookmarkAdapter and DefaultBookmarkAdapter
* Refactored a fix for silent page reloading in AbstractPage to 
IDesktop, so it can be adapted for different Desktops/used in other places

### DIFF
--- a/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/IDesktop.java
+++ b/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/IDesktop.java
@@ -402,9 +402,9 @@ public interface IDesktop extends IWidget, IDisplayParent, IStyleable, IContextM
   void refreshPages(List<Class<? extends IPage<?>>> pages);
 
   /**
-   * @see IDesktop#refreshPages(List)
    * @param pageTypes
    *          Must be classes that implement {@link IPage}.
+   * @see IDesktop#refreshPages(List)
    */
   void refreshPages(Class<?>... pageTypes);
 
@@ -689,10 +689,10 @@ public interface IDesktop extends IWidget, IDisplayParent, IStyleable, IContextM
 
   /**
    * Activates a {@link Bookmark} on this desktop.
-   * <p />
+   * <p>
    * First the specific {@link Bookmark#getOutlineClassName()} is evaluated and activated, afterwards every page from
    * the {@link Bookmark#getPath()} will be selected (respecting the {@link AbstractPageState}).
-   * <p />
+   * <p>
    * Finally the path will be expanded. Possible exceptions might occur if no outline is set in the {@link Bookmark} or
    * the outline is not available.
    */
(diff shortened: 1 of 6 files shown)


avandorp:
Jenkins, please run tests

jwinkens:
Hey @avandorp , tests may not run as this pull request is still a WIP? I don't really know but thought I should clarify just in case so you don't lose time. Gonna reach out when it's ready.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/IDesktop.java
```diff
@@ -1038,4 +1038,11 @@ default void removeDataChangeListener(IDataChangeListener listener, Object... da
    */
   NativeNotificationDefaults getNativeNotificationDefaults();
 
+  /**
+   * Special handling for reloading pages via usage of Bookmarks. This can be used to "silently" reload a page without
+   * changing active windows.
+   *
+   * @since 22.0
+   */
+  void refreshOutlineByBookmark(IPage<?> page);
```
andibur: The fact, that a bookmark is used, is an implementation detail. What about
```suggestion
  void reloadPageFromRoot(IPage<?> page);
```

andibur (changes requested):
Thanks Justin for the change. I would rename the new method on the desktop as suggested. Please change the java doc accordingly.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/IDesktop.java
jwinkens: Hey @andibur , good suggestions. I changed it.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/IDesktop.java
```diff
@@ -1038,4 +1038,10 @@ default void removeDataChangeListener(IDataChangeListener listener, Object... da
    */
   NativeNotificationDefaults getNativeNotificationDefaults();
 
+  /**
+   * Reloads a page from root. This can be used to "silently" reload a page without changing active windows.
```
cguglielmo: Could you please explain the purpose of this method in more detail? I honestly don't really understand what it does.

jwinkens: Hey @cguglielmo I tried to come up with a more sensical description. Does that help? Default Implementation in AbstractDesktop uses Bookmarks to do this, which you do not necessarily want for every desktop because not every desktop supports Bookmarks.

cguglielmo: Yes, a lot better, thank you!

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/IDesktop.java
andibur: Perfect, thank you.

andibur: Perfect, thank you.

andibur (approved):

PR #115, opened 2022-02-24
Tree: improve performance and fix some bugs

cguglielmo:
Improves
- Expanding all nodes (change in _applyFiltersForNodeRec)
- Lazy expanding a node (change in setNodeExpanded)
- Adding nodes (change in insertNodes and LazyNodeFilter)
  delete operation is very expensive
- Initializing (change in init)
- Simplified updateSelectionPath because after splitting it up,
  the selected node was always expanded after initializing

Fixes
- bread crumb showed too many nodes because filterVisibleNodes did
  not check children (see spec)
- adding a node to a lazy expanded parent (double click table row)
  created many animation-wrappers which stayed forever in the DOM
  without creating an animation.
- nodes were invisible after parent was expanded. Happened after
  collapsing the parent while nodes were hiding due to a filter.
- _nodesFiltered should always be executed, even if apply is false,
  to have a consistent model (not an actual bug)

312492

### DIFF
--- a/eclipse-scout-core/src/tree/LazyNodeFilter.js
+++ b/eclipse-scout-core/src/tree/LazyNodeFilter.js
@@ -18,10 +18,9 @@ export default class LazyNodeFilter {
   accept(node) {
     if (node.expanded) {
       return true;
-    } else {
-      // not expanded: remove lazy expand marker (forget lazy expanded children)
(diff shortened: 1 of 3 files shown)


cguglielmo:
Jenkins, run tests

fschinkel (approved):

PR #120, opened 2022-03-02
Add new LambdaUtility

csk-bsi:
Adds two new interfaces UnsafeRunnable and UnsafeSupplier for invoking a supplier/runnable with checked Exceptions.


avandorp (approved):
Thanks. Looks good to me.

PR #122, opened 2022-03-02
Reduce MemoryFootprint for InternalTableRow

csk-bsi:
Use Integer as BitMask instead of Set<Integer>

315409

### DIFF
--- a/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/internal/InternalTableRow.java
+++ b/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/internal/InternalTableRow.java
@@ -34,16 +34,10 @@ import org.eclipse.scout.rt.shared.data.basic.FontSpec;
  */
 public class InternalTableRow extends TableRow implements ICellObserver {
 
+  private final Map<ICell, Integer> m_updatedCells;
   private ITable m_table;
   private int m_rowIndex;
   private int m_rowChanging = 0;
-  private final Map<ICell, Set<Integer>> m_updatedCells;
-
-  private InternalTableRow() {
-    super(null);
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/internal/InternalTableRow.java
```diff
@@ -528,6 +516,21 @@ private void setCellChanged(ICell cell, int changeBit) {
     return indexesByCell;
   }
 
+  private Integer setBit(Integer bitMask, int bitPos) {
+    if (bitMask == null) {
+      bitMask = 0;
+    }
+    bitMask |= 1 << bitPos;
+    return bitMask;
```
andibur: This version saves a boxing and unboxing operation in case `bitMask` is `null`:
```suggestion
    return bitMask == null
        ? 1 << bitPos
        : bitMask.intValue() | 1 << bitPos;
```

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/internal/InternalTableRow.java
```diff
@@ -528,6 +516,21 @@ private void setCellChanged(ICell cell, int changeBit) {
...
+  }
+
+  private boolean isBitSet(Integer bitMask, int bitPos) {
+    if (bitMask == null) {
+      return false;
+    }
+    return (bitMask & 1 << bitPos) != 0;
```
andibur:
```suggestion
    return bitMask != null
        && (bitMask.intValue() & 1 << bitPos) != 0;
```

andibur (approved):
Looks good. Thanks.

Added two suggestions (both minor).

PR #137, opened 2022-03-15
Fix reloading from root

jwinkens:
When reloading a page from root, reload active page instead of page
that triggered the reload. Because intermediate pages may need to reload
as well if a child node changes, reloading the triggering page causes
jumping around in outline.


egloff (changes requested):
Looks good for me, this was done the same way previously in AbstractPage.
But shouldn't the commend read "... from root, **bookmark** active page instead..." as we want to bookmark the active page but reload the page which triggered the change?

egloff (approved):

PR #152, opened 2022-03-23
Improve memory allocation for InternalTableRow

csk-bsi:
* improve getColumnIndexes creation of new map has no significant impact on performance
* improve ArrayList creation by giving proper length

### DIFF
--- a/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/internal/InternalTableRow.java
+++ b/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/internal/InternalTableRow.java
@@ -11,6 +11,7 @@
 package org.eclipse.scout.rt.client.ui.basic.table.internal;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/internal/InternalTableRow.java
```diff
@@ -207,15 +209,16 @@ public Object getCellValue(int columnIndex) {
...
-      for (int keyIndex : keyColumns) {
-        pk.add(getCell(keyIndex).getValue());
-      }
+    if (getTable() == null) {
+      return new ArrayList<>(0);
+    }
+    int[] keyColumns = getTable().getColumnSet().getKeyColumnIndexes();
+    if (keyColumns.length == 0) {
+      keyColumns = getTable().getColumnSet().getAllColumnIndexes();
+    }
```
andibur: This would save creating an empty array:
```suggestion
    ColumnSet columnSet = getTable().getColumnSet();
    int[] keyColumns = columnSet.getKeyColumnCount() > 0 ? columnSet.getKeyColumnIndexes() : columnSet.getAllColumnIndexes();
```

andibur (commented):
Small improvement for the case that no primary key columns are defined.

PR #155, opened 2022-03-24
ProposalField: reset error status if search text changes

gabrielstalder:
Fixes an issue that a proposal field does not clear the error status if the search text changes (only if a new lookup row is selected). This led to a wrong error message if a wrong text value is replaced with a correct text value.

Ticket Ref.: 315639

### DIFF
--- a/eclipse-scout-core/src/form/fields/smartfield/ProposalField.js
+++ b/eclipse-scout-core/src/form/fields/smartfield/ProposalField.js
@@ -189,11 +189,12 @@ export default class ProposalField extends SmartField {
     // this causes a lookup which may fail and open a new proposal chooser (property
     // change for 'result').
     if (searchTextChanged) {
+      this.clearErrorStatus();
       this._acceptByText(sync, searchText);
     } else if (!this._hasUiError()) {
       this._inputAccepted(false);
     } else {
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/src/form/fields/smartfield/ProposalField.js
```diff
@@ -189,11 +189,12 @@ export default class ProposalField extends SmartField {
     // this causes a lookup which may fail and open a new proposal chooser (property
     // change for 'result').
     if (searchTextChanged) {
+      this.clearErrorStatus();
```
awegmueller, line 192: Code changes look Ok, but please add a Jasmine (ProposalFieldSpec.js) or a Selenium (com.bsiag.widgets.ui.html.form.fields.smartfield.SmartFieldTest) test for the reported issue and bugfix. Without it, someone could easily break your bugfix with a later change without noticing.

Also fix the commit message. We only add the plain ticket-no. on the last line, without further explanation or prefixes.

gabrielstalder: Thanks for your feedback. I've added a simple test and adjusted the commit message.

PR #172, opened 2022-04-08
Table/Tree: enable certain keystrokes even if widget is disabled

cguglielmo:
Some keystroke like navigation keystrokes (up, down, etc.) or
copy keystrokes should work even if the widget is disabled.

In order to make the keystrokes work, a disabled widget must be able
to receive the focus. This is normally not the case.
Also, a keystroke won't be executed if the widget is disabled.
-> A disabled table, tree or tile grid is now focusable but still not
   tabbable.
-> A keystroke now has the property inheritAccessibility which is true
   by default. If the keystroke should not check enabledComputed of the
   widget, inheritAccessibility can be set to false.
-> Navigation keystrokes (up, down, collapse etc.) now have
   inheritAccessibility set to false. So has the copy keystroke of the
   table.
-> A keystroke now checks the property enabledComputed of the widget
   rather than just enabled.
   Because: Keystrokes of the tile grid worked if the tile field was
   disabled but did not work if the tile grid itself was disabled.

229591


eclipse-scout-core/src/keystroke/KeyStroke.js
```diff
@@ -138,9 +139,19 @@ export default class KeyStroke {
...
-      if (this.field.enabled !== undefined && !this.field.enabled) {
-        return false;
+      // Check enabled state (if inheritAccessibility is true)
+      if (!this.inheritAccessibility) {
+        return true;
+      }
+      if (this.field.enabledComputed !== undefined) {
+        if (!this.field.enabledComputed) {
+          return false;
+        }
```
mvilliger: just return this.field.enabledComputed here?

eclipse-scout-core/src/keystroke/KeyStroke.js
```diff
@@ -138,9 +139,19 @@ export default class KeyStroke {
...
+      } else if (this.field.enabled !== undefined) {
```
mvilliger: Then the else here is not necessary

eclipse-scout-core/src/keystroke/KeyStroke.js
```diff
@@ -138,9 +139,19 @@ export default class KeyStroke {
...
+        // This should actually not happen because this.field should always be a hypothetical case if this.field is not a widget
+        if (!this.field.enabled) {
+          return false;
+        }
```
mvilliger: and here the same: just return this.field.enabled?

is more readable in my opinion and should have the same logic?

eclipse-scout-core/src/keystroke/TabItemKeyStroke.js
```diff
@@ -19,9 +19,9 @@ export default class TabItemKeyStroke extends KeyStroke {
 
     this.renderingHints.offset = 16;
     this.renderingHints.hAlign = HAlign.RIGHT;
-    this.renderingHints.$drawingArea = function($drawingArea, event) {
+    this.renderingHints.$drawingArea = ($drawingArea, event) => {
```
mvilliger: maybe inline this arrow func?

eclipse-scout-core/src/menu/MenuExecKeyStroke.js
```diff
@@ -19,9 +19,9 @@ export default class MenuExecKeyStroke extends KeyStroke {
     this.stopPropagation = true;
 
     this.renderingHints.offset = 16;
-    this.renderingHints.$drawingArea = function($drawingArea, event) {
+    this.renderingHints.$drawingArea = ($drawingArea, event) => {
       return this.field.$container;
-    }.bind(this);
+    };
```
mvilliger: maybe inline here too?

mvilliger (approved):

eclipse-scout-core/src/keystroke/KeyStroke.js
cguglielmo: much better, thanks

eclipse-scout-core/src/keystroke/KeyStroke.js
cguglielmo: yep

eclipse-scout-core/src/keystroke/KeyStroke.js
cguglielmo: yep

eclipse-scout-core/src/keystroke/TabItemKeyStroke.js
cguglielmo: it was inlined with the second commit

eclipse-scout-core/src/menu/MenuExecKeyStroke.js
cguglielmo: it was inlined with the third commit

PR #179, opened 2022-04-19
Improve filterBox support for LookupBox

fschinkel:
Display inactive lookupRows in italic on ListBox, TreeBox and SmartField
(non-hierarchic mode) and add an '(inactive)'-suffix.

### DIFF
--- a/eclipse-scout-core/src/form/fields/LookupBox.js
+++ b/eclipse-scout-core/src/form/fields/LookupBox.js
@@ -3,7 +3,7 @@
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
+ * https://www.eclipse.org/legal/epl-v10.html
  *
  * Contributors:
  *     BSI Business Systems Integration AG - initial API and implementation
@@ -67,6 +67,8 @@ export default class LookupBox extends ValueField {
     this.$field.addDeviceClass();
     this.$field.addClass('structure');
     this._renderFilterBox();
+
+    this.$container.css('--inactive-lookup-row-suffix-text', `'${this.session.text('InactiveState')}'`);
   }
 
   _renderFilterBox() {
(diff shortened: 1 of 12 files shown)


eclipse-scout-core/src/form/fields/LookupBox.less
```diff
@@ -43,6 +43,23 @@
         top: 0;
         height: 100%;
       }
+
+      .active-filter-undefined&.inactive {
```
cguglielmo: This does not work with Scout Classic, AbstractListBox already adds inactive text and style

eclipse-scout-core/src/form/fields/LookupBox.less
```diff
@@ -43,6 +43,23 @@
...
+        font-style: italic;
+
+        & > .table-cell.last > .text::after {
+          content: ' (' var(--inactive-lookup-row-suffix-text) ')';
+        }
+      }
+    }
+
+    .active-filter-undefined& > .tree-data > .tree-node.inactive,
```
cguglielmo: Does this work in the Scout classic widgets app? In my test, the suffix is added even if only inactive elements are shown (inactive radio button checked)

eclipse-scout-core/src/form/fields/smartfield/SmartField.js
```diff
@@ -932,6 +915,10 @@ export default class SmartField extends ValueField {
     });
 
     this.popup.open();
+    this.popup.$container.css('--inactive-lookup-row-suffix-text', '\'' + this.session.text('InactiveState') + '\'');
+    if (this.activeFilter) {
```
cguglielmo: Maybe call _renderActiveFilter?

eclipse-scout-core/src/form/fields/smartfield/SmartField.js
```diff
@@ -1341,6 +1328,17 @@ export default class SmartField extends ValueField {
     this.setProperty('activeFilter', this.activeFilterEnabled ? activeFilter : null);
   }
 
+  _renderActiveFilter() {
```
cguglielmo: Is this a realistic case? If I simulate it (update activeFilter while popup is open), the nodes are updated but the active filter bar will be removed.

eclipse-scout-core/src/form/fields/LookupBox.js
```diff
@@ -67,6 +67,8 @@ export default class LookupBox extends ValueField {
     this.$field.addDeviceClass();
     this.$field.addClass('structure');
     this._renderFilterBox();
+
+    this.$container.css('--inactive-lookup-row-suffix-text', '\'' + this.session.text('InactiveState') + '\'');
```
cguglielmo: Template string syntax would be more readable

PR #181, opened 2022-04-19
Date and Time column: value picker does not work in mobile mode

cyrill-wyss:
When in mobile mode, the quick value picker for date and time columns
can not update the column. Choosing the new value via text input works.

In mobile mode the picker popup will be closed as soon as a value is
chosen. Closing the popup so soon will fire the cell changed event
before the date field was able to fire an accept input event.

By calling triggerAcceptInput just before closePopup in the date field
we ensure to first send the new field value before the table sends a
cell changed event.

319005


cguglielmo:
Thank you! We will track the issue with this PR: https://github.com/eclipse-scout/scout.rt/pull/183

PR #185, opened 2022-04-21
Date field in an editable table throws an error in touch mode

torstentrompler:
Prerequisites
1. a table with an editable date column
2. active touch mode by using a mobile device or browser developer tools

Use case:
1. click on the date column
2. pick a date -> UI error

Problem 1:
The date field registers a popup listener on the 'remove' event within
the method openDatePopup(). The listener calls the method
_onPopupRemove() and then calls the method removeClass() on the date
field, but at this point the field no longer exists. The method
_onPopupRemove() implicitly destroys the date field, because in touch
mode completeEdit() gets called on the cell editor popup and this will
destroy the corresponding field.

To fix problem 1 removeClass() must be called before _onPopupRemove().

Problem 2 (requires problem 1 to be solved):
By picking a date the method _onDatePickerDateSelect() gets called and
closes the popup, that leads to a call of destroy() on the popup itself.
The method Widget#destroy triggers a 'remove' event and then removes all
children from the owner (i.e. the date field), but at this point the
field no longer exists.

The popup listener mentioned in problem 1 is registered to the 'remove'
event and will call completeEdit() in touch mode, that will destroy the
corresponding field.

To fix problem 2 the listener must be registered to the 'close' event of
the popup, because this event will be called in the close method of the
popup before it gets destroyed.

319180

### DIFF
--- a/eclipse-scout-core/src/form/fields/datefield/DateField.js
+++ b/eclipse-scout-core/src/form/fields/datefield/DateField.js
@@ -1385,10 +1385,12 @@ export default class DateField extends ValueField {
     this.popup = this.createDatePopup();
     this.popup.open();
     this.$dateField.addClass('focused');
-    this.popup.on('remove', event => {
-      this._onPopupRemove(event);
-      this.popup = null;
+    this.popup.one('destroy', event => {
+      // Removing the class must happen before _onPopupDestroy() is called, otherwise the date field no longer exists,
+      // because in touch mode _onPopupDestroy() destroys the date field.
       this.$dateField.removeClass('focused');
+      this._onPopupDestroy(event);
+      this.popup = null;
     });
     this.getDatePicker().on('dateSelect', this._onDatePickerDateSelect.bind(this));
   }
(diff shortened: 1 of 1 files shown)


eclipse-scout-core/src/form/fields/datefield/DateField.js
```diff
@@ -1385,10 +1385,16 @@ export default class DateField extends ValueField {
...
     this.popup.open();
     this.$dateField.addClass('focused');
-    this.popup.on('remove', event => {
-      this._onPopupRemove(event);
-      this.popup = null;
+    // After picking a date the popup gets closed and destroyed. The method Widget#destroy triggers a 'remove' event
+    // and then tries to remove all children from the owner (i.e. the date field). Do not register the following popup
+    // listener on the 'remove' event, because it will destroy the date field in touch mode! That will lead to an error,
+    // when the popup tries to remove the children from the no longer existing owner.
+    this.popup.on('close', event => {
```
cguglielmo: Please use destroy instead of close, because close can be prevented and destroy will be triggered at the very end.
Please also use one instead of on. Although it should not be possible that it will be executed more than once, it is semantically more correct.

eclipse-scout-core/src/form/fields/datefield/DateField.js
```diff
@@ -1385,10 +1385,16 @@ export default class DateField extends ValueField {
...
+      // Removing the class must happen before _onPopupClose() is called, otherwise the date field no longer exists,
+      // because in touch mode _onPopupClose() destroys the date field!
```
cguglielmo: No need to use an exclamation mark in comments

eclipse-scout-core/src/form/fields/datefield/DateField.js
```diff
@@ -1385,10 +1385,16 @@ export default class DateField extends ValueField {
     this.popup = this.createDatePopup();
     this.popup.open();
     this.$dateField.addClass('focused');
-    this.popup.on('remove', event => {
-      this._onPopupRemove(event);
-      this.popup = null;
+    // After picking a date the popup gets closed and destroyed. The method Widget#destroy triggers a 'remove' event
```
cguglielmo: I don't think the comment here is necessary, it is described in the commit message well enough. Destroy is the better event to listen for, so the chance that it will be reverted to remove is very small.

cguglielmo (commented):
Good solution! Thank you.

PR #192, opened 2022-04-26
Planner: Wrong scale after horizontal scrolling and switching weeks

torstentrompler:
Prerequisites:
a PlannerField with horizontal scrollbars and at least one resource

Use case:
1. scroll to the right
2. press next or previous button
-> the scale is offset

Problem:
Pressing the next, today or previous button will set a new view range
and leads to a call of the _renderScale() method. This method empty and
recreate the scale but does not consider the current horizontal
scrollbar position.

To fix the problem the scroll position of the scale must be restored
after rendering.

317594


torstentrompler:
rerun tests

PR #199, opened 2022-04-29
SmartField: display text is empty but a value set

cguglielmo:
Use case:
1. Use RestLookupCall (which implements abort())
2. Call setValue multiple times -> display text may be empty

Problem:
The second setValue will call lookupCall.abort() which rejects the
lookup call promise. It will then schedule a new lookupCall to update
the text. Unfortunately, the fail callback of the aborted promise
will be executed after the done callback of the new pormise and
overrides the display text with empty string again.

Reason:
The fail callbacks will be executed at the end of a task,
even if the promises are rejected before another promise is resolved.

Example:
$.rejectedPromise()
  .then(() => console.log('resolve 1'))
  .catch(() => console.log('reject 1'))
$.resolvedPromise()
  .then(() => console.log('resolve 2'))
  .catch(() => console.log('reject 2'))

output:
resolve 2
reject 1

319091

### DIFF
--- a/eclipse-scout-core/src/form/fields/ValueField.js
+++ b/eclipse-scout-core/src/form/fields/ValueField.js
@@ -28,6 +28,7 @@ export default class ValueField extends FormField {
     this.value = null;
     this.validators = [];
     this.validators.push(this._validateValue.bind(this));
+    this._updateDisplayTextPending = false;
 
     this.$clearIcon = null;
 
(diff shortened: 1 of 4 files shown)


eclipse-scout-core/src/form/fields/ValueField.js
```diff
@@ -517,16 +518,24 @@ export default class ValueField extends FormField {
     }
     value = scout.nvl(value, this.value);
     let returned = this.formatValue(value);
-    if (returned && $.isFunction(returned.promise)) {
+    if (returned && objects.isPromise(returned)) {
```
bschwarzent:
```suggestion
    if (objects.isPromise(returned)) {
```

bschwarzent (changes requested):

PR #201, opened 2022-05-03
FocusManager: Allow preventing to scroll when focusing an element

gabrielstalder:
Change to support _preventScroll_ when focusing elements. The second commit uses this feature to improve the TileGrid.


eclipse-scout-core/src/focus/FocusManager.js
```diff
@@ -251,26 +251,43 @@ export default class FocusManager {
...
-    return this.requestFocus(element, filter, true);
+    return this.requestFocus(element, filter, {
+      onlyIfReady: true
+    });
   }
 
   /**
    * Requests the focus for the given element, but only if being a valid focus location.
    *
+   * @param {HTMLElement, $} [element]
```
cguglielmo: Please use | instead of , , as it is the official syntax: https://jsdoc.app/tags-param.html?msclkid=1495682dcd2911eca72dcc8e2122179a#multiple-types-and-repeatable-parameters

PR #208, opened 2022-05-09
Fix size of chart avg icon

samueleiben:
Also centered the sum icon vertically.

320550


cguglielmo:
Looks good, thank you!

PR #212, opened 2022-05-11
Improve DefaultValuesFilter for single properties

csk-bsi:
If stack contains only one element skip StringBuilder und directly return element


bschwarzent (approved):
Looks good to me.

PR #213, opened 2022-05-11
Add config property for username used by DevelopmentAccessController

michaelze:
This change adds a new Scout config property that allows to overrride
the username used by the DevelopmentAccessController when constructing
the AnonymousAuthConfig. This is useful, when the username used by your
application differs from the one you use on your development machine.


michaelze:
@imotsch We already discussed this. Do you think this is a good solution for our problem?

avandorp:
Jenkins, please run tests.

@michaelze : I like the change.

imotsch:
Yes this change makes sense, especially when using indirect authentication providers.

michaelze:
Thanks for your comments! The current config property key `dev.user.name` is derived from the previously used (and now used as a default value) system property `user.name`.
Do you think `dev.username` would be a more suitable choice?

michaelze:
@andibur I updated my PR according to our discussion. Please have a look!

andibur (approved):
Well done, Thanks Michael.

andibur:
Jenkins, please run tests

PR #214, opened 2022-05-12
TileTableField: reduce table header padding as item state is not visible

fschinkel:


eclipse-scout-core/src/tile/fields/tablefield/TileTableField.less
```diff
@@ -53,6 +53,27 @@
...
+          padding-right: @table-header-item-padding-right;
+
+          &.halign-right,
+          &.halign-center {
+            padding-right: @table-cell-padding-left;
+          }
+        }
+
+        &.sorted.grouped,
+        &.sorted.filtered {
```
cguglielmo: Please add the rules to the ones above, it should not be necessary to duplicate them. Actually, they should not be necessary at all because the rules for the tile table field are already more specific than the ones from the TableHeader.less.

eclipse-scout-core/src/tile/fields/tablefield/TileTableField.less
```diff
@@ -53,6 +53,27 @@
...
+          padding-right: @table-header-item-padding-right;
+
+          &.halign-right,
+          &.halign-center {
+            padding-right: @table-cell-padding-left;
```
cguglielmo: The header text and the cell content are not aligned correctly for a right aligned column if it is not the last column.

Proposal:

```
&.sorted,
&.grouped,
&.filtered {
  padding-right: @table-header-item-padding-right;

  &.halign-right,
  &.halign-center {
    padding-right: @table-header-item-padding-left;
  }

  &.last.halign-left,
  &.last.halign-center,
  &.last.halign-right {
    padding-right: @table-header-item-padding-right-last;
  }
}
```

PR #221, opened 2022-05-23
DoValue/DoList: Use initialValue instead of creating new empty values

paolobazzi:
DoValue/DoList: use provided initial value directly instead of creating
an empty list which is replaced by the initial value later on while
constructing the DoEntity.

320299

### DIFF
--- a/org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/DoEntity.java
+++ b/org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/DoEntity.java
@@ -10,6 +10,8 @@
  */
 package org.eclipse.scout.rt.dataobject;
 
+import static org.eclipse.scout.rt.platform.util.Assertions.*;
+
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/DoEntity.java
```diff
@@ -76,15 +76,19 @@ public void putNode(String attributeName, DoNode<?> attribute) {
...
   }
 
   /**
-   * Adds new list value to attribute map. The value is wrapped within a {@link DoList}.
+   * Associates the specified list value with the specified attribute name in this entity. If the entity previously
+   * contained a mapping for the attribute name, the old list value is replaced by the specified value.<br>
+   * The value is wrapped within a {@link DoList}.
    */
   @Override
   public <V> void putList(String attributeName, List<V> value) {
```
ralph-steiner, line 100: Does this method still create a new empty list which will be immediately thrown away?

paolobazzi: No - the javadoc comment is misleading. I'll change it.

PR #223, opened 2022-05-24
Support column order update by id only

cyrill-wyss:
When updating the column order, the new order must consist of a list
with whole column elements. However, the function itself would only need
an ordered list of ids.

To support both cases (whole column object or id only), updateColumnOrder
now only uses the id part of the column.
This allows updating by id only while maintaining the original api.

321037


cguglielmo:
Looks good, thanks!

PR #224, opened 2022-05-25
DoValue/DoList: Use initialValue instead of creating new empty values

paolobazzi:
DoValue/DoList: use provided initial value directly instead of creating
an empty list which is replaced by the initial value later on while
constructing the DoEntity.

320299


paolobazzi:
Jenkins please run tests

paolobazzi:
Jenkins please run tests

PR #228, opened 2022-05-31
DoEntity: Clean empty/null-values function

paolobazzi:
New method DataObjectHelper.clean to remove all null DoValues and
empty DoCollection nodes from a DoEntity and its nested objects

321620


bschwarzent (approved):
👍

PR #234, opened 2022-06-07
NotificationBadge: only add when necessary and fix compareTo

cguglielmo:
Adding status with the same message did not work because
compareTo returned 0 and MultiStatus uses a TreeSet.

### DIFF
--- a/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/status/Status.java
+++ b/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/status/Status.java
@@ -11,6 +11,7 @@
 package org.eclipse.scout.rt.platform.status;
 
 import java.io.Serializable;
+import java.util.Comparator;
 
 import org.eclipse.scout.rt.platform.IOrdered;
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/status/Status.java
```diff
@@ -167,25 +168,14 @@ public boolean isMultiStatus() {
...
-      return -1;
-    }
-    return 0;
+    return Comparator
+        .comparing(IStatus::getSeverity)
+        .thenComparing(IStatus::getOrder)
+        .thenComparing(IStatus::getMessage,Comparator.nullsLast(Comparator.naturalOrder()))
+        .thenComparing(IStatus::getIconId,Comparator.nullsLast(Comparator.naturalOrder()))
+        .thenComparing(IStatus::getCssClass,Comparator.nullsLast(Comparator.naturalOrder()))
+        .thenComparing(IStatus::getCode)
```
mwagnerxaq, line 176: Maybe we should compare the code before the cssClass, because the code was in place before cssClass was added to Status

PR #239, opened 2022-06-15
Security: download as attachment for HTML files not enforced

cguglielmo:
HTML files must not be opened directly because it could
contain malicious JavaScript which will be executed in the context
of the current http session

322241


mvilliger:
Looks good for Scout Classic.
I wonder how it is handled in Scout JS? The OpenUriHandler.js uses OPEN as default action. Maybe better use DOWNLOAD as default?
Because e.g. the FileChooserForm uses openUri with default action? And it could be a html file?
What do you think?

cguglielmo:
If we want to improve it for Scout JS apps in general, we should provide a similar solution as for Scout Classic: Determine the action based on the extension. We should consider that for Scout 23. Currently, the app is responsible to choose the right action.

mvilliger:
Ok for me

PR #241, opened 2022-06-16
Detect content type from content or filename

egloff:
Do not use content type passed by the browser as this can be influenced
by the uploader.

323187


imotsch:
The solution is good. However please add a junit test that verifies that the java implementation in fact checks the content type. For all major enums in the MimeType.java check in the junit test if the found mime type matches the one in MimeType.java.

egloff:
Ok, it seems the detection with the file-extension in  is sufficient, especially as the content-type is matched against the content in org.eclipse.scout.rt.ui.html.json.UploadRequestHandler#verifyFileIntegrity afterwards. So we keep only the overridable method that it can be implemented with a custom behavior.

PR #245, opened 2022-06-24
Use workaround for uri handler only if protocol is externally handled

egloff:
According to the description the (still not fixed) bug [1] affects
only externally handled protocols (which is already handled correctly
in handleUriActionOpen). This commit implements the same behavior for
handleUriActionDownload and adds msteams to externally handled
protocols list.

[1] https://bugs.chromium.org/p/chromium/issues/detail?id=663325

319237


egloff:
Additionally tested on Android (Chrome)

PR #249, opened 2022-07-01
REST Client: ScoutApacheConnector: Initial re-implementation

paolobazzi:
Reimplementation of JAX-RS Jersey connector interface providing Apache
HTTP client support for Jersey REST clients.
The new implementation supports various properties, e.g. dynamic proxy
selection, see RestClientProperties interface for all properties.
Furthermore the new implementation suppresses Jersey and Apache HTTP
client from sending default user agent header  (no disclosing of
internal information). If the invoker does not set a custom user agent
header, a general user agent header "Generic" is added.

See Bugzilla 572405

322071, 300860


paolobazzi:
Please run tests

paolobazzi:
run tests

PR #253, opened 2022-07-08
FilterSupport: fix background-color for mobile mode

fschinkel:
Add a post layout validation function if the widget is not yet rendered
to ensure that the background color can be determined correctly.

321454

### DIFF
--- a/eclipse-scout-core/src/widget/FilterSupport.js
+++ b/eclipse-scout-core/src/widget/FilterSupport.js
@@ -3,7 +3,7 @@
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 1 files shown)


eclipse-scout-core/src/widget/FilterSupport.js
```diff
@@ -164,6 +162,16 @@ export default class FilterSupport extends WidgetSupport {
     this._filterField.keyStrokeContext.registerKeyStroke(this._exitFilterFieldKeyStroke);
   }
 
+  _updateFilterFieldBackgroundColor() {
+    let color = styles.getFirstOpaqueBackgroundColor(this._filterField.$container),
```
cguglielmo, line 169: Please check if filterField is still rendered. It may have been removed in the mean time.

cguglielmo (commented):
See comment, otherwise it looks good, thanks.

PR #255, opened 2022-07-12
Table: do not trigger row click if contextMenuPopup was opened

fschinkel:
311794


eclipse-scout-core/src/table/Table.js
```diff
@@ -623,6 +623,11 @@ export default class Table extends Widget {
       return;
     }
 
+    if (mouseButton === 3 && this.contextMenu && this.contextMenu.rendered) {
```
cguglielmo, line 626: This check does not work most of the time because context menu will be opened later (when requests are done, see showContextMenu). And the tree triggers a click event as well even if the context menu is open. In previous releases a row click was possible as well if the user slightly moved the cursor while clicking. So I would suggest to not change the current behavior.

PR #258, opened 2022-07-14
Add JsForm

fschinkel:
The JsForm is a Java-form that wraps a form implemented in JS. This
allows a Scout Classic application to open a ScoutJS form. The whole
lifecycle of this form will be handled by ScoutJS.
Add possibility to not show a ScoutJS form on open().
Improve detailForm support of ScoutJS page.

325223

### DIFF
--- a/eclipse-scout-core/src/desktop/outline/pages/Page.js
+++ b/eclipse-scout-core/src/desktop/outline/pages/Page.js
@@ -1,9 +1,9 @@
 /*
- * Copyright (c) 2010-2021 BSI Business Systems Integration AG.
+ * Copyright (c) 2010-2022 BSI Business Systems Integration AG.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
+ * https://www.eclipse.org/legal/epl-v10.html
  *
  * Contributors:
  *     BSI Business Systems Integration AG - initial API and implementation
@@ -198,6 +198,11 @@ export default class Page extends TreeNode {
     if (form instanceof Form) {
       form.setModal(false);
       form.setClosable(false);
+
+      form.setDisplayHint(Form.DisplayHint.VIEW);
+      form.setDisplayViewId('C');
+
+      form.setShowOnOpen(false);
     }
     if (form instanceof TileOverviewForm) {
       form.setPage(this);
--- a/eclipse-scout-core/src/form/Form.js
+++ b/eclipse-scout-core/src/form/Form.js
@@ -1,9 +1,9 @@
 /*
- * Copyright (c) 2010-2021 BSI Business Systems Integration AG.
+ * Copyright (c) 2010-2022 BSI Business Systems Integration AG.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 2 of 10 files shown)


eclipse-scout-core/src/form/Form.js
```diff
@@ -650,7 +651,9 @@ export default class Form extends Widget {
   }
 
   show() {
-    this.session.desktop.showForm(this);
+    if (this.showOnOpen) {
```
cguglielmo: The name says show on open but the implementation blocks every call of show() -> please move the check to open().

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/form/js/AbstractJsForm.java
```diff
@@ -0,0 +1,166 @@
...
+
+  @Override
+  public Class<OUT> getOutputDataType() {
+    return m_outputDataType;
+  }
+
+  @Override
+  public IN getInputData() {
+    @SuppressWarnings("unchecked")
+    IN value = (IN) propertySupport.getProperty(PROP_INPUT_DATA);
```
cguglielmo: No need to have a variable here

eclipse-scout-core/src/desktop/outline/pages/Page.js
```diff
@@ -198,6 +198,14 @@ export default class Page extends TreeNode {
     if (form instanceof Form) {
       form.setModal(false);
       form.setClosable(false);
+
+      form.setDisplayHint(Form.DisplayHint.VIEW);
+      form.displayViewId = 'C';
+
+      form.showOnOpen = false;
+      if (form.rootGroupBox) {
```
cguglielmo: Why is it necessary to set menubar position explicitly? This prevents having a bottom menubar. I checked some detail forms with Scout Classic and JS, the position is set to AUTO. The GroupBox renders AUTO to top and does not change the value. Does your groupbox have set it to bottom explicitly?

fschinkel: The detailForm is created first and therefore _Form.\_init_ is executed before _Page.\_initDetailForm_.
_Form.\_init_ calls _Form.\_setRootGroupBox_ and this sets the _menuBarPosition_ to _'bottom'_ as the default-_displayHint_ of a Form is _'dialog'_. Afterwards _Page.\_initDetailForm_ sets the _displayHint_ of the detailForm to _'view'_.
This only happens for ScoutJS-detailForms that have not set their _displayHint_ to _'view'_, which should not be necessary as the Page handles it.

Possible solutions are:
- Every Form used as detailForm needs to set its _displayHint_ to _'view'_ before it is created
 -> _menuBarPosition = 'bottom'_ is possible but the form can no longer be used somewhere else as a dialog
 -> works for Js-Page with Js-DetailForm and Java-Page with Js-DetailForm
- The page will manipulate its _\_detailFormModel_ before the detailForm is created
 -> _menuBarPosition = 'bottom'_ is possible and the form can be used somewhere else as a dialog but using _Page.setDetailForm_ might give a different result than passing the form in the model
 -> works for Js-Page with Js-DetailForm but not for Java-Page with Js-DetailForm
- Explicitly set the _menuBarPosition_ in _Page.\_initDetailForm_
 -> _menuBarPosition = 'bottom'_ is not possible but the form can be used somewhere else as a dialog and using _Page.setDetailForm_ or passing the form in the model has the same result
 -> works for Js-Page with Js-DetailForm and Java-Page with Js-DetailForm

PR #259, opened 2022-07-15
Add ifSet method for LazyValue and FinalValue

matthiaso:
Similar to java.util.Optional

### DIFF
--- a/org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/util/FinalValueTest.java
+++ b/org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/util/FinalValueTest.java
@@ -18,6 +18,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 import org.eclipse.scout.rt.platform.exception.PlatformException;
+import org.eclipse.scout.rt.platform.holders.BooleanHolder;
 import org.eclipse.scout.rt.platform.util.Assertions.AssertionException;
 import org.junit.Assert;
 import org.junit.Test;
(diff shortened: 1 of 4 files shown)


org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/util/FinalValue.java
```diff
@@ -125,4 +126,10 @@ public boolean setIfAbsent(final Callable<VALUE> producer) {
   public boolean isSet() {
     return m_value.get() != NULL_VALUE;
   }
+
+  public void ifSet(Consumer<? super VALUE> action) {
```
paolobazzi, line 131: The actual implementation reads the value twice, the value could be changed in the meantime. Unfortunately there is no atomic ifSet method on AtomicReference. Therefore I would suggest to cache the value in a variable in order to be consistent between get() and action-execution:

```
    VALUE value = get();
    if (value != NULL_VALUE) {
      action.accept(value);
    }
```

org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/util/LazyValueTest.java
```diff
@@ -105,4 +106,18 @@ public void testGetWithException() {
   private class P_MyRuntimException extends RuntimeException {
     private static final long serialVersionUID = 1L;
   }
+
+  @Test
+  public void testifSet() {
```
paolobazzi: `testIfSet`  uppercase i

org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/util/FinalValueTest.java
```diff
@@ -104,6 +105,20 @@ public void testNoDoubleInitializationTry() {
     Assert.assertNull(s.setIfAbsentAndGet("should not matter"));
   }
 
+  @Test
+  public void testifSet() {
```
paolobazzi: testIfSet (uppercase-i)

paolobazzi (approved):
Looks good, see minor comments

PR #262, opened 2022-07-20
SequenceBox: prevent suppressStatus-Changes for lastVisibleField

fschinkel:
Some fields (e.g. DateField) change their suppressStatus property during
editing, validation, etc. If such a field is the lastVisibleField of a
SequenceBox it causes duplicated errorStatuses.

306308

### DIFF
--- a/eclipse-scout-core/src/form/fields/sequencebox/SequenceBox.js
+++ b/eclipse-scout-core/src/form/fields/sequencebox/SequenceBox.js
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014-2018 BSI Business Systems Integration AG.
+ * Copyright (c) 2010-2022 BSI Business Systems Integration AG.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/src/form/fields/sequencebox/SequenceBox.js
```diff
@@ -193,6 +196,11 @@ export default class SequenceBox extends CompositeField {
     }
   }
 
+  _onLastVisibleFieldSuppressStatusChanged(e) {
```
cguglielmo: According to our naming convention it should be called XyChange instead of XyChanged

eclipse-scout-core/src/form/fields/sequencebox/SequenceBox.js
```diff
@@ -183,6 +185,7 @@ export default class SequenceBox extends CompositeField {
 
     // Make sure the last field won't display a status (but shows status CSS class)
     this._lastVisibleField.setSuppressStatus(FormField.SuppressStatus.ICON);
+    this._lastVisibleField.on('propertyChange:suppressStatus', this._lastVisibleFieldSuppressStatusHandler);
```
cguglielmo, line 195: The listener needs to be removed when the sequencebox is removed, otherwise it will create a memory leak.

I just noticed that there are already some listeners creating a leak:
- propertyChange listener added at _render
- destroy listener added at setParent of Widget caused by FieldStatus that is only removed and not destroyed

Can you please fix the propertyChange listener as well? I will have a look at the FieldStatus.

cguglielmo (changes requested):

cguglielmo (approved):

PR #273, opened 2022-08-05
Tree: improve TreeEventMerger and merge commonParentNode

fschinkel:
If multiple TreeEvents are merged into one, try to merge the
commonParentNode as well, because it is possible, that the nodes do no
longer contain their parentNode to determine a commonParent.

Example: all 3 childNodes A, B, C are removed from parent P individually
and the events are merged afterwards

 * remove A will produce following event: nodes_deleted[nodes: A,
   commonParent: P]
 * remove B -> nodes_deleted[nodes: B, commonParent: P]
 * remove C -> all_child_nodes_deleted[nodes: , commonParent: P]

Now the TreeEventBuffer will merge first two events. As A and B have
been removed their parent is null and it is not possible to determine
the commonParent just from A and B at this point. Therefore, also merge
the commonParent of the two events.
The merged event will look like this:
 * old: nodes_deleted[nodes: A, B, commonParent: ]
 * new: nodes_deleted[nodes: A, B, commonParent: P]

Afterwards the JsonTree tries to optimize the events as well. If there
is a nodes_deleted-event and the commonParent has no children left, it
will send an all_child_nodes_deleted-event instead. The old merged event
did no longer contain the commonParent and therefore this optimization
did not happen.
As the merged event will be sent after the
all_child_nodes_deleted-event triggered by the removal of C the tree in
the browser will remove A, B and C before processing the merged event.
If the merged event explicitly states something about A, B or C (like
the old one did), this will lead to errors because the nodes no longer
exist.

314646

### DIFF
--- a/org.eclipse.scout.rt.client.test/src/test/java/org/eclipse/scout/rt/client/ui/basic/tree/TreeEventBufferTest.java
+++ b/org.eclipse.scout.rt.client.test/src/test/java/org/eclipse/scout/rt/client/ui/basic/tree/TreeEventBufferTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010-2018 BSI Business Systems Integration AG.
+ * Copyright (c) 2010-2022 BSI Business Systems Integration AG.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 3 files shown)


cguglielmo (approved):
Looks good, thanks!

PR #275, opened 2022-08-10
Add FormField.addErrorStatus(String)

fschinkel:
Create an error status with severity ERROR if a String is passed.

286466

### DIFF
--- a/eclipse-scout-core/test/form/fields/FormFieldSpec.js
+++ b/eclipse-scout-core/test/form/fields/FormFieldSpec.js
@@ -1,9 +1,9 @@
 /*
- * Copyright (c) 2010-2020 BSI Business Systems Integration AG.
+ * Copyright (c) 2010-2022 BSI Business Systems Integration AG.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
+ * https://www.eclipse.org/legal/epl-v10.html
  *
  * Contributors:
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/test/form/fields/FormFieldSpec.js
```diff
@@ -393,6 +393,32 @@ describe('FormField', () => {
     expect(formField._isSuppressStatusIcon()).toBe(false);
   });
 
+  it('error status', () => {
```
cguglielmo: Please follow the existing patterns in this file (the test above is not a good example).
Create a describe block for a property or a function: describe('addErrorStatus')
and then a an it block describing what the expectation is: it('adds the given status to the list of error status
and for the new case: it('adds a status with severity error if the parameters is a string

-> Write complete sentences. Write in third person so describe and it create a sentence -> addErrorStatus adds the given status to the list of error status

Please remove the empty lines before the expectations (but keep the ones after the expectations).

eclipse-scout-core/test/form/fields/FormFieldSpec.js
```diff
@@ -393,6 +393,39 @@ describe('FormField', () => {
...
+    beforeEach(() => {
+      model = helper.createFieldModel();
+      formField = createFormField(model);
+    });
+
+    it('added by addErrorStatus with Status and String as argument', () => {
+      expect(formField.errorStatus).toEqual(null);
+
+      formField.addErrorStatus(Status.warning('warning'));
+
```
cguglielmo: There are still unnecessary blank lines between the setup and the assertion ;-)

-> remove blank lines before expect()

PR #283, opened 2022-08-25
Remove INode, use NodeId.current instead

matthiaso:
321468


matthiaso:
Run tests

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/clientnotification/ClientNotificationPoller.java
```diff
@@ -70,7 +70,7 @@ public void stop() {
       return;
     }
 
-    LOG.debug("Stopping client notification poller [clientNodeId={}].", INode.ID);
+    LOG.debug("Stopping client notification poller [clientNodeId={}].", NodeId.current());
```
paolobazzi, line 73: Use IIds.toString(NodeId.current()) to avoid logging "NodeId" class name, see org.eclipse.scout.rt.dataobject.id.AbstractId#toString

paolobazzi (approved):
Looks good

PR #284, opened 2022-08-25
Allow to render dates with 5-digit years without JS errors

awegmueller:
Note: it is still not allowed to _enter_ a 5-digit year via UI
(DateField), since this could lead to issues in the backend or in a
database.

327188

### DIFF
--- a/eclipse-scout-core/src/text/DateFormat.js
+++ b/eclipse-scout-core/src/text/DateFormat.js
@@ -86,11 +86,20 @@ export default class DateFormat {
     //   (e.g. MMMM) to short (e.g. M).
     this._patternDefinitions = [
       // --- Year ---
+      // This definition can _format_ dates with years with 4 or more digits.
+      // See: http://docs.oracle.com/javase/6/docs/api/java/text/SimpleDateFormat.html
+      //      chapter 'Date and Time Patterns', paragraph 'Year'
+      // We do not allow to _parse_ a date with 5 or more digits. We could allow that in a
+      // future release, but it could have an impact on backend logic, databases, etc.
       new DateFormatPatternDefinition({
         type: DateFormatPatternType.YEAR,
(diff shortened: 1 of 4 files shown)


eclipse-scout-core/src/util/dates.js
```diff
@@ -424,15 +432,15 @@ export function toJsonDateRange(range) {
  *
  * The format is as follows:
  *
- * [Year#4]-[Month#2]-[Day#2] [Hours#2]:[Minutes#2]:[Seconds#2].[Milliseconds#3][Z]
+ * [Year#4|5]-[Month#2]-[Day#2] [Hours#2]:[Minutes#2]:[Seconds#2].[Milliseconds#3][Z]
```
bschwarzent, line 437: There is an additional function timestamp() that has a similar comment --> check and adjust if necessary

eclipse-scout-core/src/text/DateFormat.js
```diff
@@ -86,11 +86,18 @@ export default class DateFormat {
     //   (e.g. MMMM) to short (e.g. M).
     this._patternDefinitions = [
       // --- Year ---
+      // This definition can _format_ dates with years with 4 or more digits.
+      // However: it is not allowed to _parse_ a date with 5 or more digits. In a future release we could allow
+      // that, but it could have an impact on backend logic, databases, etc.
```
bschwarzent: Can you add a link to more information about 5-digit years? (RFC, MDN, ISO standard, stackoverflow or similar)
I think this would help a future developer in understanding the reason for this code.

eclipse-scout-core/src/util/dates.js
awegmueller: I've noticed the function is never used. I'll mark the function as deprecated in 22.0 and remove it in release 23.1

PR #301, opened 2022-09-14
Hide menus with ValueFieldMenuType.NULL on SmartFields if value is present

samueleiben:
Previously, menus with menuType=ValueFieldMenuType.NULL were displayed on smartfields in some cases (e.g. when a value was set when opening a form, which used 0 as the smart field value, which was then interpreted as an empty field).

329690


fschinkel:
Solved by https://github.com/eclipse-scout/scout.rt/pull/302

PR #303, opened 2022-09-19
Clean filename when downloading binary resource

jwinkens:
Clean semicolon from filenames when downloading binary resources. Having a semicolon in filenames causes trouble with reverse proxies when opening the binary resources dynamic URI.


jwinkens:
@egloff will look at this

PR #317, opened 2022-10-11
Popup: add modal behaviour

tgygax:
The popup has now a new property "modal", which adds a glass pane and prevents any closing from mouse down (on anchor or outside) or other popups.

331648

### DIFF
--- a/eclipse-scout-core/src/popup/Popup.js
+++ b/eclipse-scout-core/src/popup/Popup.js
@@ -3,7 +3,7 @@
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
+ * https://www.eclipse.org/legal/epl-v10.html
  *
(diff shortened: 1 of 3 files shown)


eclipse-scout-core/src/popup/Popup.js
```diff
@@ -383,9 +390,97 @@ export default class Popup extends Widget {
...
+    }
+  }
+
+  _renderCloseOnOtherPopupOpen() {
+    if (this.closeOnOtherPopupOpen) {
+      if (!this._popupOpenHandler) {
+        this._popupOpenHandler = this._onPopupOpen.bind(this);
+        this.session.desktop.on('popupOpen', this._popupOpenHandler);
+      } else {
+        if (this._popupOpenHandler) {
```
cguglielmo: There is a bracket error: else should probably refer to the outer if not the inner one.
Maybe it would be clearer if the if statements were merged like in _renderWithGlassPane

cguglielmo (changes requested):
Looks great, thank you!

cguglielmo (approved):

PR #319, opened 2022-10-17
CalendarComponent: add new attributes and tooltip style

nsteger:
The calendar component itself can now be displayed with an icon in front of the subject.

The tooltip of the calendar component can display the following new attributes:
- subjectLabel
- subjectAppLink
- descriptionElements (list of objects with text, iconId and appLink)

Further, the tooltip has a new style.

306748

### DIFF
--- a/eclipse-scout-core/src/calendar/CalendarComponent.js
+++ b/eclipse-scout-core/src/calendar/CalendarComponent.js
@@ -1,14 +1,14 @@
 /*
- * Copyright (c) 2010-2019 BSI Business Systems Integration AG.
+ * Copyright (c) 2010-2022 BSI Business Systems Integration AG.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
+ * https://www.eclipse.org/legal/epl-v10.html
  *
  * Contributors:
  *     BSI Business Systems Integration AG - initial API and implementation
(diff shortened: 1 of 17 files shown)


org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/services/common/calendar/CalendarItemDescriptionElement.java
```diff
@@ -0,0 +1,86 @@
...
+  private String m_text;
+  private String m_iconId;
+  private String m_appLink;
+
+  public CalendarItemDescriptionElement() {
+    this(null, null, null);
+  }
+
+  public CalendarItemDescriptionElement(String text, String iconId, String appLink) {
+    super();
```
egloff: No need to call super()

eclipse-scout-core/src/calendar/CalendarComponent.js
```diff
@@ -306,19 +313,27 @@ export default class CalendarComponent extends Widget {
     return dates.format(date, this.session.locale, pattern);
   }
 
-  _description() {
-    let descParts = [],
-      range = null,
+  _description(linkAllowed) {
+    let range = null,
       text = '',
```
egloff: - Maybe this variable should be called htmlText or html as it contains html code and not just text?
- Or, create a description html component and at the end use `return $description.outerHtml`?
- Or, this method may return a Scout widget (Label) or directly the html component?

PR #329, opened 2022-10-21
DateField: allowed dates are not enabled if date has a time

cguglielmo:
If the date (value of the date field) contains a time other than 00:00,
no dates can be picked.

Also make it work for Scout JS DateFields as well.
- sort and trunc allowed dates (picker needs the dates that way)
- make sure allowed dates validation does not happen on js side for
 remote fields because ui server does not know about validation errors
  -> error status would not be shown

332172

### DIFF
--- a/eclipse-scout-core/src/form/fields/datefield/DateField.js
+++ b/eclipse-scout-core/src/form/fields/datefield/DateField.js
@@ -16,6 +16,7 @@ export default class DateField extends ValueField {
   constructor() {
     super();
 
+    this.allowedDates = [];
     this.popup = null;
     this.autoDate = null;
     this.dateDisplayText = null;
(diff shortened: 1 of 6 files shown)


eclipse-scout-core/src/form/fields/datefield/DateField.js
```diff
@@ -458,15 +472,19 @@ export default class DateField extends ValueField {
...
-      this._setProperty('allowedDates', allowedDates);
-    } else {
-      this._setProperty('allowedDates', null);
-    }
+    let truncDates = [];
+    arrays.ensure(allowedDates).forEach(date => {
+      if (date) {
+        truncDates.push(dates.trunc(dates.ensure(date)));
+      }
+    });
```
fschinkel, line 485:
```suggestion
    let truncDates = arrays.ensure(allowedDates)
      .filter(date => date)
      .map(date => dates.ensure(date))
      .map(date => dates.trunc(date));
```

fschinkel (changes requested):

PR #337, opened 2022-10-31
Support multiple instances of BearerAuthAccessController

brunokoeferli:
Removing the ApplicationScoped annotation from ITokenVerifier and ITokenPrincipalProducer allows to have different instances of the same implementation. This is necessary in case of multiple BearerAuthAccessController instance with different configurations/states are required.

333399


brunokoeferli:
see also the changes in the migration guide: https://github.com/bsi-software/org.eclipse.scout.docs/pull/119

stephan-merkli (approved):

PR #354, opened 2022-11-23
Data object serialization: Allow nested DoMapEntity structures

paolobazzi:
Nesting DoMapEntity structures like DoMapEntity<AnotherDoMapEntity> with a TypeVersion annotation caused errors during deserialization because the _typeVersion attribute was tried to be parsed as data object instead of a plain string. This change forces the deserializer to treat the _typeVersion attribute as fixed string independent of the containing entity.

330574


paolobazzi:
Please run tests

PR #356, opened 2022-11-24
UriUtility#getQueryParameters: fix handling of '=' in an URI query pa…

stephan-merkli:
…rameter value

A '=' in query parameter resulted in the character itself and all following characters to be ignored.

335458


stephan-merkli:
Invalid change: an URL must not contain a =, parameters must already be provided encoded

PR #367, opened 2022-12-01
Disable automatic LogbackServletContainerInitializer

matthiaso:
Logback provides an automatic LogbackServletContainerInitializer which registers LogbackServletContextListener to stop logback during web application context stop. However this listener may be run to early as we cannot control at what time during context stop this listener is run,
 this may lead to logging messages which are not logged anymore.

Solution: Stop logback a little bit later during context stop using a platform listener and disable LogbackServletContainerInitializer.

LogbackServletContainerInitializer may be disabled by adding

<context-param>
  <param-name>logbackDisableServletContainerInitializer</param-name>
  <param-value>true</param-value>
</context-param>

in the web.xml; the new LoggerShutdownPlatformListener is automatically run during platform stop as late as possible (using a high order).

331773

### DIFF
--- a/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/logger/LoggerShutdownPlatformListener.java
+++ b/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/logger/LoggerShutdownPlatformListener.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2010-2022 BSI Business Systems Integration AG.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     BSI Business Systems Integration AG - initial API and implementation
+ */
+package org.eclipse.scout.rt.platform.logger;
+
+import org.eclipse.scout.rt.platform.BEANS;
+import org.eclipse.scout.rt.platform.IPlatform.State;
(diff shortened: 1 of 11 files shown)


org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/logger/LoggerShutdownPlatformListener.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+@Order(5950)
+public class LoggerShutdownPlatformListener implements IPlatformListener {
+
+  @Override
+  public void stateChanged(PlatformEvent event) {
+    if (event.getState() == State.PlatformStopped) {
+      ILoggerSupport loggerSupport = BEANS.get(ILoggerSupport.class);
+      if (loggerSupport != null) {
+        loggerSupport.shutdown();
```
paolobazzi, line 31: Maybe add a final log message similar to:

LoggerInstallPlatformListener:  LOG.info("Registered logger support {}", loggerSupport.getClass().getName());

paolobazzi (approved):
Looks good!

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/logger/LoggerShutdownPlatformListener.java
matthiaso: Actually logback logger support already adds the final log message (even more detailed); as the other loggers currently do not do any shutdown action anyways I would leave it as it is and leave it to the specific support class to decide if a final log message is necessary/what the final log message should include.

PR #371, opened 2022-12-02
Data object visitor: add support for custom visitors

stephan-merkli:
And add implementation for TypedId.

335051


stephan-merkli:
Jenkins, please run tests

PR #372, opened 2022-12-02
Data object migration: Additional unit tests for value migrations

stephan-merkli:
Unit tests documenting value migrations for DoList, DoSet and Map.

315763


stephan-merkli:
Jenkins, please run tests

PR #376, opened 2022-12-05
Disable automatic LogbackServletContainerInitializer

matthiaso:
Logback provides an automatic LogbackServletContainerInitializer which registers LogbackServletContextListener to stop logback during web application context stop. However this listener may be run to early as we cannot control at what time during context stop this listener is run,
 this may lead to logging messages which are not logged anymore.

Solution: Stop logback a little bit later during context stop using a platform listener and disable LogbackServletContainerInitializer.

LogbackServletContainerInitializer may be disabled by adding

<context-param>
  <param-name>logbackDisableServletContainerInitializer</param-name>
  <param-value>true</param-value>
</context-param>

in the web.xml; the new LoggerShutdownPlatformListener is automatically run during platform stop as late as possible (using a high order).

331773


matthiaso:
Run tests

matthiaso:
Please run tests

PR #378, opened 2022-12-05
Allow easy overriding if cell tooltip has html enabled

nsteger:
331054


eclipse-scout-core/src/table/Table.ts
```diff
@@ -901,7 +901,7 @@ export class Table extends Widget implements TableModel {
       parent: this,
       selector: '.table-cell',
       text: this._cellTooltipText.bind(this),
-      htmlEnabled: this._isAggregatedTooltip.bind(this),
+      htmlEnabled: this._isHtmlEnabledTooltip.bind(this),
```
cguglielmo: Please rename to _isCellTooltipHtmlEnabled

PR #397, opened 2022-12-19
Fix e-mail address internalization, only encode domain part of address

matthiaso:
267855


matthiaso:
Assume fixed in https://github.com/eclipse-scout/scout.rt/pull/344

PR #398, opened 2022-12-19
E-Mail addresses: Add test for long domain names

matthiaso:
267855

### DIFF
--- a/org.eclipse.scout.rt.mail.test/src/test/java/org/eclipse/scout/rt/mail/MailIDNConverterTest.java
+++ b/org.eclipse.scout.rt.mail.test/src/test/java/org/eclipse/scout/rt/mail/MailIDNConverterTest.java
@@ -10,7 +10,7 @@
  */
 package org.eclipse.scout.rt.mail;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.*;
 
 import org.eclipse.scout.rt.platform.BEANS;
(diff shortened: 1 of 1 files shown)


org.eclipse.scout.rt.mail.test/src/test/java/org/eclipse/scout/rt/mail/MailIDNConverterTest.java
```diff
@@ -69,4 +69,18 @@ public void testUnicodeConvertionWithSpecialChar() {
     // Assert
     assertEquals(emailAddress, convertedAddress);
   }
+
+  @Test
+  public void testLongDomainNames() {
+    // sample e-mail address with long domain name
+    String validLongEmail = "nobody@abcdefghijklmnopqrstuvwxyabcdefghijklmnopqrstuvwxyabcdefghijklm.de";
+    assertEquals(validLongEmail, BEANS.get(MailIDNConverter.class).toASCII(validLongEmail));
```
stephan-merkli: local variable MailIDNConverter converter = BEANS.get(MailIDNConverter.class); and use 3x

matthiaso: Thanks, fixed in 339e2717ef0a655f0c48eeb8d88f93efef8cf20a

PR #412, opened 2022-12-29
Mobile: Disabled search form

mvilliger:
335242


mvilliger:
Currently the fix changes the color of the search form icon to look like disabled (same as in desktop-mode).
En alternative would be to completely hide the menu item (display: none) if disabled to save space.

PR #413, opened 2022-12-30
Tree: recompute paddings if node becomes visible on filter change

mvilliger:
Fixes the following issue:
1. Padding of a node is computed while visible and Tree is not in breadcrumb mode
2. Filter of the Tree is changed so that this node becomes invisible
3. Tree is changed to breadcrumb mode
4. Filter of the Tree is changed so that the node becomes visible again

Now the padding must be recomputed as the old one still contains the value for the non-breadcrumb mode while the tree now is in breadcrumb mode.

337579


mvilliger:
The current fix recomputes the padding for all nodes rendered. Maybe only the ones for the nodes in newlyShown should be computed again? But the number of nodes rendered in the view range should be quite small? So maybe it is ok like that?

PR #414, opened 2023-01-04
ContextMenuPopup: check if default is prevented before closing popup

bschwarzent:
By default, clicking a menu automatically closes the context menu popup. When "toggleAction" is true, this does not happen. This change also allows non-toggle menus to keep the context menu popup open by calling event.preventDefault() in the action handler.

### DIFF
--- a/eclipse-scout-core/test/menu/ContextMenuPopupSpec.ts
+++ b/eclipse-scout-core/test/menu/ContextMenuPopupSpec.ts
@@ -241,4 +241,55 @@ describe('ContextMenuPopup', () => {
 
   });
 
+  describe('preventDefault', () => {
+
+    it('closes the popup when default is not prevented', () => {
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/test/menu/ContextMenuPopupSpec.ts
```diff
@@ -242,4 +242,55 @@ describe('ContextMenuPopup', () => {
...
+        animateOpening: false,
+        animateRemoval: false
+      });
+
+      popup.open();
+      findClone(popup, menu1).doAction();
+      expect(menu1Clicked).toBe(true);
+      expect(popup.rendered).toBe(true); // <--
+
+      popup.destroy();
```
cguglielmo: Popups should actually be closed using popup.close()

PR #415, opened 2023-01-04
Table & TileGrid: apply sort to initial elements

mvilliger:
Elements (rows or tiles) passed to the init model of a Table or TileGrid should respect the sort order configured.

267919

### DIFF
--- a/eclipse-scout-core/src/table/Table.ts
+++ b/eclipse-scout-core/src/table/Table.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010-2022 BSI Business Systems Integration AG.
+ * Copyright (c) 2010-2023 BSI Business Systems Integration AG.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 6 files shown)


eclipse-scout-core/src/tile/TileGrid.ts
```diff
@@ -463,12 +463,18 @@ export class TileGrid extends Widget implements TileGridModel {
     this.comparator = comparator;
   }
 
-  sort() {
+  protected _sortAfterInit() {
+    if (!this.sort()) {
+      this.updateFilteredElements();
```
cguglielmo: Is it necessary to call updateFilteredElements here? updateFilteredElements does only filter if filteredElementsDirty is true anyway -> Can we call it in init right after _sortAfterInit?

eclipse-scout-core/test/tile/TileGridSpec.ts
```diff
@@ -486,6 +486,30 @@ describe('TileGrid', () => {
 
   describe('sort', () => {
 
+    it('works if rows are passed in model', () => {
```
cguglielmo: works if *tiles* are passed

eclipse-scout-core/src/table/Table.ts
```diff
@@ -353,6 +353,7 @@ export class Table extends Widget implements TableModel {
     this._group();
```
cguglielmo: Remove _group here and do it only for Scout classic in TableAdapter, so it won't be done twice on init?

PR #419, opened 2023-01-06
Features/cgu/23.1/default detail form

cguglielmo:


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/outline/AbstractOutline.java
```diff
@@ -757,7 +761,21 @@ private void handleActivePageChanged(IPage<?> deselectedPage, IPage<?> selectedP
...
+      previousForm.removeFormListener(m_defaultDetailFormDisposeListener);
+      m_defaultDetailFormDisposeListener = null;
+    }
     if (form != null) {
+      m_defaultDetailFormDisposeListener = e -> {
+        if (e.getType() == FormEvent.TYPE_CLOSED) {
+          setDefaultDetailForm(null);
+        }
+      };
+      form.addFormListener(m_defaultDetailFormDisposeListener);
```
mvilliger, line 778: Should we add the other options (like in Outline.js#_adaptDefaultDetailForm) here as well (e.g. modality)?

mvilliger (approved):

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/outline/AbstractOutline.java
cguglielmo: This should not be necessary, since the ui already does it. Also, in Java, modal is automatically set to false for views.

PR #425, opened 2023-01-09
HtmlField displayText must not be read from DOM

mvilliger:
The displayText is rendered to the DOM with some modification:
- App-Links are decorated
- Scrollbars are added if enabled & necessary

Therefore the DOM does not reflect the original displayText. If the html is read from the DOM to compute the displayText, these modifications are included. If acceptInput is called (e.g. after focusing the field and then switching to another field), the displayText is recomputed and then includes these modifications. Is the field rendered again, the scrollbars are duplicated but only one is functional.

Instead, the displayText should always return the property as the content of the HtmlField cannot be changed anyway by the user. Therefore, it is not necessary to read it form the current DOM.

334077

### DIFF
--- a/eclipse-scout-core/test/form/fields/htmlfield/HtmlFieldSpec.ts
+++ b/eclipse-scout-core/test/form/fields/htmlfield/HtmlFieldSpec.ts
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2010-2023 BSI Business Systems Integration AG.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v10.html
+ *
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/test/form/fields/htmlfield/HtmlFieldSpec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  afterEach(() => {
+    jasmine.clock().uninstall();
+    jasmine.Ajax.uninstall();
+  });
+
+  describe('acceptInput', () => {
+    it('does not change field value and displayText', () => {
+      let field = new HtmlField();
+      let model = helper.createFieldModel(HtmlField, null, {
```
cguglielmo: You don't have to create the model using the helper. Using scout.create(HtmlField, {
  parent: session.desktop,
  scrollBarEnabled: true
}
should be sufficient.

PR #428, opened 2023-01-13
Improved property reading

Knicksaas:
- Refactoring
- Returning null when no property or folded property can be determined

336956


seesack (approved):
As discussed

PR #430, opened 2023-01-13
Update license to EPL-2.0

mvilliger:


eclipse-scout-core/res/log4javascript-1.4.9/log4javascript.js
```diff
@@ -1,17 +1,11 @@
-/**
```
cguglielmo, line 1: This must not be changed to epl since it is a third party lib

PR #434, opened 2023-01-17
Add knownType for QualifiedIIdDeserializer

matthiaso:
There may be cases where a qualified id is used however where the expected type is already known, allow supplying a known id type. If a qualified id the supplied type is checked again the known type. If an unqualified id is supplied and the type is known it will be used (instead of an error which would be thrown otherwise).

323333


matthiaso:
Replaced by https://github.com/eclipse-scout/scout.rt/pull/454

PR #456, opened 2023-02-08
IdCodec#fromQualified should look only at first ':' symbol

ralph-steiner:
An IStringId might contain itself the ':' symbol.


paolobazzi:
Looks good

paolobazzi (approved):

PR #435, opened 2023-01-17
Create SMTP connection outside of synchronized block

matthiaso:
Reason: Connection creation may take quite a while; if multiple connections are requested at the same time the connection establishment was previously run one after another; with this change multiple connections may be established at the same time (still limited by the pool size limits).

329715

PR #457, opened 2023-02-09
Create SMTP connection outside of synchronized block

matthiaso:
Reason: Connection creation may take quite a while; if multiple connections are requested at the same time the connection establishment was previously run one after another; with this change multiple connections may be established at the same time (still limited by the pool size limits).

Also replace synchronisation mechanism with ReantrantLock as otherwise order of waiting threads would not be guaranteed FIFO, which may result in favoring shorter waiting threads and timing out longer waiting threads.

329715


matthiaso:
This PR replaces https://github.com/eclipse-scout/scout.rt/pull/435

matthiaso:
Jenkins go

matthiaso:
Please run tests

matthiaso:
Run tests

matthiaso:
Run tests

matthiaso:
Run tests

PR #460, opened 2023-02-16
Form: replace notificationCount by NotificationBadgeStatus

fschinkel:
Add a NotificationBadgeStatus which replaces the notificationCount on the form. The notification badge is rendered in the status area of a SimpleTab if the form is displayed as view.

325085

### DIFF
--- a/eclipse-scout-core/src/tabbox/SimpleTab.js
+++ b/eclipse-scout-core/src/tabbox/SimpleTab.js
@@ -32,7 +32,6 @@ export default class SimpleTab extends Widget {
     this.$subTitle = null;
     this.$iconContainer = null;
     this.$statusContainer = null;
-    this.$notificationCount = null;
     this.$statusIcons = [];
 
     this._statusContainerUsageCounter = 0;
@@ -62,7 +61,6 @@ export default class SimpleTab extends Widget {
     this.saveNeeded = (this.view ? this.view.saveNeeded : model.saveNeeded);
     this.saveNeededVisible = (this.view ? this.view.saveNeededVisible : model.saveNeededVisible);
     this.status = (this.view ? this.view.status : model.status);
-    this.notificationCount = (this.view ? this.view.notificationCount : model.notificationCount);
 
     if (this.view) {
       this._installViewListeners();
@@ -102,7 +100,6 @@ export default class SimpleTab extends Widget {
     this._renderClosable();
     this._renderSaveNeeded();
     this._renderStatus();
-    this._renderNotificationCount();
     this._renderSelected();
   }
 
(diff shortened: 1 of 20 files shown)


eclipse-scout-core/src/tabbox/SimpleTab.js
```diff
@@ -214,52 +211,40 @@ export default class SimpleTab extends Widget {
...
+    this.$statusIcons.forEach($statusIcon => $statusIcon.remove());
     this.$statusIcons = [];
 
     if (this.status) {
       this.status.asFlatList().forEach(status => {
-        if (!status || !status.iconId) {
+        if (!status || (!status.iconId && !(status instanceof NotificationBadgeStatus))) {
+          return;
+        }
+        if (status instanceof NotificationBadgeStatus) {
```
cguglielmo: Don't limit this to NotificationBadgeStatus. If there is a message, display it. To not break existing code, add display: none to css, as done with scout 11

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/JsonNotificationBadgeStatus.java
```diff
@@ -0,0 +1,30 @@
...
+    super(status);
+  }
+
+  @Override
+  public String getObjectType() {
+    return "NotificationBadgeStatus";
+  }
+
+  @Override
+  public NotificationBadgeStatus getStatus() {
```
cguglielmo: Why is this necessary?

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/form/AbstractForm.java
```diff
@@ -919,33 +923,30 @@ private MultiStatus ensureMultiStatus(IStatus s) {
...
+        .orElse(null);
   }
 
   @Override
-  public void addNotificationCount(int notificationCount) {
-    setNotificationCount(getNotificationCount() + notificationCount);
+  public void setNotificationBadgeText(String notificationBadgeText) {
+    getNotificationBadgeStatus().ifPresent(this::removeStatus);
+    if (notificationBadgeText != null) {
+      addStatus(new NotificationBadgeStatus(notificationBadgeText)
```
cguglielmo, line 936: Why not set the status code inside Notification Badge Status? I think the constant should also be moved to that class

eclipse-scout-core/src/tabbox/SimpleTab.less
```diff
@@ -84,8 +84,8 @@
           line-height: @font-size-small;
```
cguglielmo: line-height and min-width leads to misalignment in the scout theme. Maybe use the same style as for the chat menu?
Also, I think display: inline-block, margin-top and margin-bottom have no effect and can be removed, or am I wrong?
Edit: min-width is fine, but maybe we need to adjust it if we change line-height and or paddings

PR #462, opened 2023-02-24
File chooser widgets: Improve error message when exceeding file size

buehlmic:
When a file was uploaded which exceeded the configured file size limit, the resulting error message was always referencing MB (even when the limit was very small). This change improves the error message such that it uses B, KB and GB where appropriate.

The logic to generate such error messages has been refactored into a utility method to allow for a consistent access pattern.

342855


eclipse-scout-core/src/util/files.js
```diff
@@ -48,7 +48,23 @@ export function fileListToArray(fileList) {
   return files;
 }
 
+export function getErrorMessageMaximumUploadSizeExceeded(session, maximumUploadSize) {
+  if (maximumUploadSize < 1024) {
+    return session.text('ui.FileSizeLimit', maximumUploadSize, 'B');
```
awegmueller: I think 'Bytes' would be much clearer than 'B'. 

KB, MB and GB are often used abbreviations, but B not so much.

eclipse-scout-core/src/util/files.js
```diff
@@ -48,7 +48,23 @@ export function fileListToArray(fileList) {
...
+  }
+  let maximumUploadSizeKB = maximumUploadSize / 1024;
+  if (maximumUploadSizeKB < 1024) {
+    return session.text('ui.FileSizeLimit', maximumUploadSizeKB, 'KB');
```
awegmueller: Should you use the Math.ceil/floor function to show a less precise, but prettier number string in the UI? (all three occurrences)

eclipse-scout-core/src/filechooser/FileChooser.js
```diff
@@ -294,7 +294,7 @@ export default class FileChooser extends Widget {
       this.fileInput.validateMaximumUploadSize(files);
     } catch (errorMessage) {
       MessageBoxes.createOk(this)
-        .withHeader(this.session.text('ui.FileSizeLimitTitle'))
```
awegmueller, line 297: Do you really want to change the header/title? If yes, is the text-key 'FileSizeLimitTitle' still used?

Maybe you'd better leave header/title untouched and add the result from getErrorMessageMaximumUploadSizeExceeded to the body/errorMessage?

eclipse-scout-core/src/util/files.js
```diff
@@ -48,7 +48,28 @@ export function fileListToArray(fileList) {
...
+    return session.text('ui.FileSizeLimit', roundToTwoDecimals(maximumUploadSizeKB), 'KB');
+  }
+  let maximumUploadSizeMB = maximumUploadSizeKB / 1024;
+  if (maximumUploadSizeMB < 1024) {
+    return session.text('ui.FileSizeLimit', roundToTwoDecimals(maximumUploadSizeMB), 'MB');
+  }
+  return session.text('ui.FileSizeLimit', roundToTwoDecimals(maximumUploadSizeMB / 1024), 'GB');
+}
+
+function roundToTwoDecimals(number) {
```
awegmueller: I'd suggest you use Scout's numbers-Utility #round method. As the JS-Doc there says: <<Numbers should not be rounded with the built-in Number.toFixed() method, since it behaves differently on different browsers>>

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiTextContributor.java
```diff
@@ -198,6 +198,7 @@ public void contributeUiTextKeys(Set<String> textKeys) {
         "ui.All",
         "ui.CloseAllTabs",
         "ui.CloseOtherTabs",
-        "ui.Done"));
+        "ui.Done",
+        "ui.Bytes"));
```
awegmueller, line 202: Not sure if we ever rename "Bytes" to something else, but Ok ;-)

PR #469, opened 2023-03-01
TableEventBuffer: Adjust merger to support updated columns

gabrielstalder:
- Fixes the issue that when merging table events, the field updated columns of tables is ignored. Until now, the last event wins.
- Adds corresponding tests including a simple performance measure

339745

### DIFF
--- a/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/TableEventBuffer.java
+++ b/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/TableEventBuffer.java
@@ -429,7 +429,8 @@ public class TableEventBuffer extends AbstractEventBuffer<TableEvent> {
         && ObjectUtility.equals(event1.getDragObject(), event2.getDragObject())
         && ObjectUtility.equals(event1.getDropObject(), event2.getDropObject())
         && ObjectUtility.equals(event1.getCopyObject(), event2.getCopyObject())
-        && ObjectUtility.equals(event1.getColumns(), event2.getColumns());
+        && ObjectUtility.equals(event1.getColumns(), event2.getColumns())
+        && ObjectUtility.equals(event1.getUpdatedColumns(), event2.getUpdatedColumns());
     return identical;
   }
 
@@ -544,9 +545,11 @@ public class TableEventBuffer extends AbstractEventBuffer<TableEvent> {
     private final Set<IColumn<?>> m_targetColumnSet;
     private final List<ITableRow> m_targetRows;
     private final Set<ITableRow> m_targetRowSet;
+    private final Map<ITableRow, Set<IColumn<?>>> m_targetUpdatedColumns;
 
     private List<IColumn<?>> m_mergedColumns;
     private List<ITableRow> m_mergedRows;
+    private boolean m_completed = false;
 
     public TableEventMerger(TableEvent targetEvent) {
       assertNotNull(targetEvent, "targetEvent must not be null");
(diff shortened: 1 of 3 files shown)


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/TableEventBuffer.java
```diff
@@ -585,6 +590,12 @@ public void complete() {
       m_mergedRows.addAll(m_targetRows);
       m_targetEvent.setRows(m_mergedRows);
       m_mergedRows = null;
+
+      m_mergedUpdatedColumns.putAll(m_targetUpdatedColumns);
```
bschwarzent: Is this necessary? The map of updatedColumns on the target event will not be replace entirely (as with the other two lists), but only agumented.

If not necessary, do we even need the m_targetUpdatedColumns member?

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/TableEventBuffer.java
```diff
@@ -596,6 +607,13 @@ protected <TYPE> void mergeCollections(Collection<TYPE> source, List<TYPE> targe
       source.removeIf(sourceElement -> !targetSet.add(sourceElement));
       target.addAll(0, source);
     }
+
+    /**
+     * Merge maps, such that, if an element is in both maps, only the one of the second map (later event) is kept.
```
bschwarzent: Is this correct? Only the keys seem to be merged, the values are being overwritten. If there are two different events for the same row, but with different updated columns, shouldn't both columns be in the resulting map afterwards? 

Add test case for this in unit test.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/TableEventBuffer.java
gabrielstalder: removed `m_mergedUpdatedColumns`

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/TableEventBuffer.java
gabrielstalder: Fixed and added test

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/TableEventBuffer.java
```diff
@@ -544,9 +545,11 @@ protected static class TableEventMerger {
     private final Set<IColumn<?>> m_targetColumnSet;
     private final List<ITableRow> m_targetRows;
     private final Set<ITableRow> m_targetRowSet;
+    private final Map<ITableRow, Set<IColumn<?>>> m_targetUpdatedColumns;
 
     private List<IColumn<?>> m_mergedColumns;
     private List<ITableRow> m_mergedRows;
+    private boolean m_isUpdatedColumnsMerged = false;
```
bschwarzent: Necessary? The checks for m_mergedColumns/m_mergedRows in complete() already detect multiple invocations. The last "if" clause will never be evaluated.

If it should still remain, m_updatedColumnsMerged would be a better name (the "is" prefix is usually reserved for getters).

gabrielstalder: True. I would also suggest to remove `m_mergedRows == null` as this will never be evaluated as well. Or am I missing something?

bschwarzent (approved):

PR #473, opened 2023-03-03
SmartField: lookup-by-key: improve error handling

buehlmic:
If a lookup-by-key doesn't return a result or an error was thrown when trying to resolve the key, a validation status is set on the SmartField. This allows the user to differentiate whether the value set on the SmartField is invalid or doesn't exist.

329358


eclipse-scout-core/src/form/fields/smartfield/SmartField.ts
```diff
@@ -692,14 +692,35 @@ export class SmartField<TValue> extends ValueField<TValue> implements SmartField
...
+  }
+
+  protected _invalidKeyLookup(key : TValue) : string {
+    // lookup call didn't return a result. Maybe the data belonging to the key has been deleted.
+    let invalidValueMessage = this.session.text('ui.InvalidValue');
+    this.setLookupRow(scout.create((LookupRow<TValue>), {
+      key: key,
+      text: invalidValueMessage
+    }));
+    this.setErrorStatus(scout.create(ValidationFailedStatus, {
```
cguglielmo, line 719: Actually, we should only use addErrorStatus to not remove the existing status, but since the smart field does not make use of that feature anyway, it is ok for now. We can refactor all setErrorStatus calls at once later

eclipse-scout-core/test/form/fields/smartfield/SmartFieldSpec.ts
```diff
@@ -604,6 +607,52 @@ describe('SmartField', () => {
       expect(displayText).toBe('Baz');
     });
 
+    it('lookupByKey should set a validation warning status if there was an error during processing', () => {
+      let field = _createFieldWithNoDataKeyLookupCall(true);
+      field._formatValue(4 /* non-existing key */); // triggers lookup by key
```
cguglielmo: You should be able to use field.setValue instead of field._formatValue.

eclipse-scout-core/test/form/fields/smartfield/SmartFieldSpec.ts
```diff
@@ -604,6 +607,52 @@ describe('SmartField', () => {
...
+    it('lookupByKey should set a validation warning status if the lookup call returns no lookup row', () => {
+      let field = _createFieldWithNoDataKeyLookupCall(false);
+
+      field._formatValue(0); // triggers lookup by key
+      jasmine.clock().tick(500);
+      expect(field.errorStatus).toBeInstanceOf(ValidationFailedStatus);
+      expect(field.errorStatus.isWarning()).toBeTrue();
+    });
+
+    function _createFieldWithNoDataKeyLookupCall(rejectPromise: boolean): SpecSmartField {
```
cguglielmo: There is no need to prefix the function with _ because it cannot be called from outside anyway. We use that pattern only for class methods because they can be accessed.
If you use setValue above, you can return SmartField<number> instead of SpecSmartField. In that case you need to change the scout.create slighly: return scout.create((SmartField<number>)
Please remove the 3 casts , they are not necessary.

eclipse-scout-core/test/form/fields/smartfield/SmartFieldSpec.ts
```diff
@@ -604,6 +607,52 @@ describe('SmartField', () => {
...
+      override _getByKey(key: number): JQuery.Promise<LookupResult<number>> {
+        let deferred = $.Deferred();
+        setTimeout(() => {
+          if (this.rejectPromise) {
+            deferred.reject();
+          } else {
+            deferred.resolve({
+              queryBy: QueryBy.KEY,
+              lookupRows: []
+
```
cguglielmo: remove empty line

eclipse-scout-core/test/form/fields/smartfield/SmartFieldSpec.ts
```diff
@@ -604,6 +607,52 @@ describe('SmartField', () => {
...
+      let field = _createFieldWithNoDataKeyLookupCall(true);
+      field._formatValue(4 /* non-existing key */); // triggers lookup by key
+      jasmine.clock().tick(500);
+      expect(field.errorStatus).toBeInstanceOf(ValidationFailedStatus);
+      expect(field.errorStatus.isWarning()).toBeTrue();
+    });
+
+    it('lookupByKey should set a validation warning status if the lookup call returns no lookup row', () => {
+      let field = _createFieldWithNoDataKeyLookupCall(false);
+
```
cguglielmo: remove empty line

cguglielmo:
Jenkins, please run tests

PR #497, opened 2023-03-16
Delegate calls from VirtualDesktop to the real Desktop when available

mvilliger:
If a ClientRunContext is created (e.g. by using copyCurrent) the values
from the currently active context are copied. The values are the ones
from the moment of the RunContext creation.

E.g. if an async task is scheduled while the Desktop is constructing,
the VirtualDesktop is on the RunContext (as no other Desktop has been
created yet). This means the async task is executed with the
VirtualDesktop as well, even if it is actually executed after the real
Desktop has been created.

This means that the second task uses the VirtualDesktop which has
already been removed from the Session and has been replaced with the
real one. As the VirtualDesktop throws Exceptions for most of the
methods, this may lead to unexpected behavior.

Now the VirtualDesktop gets the real Desktop as soon as it is set to the
session and delegates all calls to it. Older RunContexts still having
the VirtualDesktop will then start to use the real one transparently and
immediately after it is available on the session and without changing
the RunContext or re-scheduling anything. After the real Desktop has
been set to the virtual one, the VirtualDesktop acts like a proxy
delegating all calls to the real one.

As calling methods on the Desktop requires to be in a ModelJob, the
Desktop cannot change during a single execution.

320004

### DIFF
--- a/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/internal/VirtualDesktop.java
+++ b/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/internal/VirtualDesktop.java
@@ -1,9 +1,9 @@
 /*
- * Copyright (c) 2010-2021 BSI Business Systems Integration AG.
+ * Copyright (c) 2010-2023 BSI Business Systems Integration AG.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
+ * https://www.eclipse.org/legal/epl-v10.html
  *
  * Contributors:
  *     BSI Business Systems Integration AG - initial API and implementation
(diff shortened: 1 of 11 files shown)


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/internal/VirtualDesktop.java
```diff
@@ -70,1001 +74,1094 @@
...
+   * @throws IllegalArgumentException
+   *           if the realDesktop is null or a {@link VirtualDesktop}.
+   * @throws IllegalStateException
+   *           if a different real desktop has already been set
+   * @throws WrongThreadException
+   *           if this method is not called from a model thread.
+   */
+  public void setRealDesktop(IDesktop realDesktop) {
+    if (realDesktop == null) {
+      throw new IllegalArgumentException("Real desktop must not be virtual");
```
andibur:
```suggestion
      throw new IllegalArgumentException("Real desktop must not be null");
```

andibur (approved):
Nice, Thanks

PR #499, opened 2023-03-17
GlobalTrustManager ignores configured KeyStore

paolobazzi:
The GlobalTrustManager replaces the default SSLContext with a custom implementation allowing to use certificates placed in a configurable folder. Before this fix, the replaced SSLContext was set up without loading available key managers. This fix introduces loading available key managers based on javax.net.ssl.* system properties according to the default implementation in SSLContextImpl.

265333

### DIFF
--- a/org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/GlobalTrustManager.java
+++ b/org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/GlobalTrustManager.java
@@ -10,13 +10,17 @@
  */
 package org.eclipse.scout.rt.server.commons;
 
+import java.io.FileInputStream;
 import java.io.FilenameFilter;
+import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 import java.security.GeneralSecurityException;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
+import java.security.NoSuchProviderException;
+import java.security.UnrecoverableKeyException;
 import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
@@ -25,6 +29,8 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
+import javax.net.ssl.KeyManager;
+import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.TrustManager;
 import javax.net.ssl.TrustManagerFactory;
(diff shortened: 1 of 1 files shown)


org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/GlobalTrustManager.java
```diff
@@ -77,16 +84,44 @@ public void installGlobalTrustManager(String protocol, String tmAlgorithm) {
...
+      try (FileInputStream keystoreInputStream = new FileInputStream(keystoreLocation)) {
+        keyStore.load(keystoreInputStream, keystorePassword);
+      }
+      KeyManagerFactory keyManagerFactory = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
+      keyManagerFactory.init(keyStore, keystorePassword);
+      return keyManagerFactory.getKeyManagers();
+    }
+    return null;
+  }
+
```
mvilliger, line 144: How about supporting keyStoreType "PKCS11" (token)? It requires "javax.net.ssl.keyStore" to have the value "NONE" and must not provide a password.

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/GlobalTrustManager.java
```diff
@@ -77,16 +84,44 @@ public void installGlobalTrustManager(String protocol, String tmAlgorithm) {
...
   }
 
+  /**
+   * Loads a list of {@link KeyManager} based on {@code javax.net.ssl.*} system properties.
+   *
+   * @see sun.security.ssl.SSLContextImpl.DefaultManagersHolder#getKeyManagers()
+   */
+  @SuppressWarnings("squid:S1168")
+  protected KeyManager[] loadKeyManagers() throws KeyStoreException, IOException, NoSuchAlgorithmException, CertificateException, UnrecoverableKeyException, NoSuchProviderException {
+    String keystoreLocation = System.getProperty("javax.net.ssl.keyStore");
```
mvilliger: Maybe use the ConfigUtility instead of System.getProperty? This would allow us specifying this value using config.properties or environment variables (which might be handy for containers)? What do you think?

paolobazzi: Changed to ConfigUtility

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/GlobalTrustManager.java
paolobazzi: Added

mvilliger (approved):

PR #496, opened 2023-03-16
Remove script and style contents while parsing html to plain text 343094

yaschwanden:

PR #501, opened 2023-03-17
Convert decimal numeric character reference to unicode 342160

yaschwanden:


egloff (commented):
Same as in #496:

- update comment (ticket number at the end)
- implement same logic in for JavaScript utility

egloff:
See #560

PR #510, opened 2023-03-22
Decrease memory footprint of AbstractContributionComposite

csk-bsi:
Remove m_contributionsByType which is only used by init methods.
Therefore, caching of those entries is not needed.

Only initialize m_contributionsByClass if contributions exists

### DIFF
--- a/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/AbstractContributionComposite.java
+++ b/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/AbstractContributionComposite.java
@@ -1,32 +1,31 @@
 /*
- * Copyright (c) 2010-2017 BSI Business Systems Integration AG.
+ * Copyright (c) 2010-2023 BSI Business Systems Integration AG.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
+ * https://www.eclipse.org/legal/epl-v10.html
  *
  * Contributors:
  *     BSI Business Systems Integration AG - initial API and implementation
  */
 package org.eclipse.scout.rt.shared.extension;
 
-import java.io.IOException;
-import java.io.ObjectInputStream;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.eclipse.scout.rt.platform.BEANS;
+import org.eclipse.scout.rt.platform.util.Assertions;
 import org.eclipse.scout.rt.platform.util.CollectionUtility;
 
 public abstract class AbstractContributionComposite implements IContributionOwner, Serializable {
 
   private static final long serialVersionUID = 1L;
 
-  private Map<Class<?> /* type of the contribution */, ArrayList<?>/* contribution instances */> m_contributionsByType;
   private Map<Class<?> /* concrete contribution class */, Object /* contribution instance */> m_contributionsByClass;
 
   protected AbstractContributionComposite() {
@@ -58,11 +57,10 @@ public abstract class AbstractContributionComposite implements IContributionOwne
       owner = o;
     }
     List<?> contributionsForMe = extensionRegistry.createContributionsFor(owner, type);
-    if (type == null) {
-      m_contributionsByType = new HashMap<>();
-      m_contributionsByClass = new HashMap<>(contributionsForMe.size());
-    }
     if (CollectionUtility.hasElements(contributionsForMe)) {
+      if (m_contributionsByClass == null) {
+        m_contributionsByClass = new HashMap<>(contributionsForMe.size(), 1.0f); // use fixed size and loadFactor
+      }
       for (Object contribution : contributionsForMe) {
         m_contributionsByClass.put(contribution.getClass(), contribution);
       }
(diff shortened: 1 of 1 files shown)


org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/AbstractContributionComposite.java
```diff
@@ -34,6 +34,8 @@ protected AbstractContributionComposite() {
   }
 
   protected AbstractContributionComposite(Object o, boolean useScope) {
+    m_contributionsByType = null;
```
andibur: explicit initialization of fields is not required. This and the following line could be dropped.

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/AbstractContributionComposite.java
```diff
@@ -59,19 +61,24 @@ private <T> void initContributionsMap(Object o, IInternalExtensionRegistry exten
     }
     List<?> contributionsForMe = extensionRegistry.createContributionsFor(owner, type);
     if (type == null) {
-      m_contributionsByType = new HashMap<>();
-      m_contributionsByClass = new HashMap<>(contributionsForMe.size());
+      m_contributionsByType = null;
+      m_contributionsByClass = null;
     }
```
andibur: It looks like `type == null` was a trigger to initialize this instance for the first time (see constructor and readOject). This is not necessary anymore because the maps could be null.

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/AbstractContributionComposite.java
andibur, line 1: The serialization hook `private void readObject(ObjectInputStream s) throws IOException, ClassNotFoundException {` is not working anymore, but is not necessary either (maps were not declared transitive. Hence they were part of the serialized form. It looks like this was done for backwards compatibility reasons). Remove the method.

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/AbstractContributionComposite.java
```diff
@@ -59,19 +61,24 @@ private <T> void initContributionsMap(Object o, IInternalExtensionRegistry exten
...
     if (CollectionUtility.hasElements(contributionsForMe)) {
+      m_contributionsByClass = new HashMap<>();
```
andibur:
```suggestion
      m_contributionsByClass = new HashMap<>(contributionsForMe.size());
```

I would also initialize the `m_contributionsByType` here as well, because the fact of having contributions suggests that they are requested later on, using `getContributionsByClass`.

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/AbstractContributionComposite.java
```diff
@@ -112,7 +125,12 @@ public <T> List<T> getContributionsByClass(Class<T> type) {
         }
       }
       contributionsOfType.trimToSize();
-      m_contributionsByType.put(type, contributionsOfType);
+      if (contributionsOfType.size() != 0) {
+        if (m_contributionsByType == null) {
```
andibur: Not necessary If the `m_contributionsByType` is initialized along with `m_contributionsByClass`. See comment on `initContributionsMap`.

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/AbstractContributionComposite.java
```diff
@@ -101,8 +111,11 @@ public <T> List<T> getContributionsByClass(Class<T> type) {
       throw new IllegalArgumentException("Contribution type class must be specified.");
     }
```
andibur, line 96: Performance:
```
if (m_contributionsByClass == null) {
  return CollectionUtility.emptyArrayList();
}
```

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/AbstractContributionComposite.java
```diff
@@ -101,8 +111,11 @@ public <T> List<T> getContributionsByClass(Class<T> type) {
...
 
-    ArrayList<T> contributionsOfType = (ArrayList<T>) m_contributionsByType.get(type);
-    if (contributionsOfType == null) {
+    ArrayList<T> contributionsOfType = null;
+    if (m_contributionsByType != null) {
+      contributionsOfType = (ArrayList<T>) m_contributionsByType.get(type);
+    }
+    if (contributionsOfType == null && m_contributionsByClass != null) {
```
andibur: If performance suggestion is applied:
```suggestion
    if (contributionsOfType == null) {
```

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/AbstractContributionComposite.java
```diff
@@ -59,19 +61,24 @@ private <T> void initContributionsMap(Object o, IInternalExtensionRegistry exten
...
     }
   }
 
   public <T> void resetContributionsByClass(Object o, Class<T> type) {
-    m_contributionsByType.remove(type);
-    m_contributionsByClass.values().removeIf(obj -> type.isAssignableFrom(obj.getClass()));
+    if (m_contributionsByType != null) {
+      m_contributionsByType.remove(type);
+    }
+    if (m_contributionsByClass != null) {
```
andibur, line 72: Only one null-check is necessary If `m_contributionsByType` and `m_contributionsByClass` are initialized together.

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/AbstractContributionComposite.java
```diff
@@ -59,19 +61,24 @@ private <T> void initContributionsMap(Object o, IInternalExtensionRegistry exten
...
       for (Object contribution : contributionsForMe) {
         m_contributionsByClass.put(contribution.getClass(), contribution);
       }
     }
   }
 
   public <T> void resetContributionsByClass(Object o, Class<T> type) {
-    m_contributionsByType.remove(type);
-    m_contributionsByClass.values().removeIf(obj -> type.isAssignableFrom(obj.getClass()));
+    if (m_contributionsByType != null) {
```
andibur: I would add `assertNotNull(type, "type is requried")`. Would have thorwn an NPE anyway (see type.isAssignableFrom...)

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/AbstractContributionComposite.java
```diff
@@ -58,63 +57,52 @@ private <T> void initContributionsMap(Object o, IInternalExtensionRegistry exten
...
-      contributionsOfType.trimToSize();
-      m_contributionsByType.put(type, contributionsOfType);
+    if (m_contributionsByClass == null) {
+      return new ArrayList<>(0);
     }
-    return CollectionUtility.arrayList(contributionsOfType);
+
+    return m_contributionsByClass.values()
+        .stream()
+        .filter(o -> type.isAssignableFrom(o.getClass()))
```
andibur:
```suggestion
        .filter(o -> type.isInstance(o))
```

or as method reference
```suggestion
        .filter(type::isInstance)
```

andibur (approved):
Thanks. One last little suggestion

PR #515, opened 2023-03-24
Remove synchronized calls

csk-bsi:
Methods are only accessed via single threaded model job.


andibur (approved):
Maybe you should document that accessing TreeNodes from within an arbitrary ClientJob (i.e. not the ModelJob) will now be unsafe.

PR #522, opened 2023-03-29
Decrease memory footprint for AbstractTreeNode

csk-bsi:
Remove extension and contributions
Lazy create child nodes
287775


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/tree/AbstractTreeNode.java
```diff
@@ -111,32 +97,23 @@ public abstract class AbstractTreeNode implements ITreeNode, ICellObserver, ICon
...
     m_visible = NamedBitMaskHelper.ALL_BITS_SET; // default visible
-    m_objectExtensions = new ObjectExtensions<>(this, this instanceof AbstractPage<?>);
     if (callInitializer) {
       callInitializer();
     }
   }
 
   protected void callInitializer() {
     if (!isInitialized()) {
-      interceptInitConfig();
```
andibur, line 139: `initConfig` of a TreeNode is not invoked anymore. I think it is better to create a new non-final method `initConfigInternal` that is just invoking `initConfig()`. `AbstractPage` should override `initConfigInternal` and invoke `AbstractPage.interceptInitConfig()`. Remove overridden method `AbstractPage.callInitializer()`.

The pattern with `initConfigInternal` is already used by `AbstractWidget`.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/tree/AbstractTreeNode.java
```diff
@@ -414,12 +341,7 @@ public final Cell getCellForUpdate() {
 
   @Override
   public final void decorateCell() {
-    try {
-      interceptDecorateCell(m_cell);
-    }
-    catch (Exception t) {
-      LOG.error("node {} {}", getClass(), getCell().getText(), t);
-    }
+    execDecorateCell(m_cell);
```
andibur, line 344: Why did you remove the exception-handling? It is not related to extensibility.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/tree/AbstractTreeNode.java
```diff
@@ -699,8 +621,6 @@ public List<IMenu> getMenus() {
 
   protected List<IMenu> lazyCreateAndInitializeMenus() {
     List<Class<? extends IMenu>> declaredMenus = getDeclaredMenus();
-    List<IMenu> contributedMenus = m_contributionHolder.getContributionsByClass(IMenu.class);
```
andibur, line 702: Menu contributions do not work anymore on pages as well. I think this is not intended. Create a new `contributeMenusInteral` method that returns an empty list on `AbstractTreeNode`, but that uses the `contributionHolder` on `AbstractPage`.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/tree/AbstractTreeNode.java
```diff
@@ -1106,66 +996,29 @@ public Object validateValue(ICell cell, Object value) {
...
-      catch (RuntimeException e) {
-        LOG.warn("Exception while disposing node.", e);
-      }
-      try {
-        interceptDispose();
-      }
-      catch (RuntimeException e) {
-        LOG.warn("Exception while disposing node.", e);
-      }
+      disposeInternal();
```
andibur, line 1012: Did you remove the outer try block by intention? If `disposeInternal` throws an Error, the disposing flag is not restored anymore.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/tree/AbstractTreeNode.java
```diff
@@ -1106,66 +996,29 @@ public Object validateValue(ICell cell, Object value) {
...
-   * any further chain elements.
-   */
-  protected static class LocalTreeNodeExtension<OWNER extends AbstractTreeNode> extends AbstractExtension<OWNER> implements ITreeNodeExtension<OWNER> {
-
-    public LocalTreeNodeExtension(OWNER owner) {
-      super(owner);
-    }
-
-    @Override
-    public void execDecorateCell(TreeNodeDecorateCellChain chain, Cell cell) {
```
andibur, line 1120: This three extension methods are currently not available on a PageExtension. I thing they should be pushed down.

andibur (changes requested):
Please check the notes. Currently, the change removes features from page extensions, instead of just removing extension support from tree nodes.

PR #525, opened 2023-03-31
Features/cgu/23.1/arche

cguglielmo:


scout-hellojs-app/src/main/resources/archetype-resources/__rootArtifactId__.ui/src/main/js/person/PersonTablePage.ts
```diff
@@ -10,14 +10,15 @@ export class PersonTablePage extends PageWithTable {
     return PersonTablePageModel();
   }
 
-  protected override _initDetailTable(table: Table) {
-    super._initDetailTable(table);
-    this._initListeners();
-  }
+  protected override _init(model: InitModelOf<this>) {
+    super._init(model);
```
mvilliger, line 14: The page is created early with the Desktop, right? But the Table later on first activation of the Page.
What happens if the dataChange listener fires before the page has been activated?
Is this.reloadPage() working then without a Table in the Page (have not tested it)?

cguglielmo: reload page can handle that case, so it should not be a problem

PR #536, opened 2023-04-14
JsForm: don't show form until loading is complete

cguglielmo:
If a form is loaded when it is already displayed, the filling of the fields is visible which looks like flickering.

326250

### DIFF
--- a/eclipse-scout-core/src/form/js/JsFormAdapter.ts
+++ b/eclipse-scout-core/src/form/js/JsFormAdapter.ts
@@ -37,8 +37,19 @@ export class JsFormAdapter extends FormAdapter {
   protected override _createWidget(model: FullModelOf<Form>): Form {
     let widget = super._createWidget(model) as Form;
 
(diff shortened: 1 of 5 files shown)


eclipse-scout-core/src/form/js/JsFormAdapter.ts
```diff
@@ -37,8 +37,16 @@ export class JsFormAdapter extends FormAdapter {
...
 
-    widget.showOnOpen = false;
-    widget.open();
+    if (!widget.showOnOpen) {
+      widget.open();
+    } else {
+      widget.blockRendering = true;
+      widget.open();
+      widget.whenPostLoad().then(() => {
+        widget.blockRendering = false;
```
fschinkel, line 46: check destroyed and showOnOpen (like in Form.open)

fschinkel (changes requested):

fschinkel (approved):

PR #540, opened 2023-04-19
Abstract Type version: add support for optional suffix

stephan-merkli:
Optional suffix starts with __ followed by any word characters. The suffix is only part of the class name, but not the resulting NamespaceVersion.

Examples:
- Scout_22_0_0__lorem
- Scout_22_0_0__ipsum_dolor

346252


LukasHuser:
Review: Looks good.

PR #542, opened 2023-04-20
Features/cgu/23.2/smartfield invalidkey

cguglielmo:


buehlmic:
The change looks good and works as expected.

buehlmic (approved):
The change looks good and works as expected.

PR #543, opened 2023-04-21
Add SearchMenu

fschinkel:
Validate the form and trigger a 'search' event if it is valid.

326221

### DIFF
--- a/eclipse-scout-core/src/form/Form.ts
+++ b/eclipse-scout-core/src/form/Form.ts
@@ -426,6 +426,15 @@ export class Form extends Widget implements FormModel, DisplayParent {
   }
 
   /**
+   * Validates the form.
+   *
+   * @returns a promise resolved with `true` if the form is valid, `false` otherwise.
+   */
+  validate(): JQuery.Promise<boolean> {
+    return this.lifecycle.validate();
+  }
+
+  /**
    * This function is called by the lifecycle, for instance when the 'ok' function is called.
    * The function is called every time the 'ok' function is called, which means it runs even when
    * there is not a single touched field. The function should be used to implement an overall validate
--- a/eclipse-scout-core/src/lifecycle/Lifecycle.ts
+++ b/eclipse-scout-core/src/lifecycle/Lifecycle.ts
@@ -266,6 +266,14 @@ export abstract class Lifecycle<TValidationResult> extends EventEmitter implemen
       });
   }
 
+  /**
+   * @returns a promise resolved with `true` if the {@link widget} is valid, `false` otherwise.
+   */
+  validate(): JQuery.Promise<boolean> {
+    return this._whenInvalid(this._validate)
+      .then(invalid => !invalid);
+  }
+
   protected _validate(): JQuery.Promise<Status> {
     let status = this._validateElements();
     if (status.isValid()) {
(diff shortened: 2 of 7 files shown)


eclipse-scout-core/src/form/Form.ts
```diff
@@ -425,6 +425,15 @@ export class Form extends Widget implements FormModel, DisplayParent {
     });
   }
 
+  /**
+   * Validates the form.
+   *
+   * @returns {JQuery.Promise} resolved with <code>true</code> if life cycle is valid, <code>false</code> otherwise.
```
cguglielmo: No need to add the type in js doc with TypeScript.
instead of 'code' you can use the backtick symbol, or omit it completely, we mostly don't mark true or false.
replace life cycle with form, the method validates the form, not the life cycle.

eclipse-scout-core/src/lifecycle/Lifecycle.ts
```diff
@@ -266,6 +266,14 @@ export abstract class Lifecycle<TValidationResult> extends EventEmitter implemen
       });
   }
 
+  /**
+   * @returns {JQuery.Promise} resolved with <code>true</code> if life cycle is valid, <code>false</code> otherwise.
```
cguglielmo: see comment at Form.ts, replace life cycle with {@link widget}

eclipse-scout-core/test/form/FormSpec.ts
```diff
@@ -1663,4 +1668,71 @@ describe('Form', () => {
...
+      mandatoryStringField = form.rootGroupBox.fields[0] as StringField;
+      mandatoryStringField.setMandatory(true);
+
+      numberField = helper.createField(NumberField);
+      numberField.setOwner(form.rootGroupBox);
+      form.rootGroupBox.insertField(numberField);
+    });
+
+    it('returns true if all fields are valid', done => {
+      mandatoryStringField.parseAndSetValue('whatever');
```
cguglielmo: Why don't you use setValue?

eclipse-scout-core/test/form/FormSpec.ts
```diff
@@ -1663,4 +1668,71 @@ describe('Form', () => {
...
+      numberField.parseAndSetValue('42');
+
+      form.validate()
```
cguglielmo: Maybe use new async await syntax to make it more compact?

eclipse-scout-core/test/form/FormSpec.ts
```diff
@@ -1663,4 +1668,71 @@ describe('Form', () => {
...
+        .catch(fail)
+        .then(done);
+    });
+
+    it('returns false if mandatory field is empty', done => {
+      jasmine.clock().install();
+
+      numberField.parseAndSetValue('42');
+
+      form.validate()
```
cguglielmo, line 1702: async await would be nice, but I don't see how we could do it. We needed to listen for a message box to be opened, but I don't think this is possible atm. Any ideas?

cguglielmo (changes requested):

eclipse-scout-core/test/form/FormSpec.ts
fschinkel: not possible

cguglielmo (approved):

PR #547, opened 2023-04-25
NumberField & NumberColumn: add fractionDigits

fschinkel:
Add a new property to round the value to a given number of fraction
digits. This gives the possibility to round the value to e.g. 5 fraction
digits while the format has only 3 fraction digits. In this example a
user input "12.3456789" will give the value "12.34568" and the
displayText "12.346".

333348

### DIFF
--- a/eclipse-scout-core/src/form/fields/integerfield/IntegerField.ts
+++ b/eclipse-scout-core/src/form/fields/integerfield/IntegerField.ts
@@ -7,9 +7,11 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-import {DecimalFormat, Locale, NumberField, objects} from '../../../index';
+import {DecimalFormat, Locale, NumberField} from '../../../index';
 
 export class IntegerField extends NumberField {
+
+  declare fractionDigits: 0;
 
   static MIN_VALUE = Number.MIN_SAFE_INTEGER;
   static MAX_VALUE = Number.MAX_SAFE_INTEGER;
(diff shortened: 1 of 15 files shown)


eclipse-scout-core/src/table/columns/NumberColumnModel.ts
```diff
@@ -13,6 +13,12 @@ export interface NumberColumnModel extends ColumnModel<number> {
   aggregationFunction?: NumberColumnAggregationFunction;
   backgroundEffect?: 'colorGradient1' | 'colorGradient2' | 'barChart';
   decimalFormat?: DecimalFormat | string | DecimalFormatOptions;
+  /**
+   * Passed to the editor.
```
cguglielmo: Please write more explicit, that it has no effect on the column itself but only passed to the cell editor field, if the cell is editable.

eclipse-scout-core/src/form/fields/integerfield/IntegerField.ts
```diff
@@ -24,11 +24,8 @@ export class IntegerField extends NumberField {
     return '#,##0';
   }
 
-  protected override _parseValue(displayText: string): number {
-    let result = super._parseValue(displayText);
-    if (objects.isNullOrUndefined(result)) {
-      return null;
-    }
-    return this.decimalFormat.round(result, false);
+  protected override _ensureValue(value: number | string): number {
```
cguglielmo: Why not set fractionDigits to 0 instead?

eclipse-scout-core/src/text/DecimalFormat.ts
```diff
@@ -243,16 +243,37 @@ export class DecimalFormat {
   }
 
   /**
-   * Convert to JS number format (remove groupingChar, replace decimalSeparatorChar with '.')
+   * Convert to JS number format:
+   * <ul>
+   *   <li> remove groupingChar and lenientGroupingChars
```
cguglielmo: Please use markdown syntax: (- instead of ul li)

eclipse-scout-core/src/text/DecimalFormat.ts
```diff
@@ -243,16 +243,37 @@ export class DecimalFormat {
   }
```
cguglielmo, line 240: Is the js doc on the header of this file still correct?

eclipse-scout-core/src/form/fields/numberfield/NumberFieldModel.ts
```diff
@@ -13,4 +13,14 @@ export interface NumberFieldModel extends BasicFieldModel<number, number | strin
   minValue?: number;
```
cguglielmo, line 20: Would be nice if you could add some jsdoc to the other properties as well to make it complete

PR #556, opened 2023-04-28
Resource view for calendar

Knicksaas:


Knicksaas:
Outdated, new is #861

PR #566, opened 2023-05-04
ScoutJS: add menus from parent table page to detailForm/detailTable

fschinkel:
Add single selection menus from the parent table page to the detailForm and detailTable of the current page.
Select corresponding row in parent table page when a page is selected.
Add excludePropertiesToClone to Widget.clone.
Add modelAdapter to JsForm.

330431

### DIFF
--- a/eclipse-scout-core/src/desktop/outline/Outline.ts
+++ b/eclipse-scout-core/src/desktop/outline/Outline.ts
@@ -453,6 +453,7 @@ export class Outline extends Tree implements DisplayParent, OutlineModel {
 
   protected override _setSelectedNodes(nodes: Page[], debounceSend?: boolean) {
     super._setSelectedNodes(nodes, debounceSend);
+    this.mediator.onPageSelected(this.selectedNode());
     // Needs to be done here so that tree.selectNodes() can restore scroll position correctly after the content has been updated
     this.updateDetailContent();
   }
--- a/eclipse-scout-core/src/desktop/outline/Outline.ts
+++ b/eclipse-scout-core/src/desktop/outline/Outline.ts
@@ -432,6 +432,7 @@ export class Outline extends Tree implements DisplayParent, OutlineModel {
 
   override selectNodes(nodes: Page | Page[], debounceSend?: boolean) {
     nodes = arrays.ensure(nodes);
+    nodes.splice(1);
     if (nodes.length > 0 && this.isNodeSelected(nodes[0])) {
       // Already selected, do nothing
       return;
--- a/eclipse-scout-core/src/desktop/outline/OutlineMediator.ts
+++ b/eclipse-scout-core/src/desktop/outline/OutlineMediator.ts
@@ -70,4 +70,17 @@ export class OutlineMediator {
   onTableFilter(event: Event<Table>, page: Page) {
     page.getOutline().filter();
   }
+
+  onPageSelected(selectedPage: Page) {
+    if (!selectedPage || !selectedPage.parentNode) {
+      return;
+    }
+
+    const table = selectedPage.parentNode.detailTable;
+    const row = selectedPage.row;
+    if (!table || !row || table !== row.getTable()) {
+      return;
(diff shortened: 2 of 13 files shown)


eclipse-scout-core/src/desktop/outline/pages/Page.ts
```diff
@@ -214,12 +216,73 @@ export class Page extends TreeNode implements PageModel {
...
+  protected _updateParentTablePageMenusForStaticMenuOwner(staticMenuOwner: StaticMenuOwner) {
+    if (!staticMenuOwner) {
+      return;
+    }
+
+    staticMenuOwner.off('propertyChange:staticMenus', this._staticMenuOwnerStaticMenusChangeHandler);
+
+    const originalMenus = staticMenuOwner.staticMenus || [],
+      parentTablePageMenus = this._computeParentTablePageMenus()
+        .map(menu => menu.clone(
```
cguglielmo: Please avoid these multiple variable assignments separated with a comma, it makes it less readable because it unnecessarily indents the code.

eclipse-scout-core/src/desktop/outline/Outline.ts
```diff
@@ -467,6 +469,10 @@ export class Outline extends Tree implements DisplayParent, OutlineModel {
     }
   }
 
+  protected _onNodesSelected(event: TreeNodesSelectedEvent) {
+    this.mediator.onPagesSelected(event);
```
cguglielmo: Why is this done here and not at _setSelectedNodes?

eclipse-scout-core/src/desktop/outline/OutlineMediator.ts
```diff
@@ -70,4 +70,40 @@ export class OutlineMediator {
...
+  onPagesSelected(event: TreeNodesSelectedEvent) {
+    const outline = event.source;
+    if (!(outline instanceof Outline)) {
+      return;
+    }
+    const selectedPages = outline.selectedNodes,
+      rowsByTable = new Map();
+
+    selectedPages.forEach(page => {
+      if (!page || !page.parentNode) {
```
cguglielmo: Can page be null? if outline.selectedNodes contains null pages, something else is wrong

eclipse-scout-core/src/desktop/outline/OutlineMediator.ts
```diff
@@ -70,4 +70,40 @@ export class OutlineMediator {
...
+        row = page.row;
+
+      if (!table || !row || table !== row.getTable()) {
+        return;
+      }
+
+      rowsByTable.set(table, [...(rowsByTable.get(table) || []), row]);
+    });
+
+    rowsByTable.forEach((rows, table) => table.selectRows(rows));
```
cguglielmo: This is not necessary because an outline cannot have more than one selected page. There is no need to write code that will never be used.

eclipse-scout-core/src/desktop/outline/pages/Page.ts
```diff
@@ -214,12 +216,73 @@ export class Page extends TreeNode implements PageModel {
...
+      .concat(parentTablePageMenus));
+
+    staticMenuOwner.on('propertyChange:staticMenus', this._staticMenuOwnerStaticMenusChangeHandler);
+  }
+
+  protected _onStaticMenuOwnerStaticMenusChange(event: Event<StaticMenuOwner>) {
+    this._updateParentTablePageMenusForStaticMenuOwner(event.source);
+  }
+
+  protected _computeParentTablePageMenus(): Menu[] {
```
cguglielmo, line 274: A table page that has a table page as parent should probably not take the menus as it is in Java. Maybe add a property so it can be disabled by PageWithTable.ts?

cguglielmo (commented):
Can you add a detail form to SamplePageWithNodes in the js widgets app so we can test it better?

cguglielmo (approved):

PR #568, opened 2023-05-05
REST multipart: avoid closing the input stream

stephan-merkli:
323842


stephan-merkli:
Using a different approach without the MessageBodyReader at all, not sure if this would work properly when the stream is maybe never closed.

PR #571, opened 2023-05-08
Allow assertions for abstract types

bschwarzent:
scout.assertInstance(obj, type) can check if "obj" is instanceof "type". This should also work if "type" is an abstract class. Because the constructor of an abstract class cannot be invoked, the signature type "new(...args) => any" does not allow abstract classes. By adding the keyword "abstract", both abstract and non-abstract constructor functions are accepted.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-2.html#abstract-construct-signatures


fschinkel (approved):

bschwarzent:
Already merged (ec3bbd8b4cd03a7e1f243d2e3d46fbaba3e2ca05)

PR #573, opened 2023-05-10
ScoutJS: do not prevent save when fields have INFO or OK status

fschinkel:
Only consider a validation result of the lifecycle invalid if the status is invalid:
- OK and INFO status are considered valid
- WARNING is considered invalid but can be ignored by the user and turned into OK
- ERROR is considered invalid

343274


eclipse-scout-core/src/form/fields/tablefield/TableField.ts
```diff
@@ -203,18 +203,24 @@ export class TableField extends FormField implements TableFieldModel {
...
+          } else if (ret.errorStatus && ret.errorStatus.severity > errorStatus.severity) {
+            errorStatus = ret.errorStatus;
+          }
           validByMandatory = validByMandatory && ret.validByMandatory;
-          return !(validByErrorStatus || validByMandatory);
+
+          return errorStatus && errorStatus.isError() && !validByMandatory;
         }
         return false;
       }, this);
```
mvilliger: these "this" are no longer required with the arrow-function

eclipse-scout-core/src/form/fields/tablefield/TableField.ts
```diff
@@ -203,18 +203,24 @@ export class TableField extends FormField implements TableFieldModel {
...
-          validByErrorStatus = validByErrorStatus && ret.validByErrorStatus;
+          if (!errorStatus) {
+            errorStatus = ret.errorStatus;
+          } else if (ret.errorStatus && ret.errorStatus.severity > errorStatus.severity) {
+            errorStatus = ret.errorStatus;
+          }
           validByMandatory = validByMandatory && ret.validByMandatory;
-          return !(validByErrorStatus || validByMandatory);
+
+          return errorStatus && errorStatus.isError() && !validByMandatory;
```
mvilliger, line 213: is this correct? shouldn't it be (errorStatus && errorStatus.isError()) || !validByMandatory?

eclipse-scout-core/src/form/lifecycle/FormLifecycle.ts
```diff
@@ -36,27 +37,24 @@ export class FormLifecycle<TValidationResult extends ValidationResult = Validati
...
       let result = field.getValidationResult();
       if (result.valid) {
         return result.visitResult;
       }
-      // error status has priority over mandatory
-      if (!result.validByErrorStatus) {
-        invalidFields.push(result);
-      } else if (!result.validByMandatory) {
-        missingFields.push(result);
+      // mandatory has priority over error status
```
mvilliger: why was this changed? if I set an error status to a field e.g. using a very descriptive text, why is the mandatory then more important?

eclipse-scout-core/src/form/fields/tablefield/TableField.ts
fschinkel: `!(validByErrorStatus || validByMandatory)`
=> `!((!errorStatus || !errorStatus.isError()) || validByMandatory)`
=> `!(!errorStatus || !errorStatus.isError()) && !validByMandatory`
=> `(!!errorStatus && !!errorStatus.isError()) && !validByMandatory`
=> `errorStatus && errorStatus.isError() && !validByMandatory`

eclipse-scout-core/src/form/lifecycle/FormLifecycle.ts
fschinkel: Discussed this with Claudio and we considered the following aspects:

1. an empty mandatory field results in an ERROR
2. a field with error status XYZ results in severity XYZ
3. the message of the error status is never shown in the validation result message box

Therefore an empty mandatory field with an additional error status needs to be added to

- `missingFields` if the error status has a severity less or equal than WARNING, otherwise the validation result can be ignored and the Form can be stored with an empty mandatory field
- `invalidFields` or `missingFields` if the error status is ERROR

The only difference with this change will be an empty mandatory field with error status ERROR being shown in the missing block and no longer in the invalid block, no information of the error status is lost as the message was never shown.
If one wants to keep the former priority it will result in more complicated code and therefore the result of our discussion was to change it.

mvilliger (approved):

PR #579, opened 2023-05-16
Deprecate unused method IFormHandler#getHandlerId

ralph-steiner:
@mvilliger Please check SDK if method is used somewhere there. Otherwise merge.


mvilliger (approved):
SDK Code does not use this method. Deprecation is fine for me.

PR #647, opened 2023-06-02
Add busy handling to Scout JS

mvilliger:
- Move the old busy handling for Scout classic from Session to Desktop
- Introduce BusySupport which cares about:
  - multiple nested busy calls
  - cancellation callbacks
  - delayed rendering
- Add automatic busy handling to form load and save

326179

### DIFF
--- a/eclipse-scout-core/src/form/Form.ts
+++ b/eclipse-scout-core/src/form/Form.ts
@@ -8,9 +8,9 @@
  * SPDX-License-Identifier: EPL-2.0
  */
 import {
-  AbortKeyStroke, App, Button, ButtonSystemType, DialogLayout, DisabledStyle, DisplayParent, DisplayViewId, EnumObject, ErrorHandler, Event, EventHandler, FileChooser, FileChooserController, FocusRule, FormController, FormEventMap,
-  FormGrid, FormInvalidEvent, FormLayout, FormLifecycle, FormModel, FormRevealInvalidFieldEvent, GlassPaneRenderer, GroupBox, HtmlComponent, InitModelOf, KeyStroke, KeyStrokeContext, MessageBox, MessageBoxController, MessageBoxes,
-  NotificationBadgeStatus, ObjectOrChildModel, Point, PopupWindow, PropertyChangeEvent, Rectangle, scout, Status, StatusOrModel, strings, tooltips, TreeVisitResult, ValidationResult, webstorage, Widget, WrappedFormField
+  AbortKeyStroke, App, BusyIndicatorOptions, Button, ButtonSystemType, DialogLayout, DisabledStyle, DisplayParent, DisplayViewId, EnumObject, ErrorHandler, Event, EventHandler, FileChooser, FileChooserController, FocusRule, FormController,
+  FormEventMap, FormGrid, FormInvalidEvent, FormLayout, FormLifecycle, FormModel, FormRevealInvalidFieldEvent, GlassPaneRenderer, GroupBox, HtmlComponent, InitModelOf, KeyStroke, KeyStrokeContext, MessageBox, MessageBoxController,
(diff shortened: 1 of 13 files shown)


eclipse-scout-core/src/form/Form.ts
```diff
@@ -363,7 +363,17 @@ export class Form extends Widget implements FormModel, DisplayParent {
     if (!allowReload && this.formLoaded) {
       return $.resolvedPromise();
     }
-    return this.lifecycle.load();
+    return this._withBusyHandlingDo(() => this.lifecycle.load());
+  }
+
+  protected _withBusyHandlingDo<T>(action: () => JQuery.Promise<T>): JQuery.Promise<T> {
```
cguglielmo: Do sounds like DataObject, maybe name it _runWithBusyHandling or just _withBusyHandling

eclipse-scout-core/src/session/BusyIndicator.ts
```diff
@@ -10,9 +10,29 @@
 import {Action, BoxButtons, BusyIndicatorEventMap, ClickActiveElementKeyStroke, CloseKeyStroke, Event, FocusRule, GlassPaneRenderer, InitModelOf, keys, KeyStrokeContext, scout, strings, Widget, WidgetModel} from '../index';
 
 export interface BusyIndicatorModel extends WidgetModel {
+  /**
+   * Specifies if the {@link BusyIndicator} is cancellable.
+   * if true, a cancel button is visible and the 'cancel' event may be fired on click.
```
cguglielmo: Start with Uppercase.

eclipse-scout-core/src/session/BusyIndicator.ts
```diff
@@ -10,9 +10,29 @@
...
+   * The default is true.
```
cguglielmo: Add empty line before last line to get a line break. Maybe also after first line.

cguglielmo: We normally write Default is instead of The default is.

eclipse-scout-core/src/session/BusySupport.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  isBusy(): boolean {
+    return this._busyCounter > 0;
+  }
+
+  /**
+   * Changes the busy state
+   *
+   * @param options0 A boolean indicating if busy or not to show the busy indicator with the default settings (as passed when creating this instance) and without a cancellation callback.
```
cguglielmo: name it options, this is what the user of the api sees. options0 is confusing

eclipse-scout-core/src/session/Session.ts
```diff
@@ -1308,68 +1304,21 @@ export class Session extends EventEmitter implements SessionModel, ModelAdapterL
...
-    }
-    busyIndicator.off('cancel');
-
-    // Set "canceling" state in busy indicator (after 100ms, would not look good otherwise)
-    setTimeout(() => {
-      busyIndicator.cancelled();
-    }, 100);
-
-    this._sendCancelRequest();
+  protected _setBusy(busy: boolean) {
```
cguglielmo: It is confusing if an underscore method calls the public method, normally it is the other way round for properties. Maybe remove the public method? Consumers (the form) can call setBusy of the desktop directly

mvilliger: I wanted to keep the Session.setBusy method because it already existed and is still used in various places. So one migration task can be omitted. Furthermore in various places writing this.session.setBusy seems to be better than this.session.desktop?.setBusy which includes the risk of a NPE.
Actually the busy methods on the Session have nothing to do with properties. Therefore I don't see any confusion here. To make that clear I renamed the protected method.

PR #662, opened 2023-06-23
Scout JS: add permission support

fschinkel:
Add a permission support to Scout JS that can be activated using the Apps bootstrap options. Permissions are loaded by rest calls to an endpoint provided by the org.eclipse.scout.rt.api-module. The org.eclipse.scout.rt.api.data-module provides data objects for IPermissions and IPermissionCollections and a mapping mechanism to create these data objects.
Add model to all ajax utility functions.

326169


eclipse-scout-core/src/security/AccessControl.ts
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+import {
+  ajax, AjaxError, ErrorHandler, Event, InitModelOf, ObjectModel, Permission, PermissionCollection, PermissionCollectionModel, PermissionCollectionType, PropertyChangeEvent, PropertyEventEmitter, PropertyEventMap, scout, SomeRequired
+} from '../index';
+
+export class AccessControl extends PropertyEventEmitter implements AccessControlModel {
+
```
bschwarzent: Convention: No empty line before "declare" block

eclipse-scout-core/src/security/AccessControl.ts
```diff
@@ -0,0 +1,124 @@
...
+    this.interval = scout.nvl(model.interval, this.interval);
+
+    this._sync();
+  }
+
+  setInterval(interval: number) {
+    this.setProperty('interval', interval);
+  }
+
+  protected _setInterval(interval: number) {
```
bschwarzent, line 49: Is this method necessary? (setProperty does this by default)

eclipse-scout-core/src/security/AccessControl.ts
```diff
@@ -0,0 +1,124 @@
...
+    this._setProperty('interval', interval);
+  }
+
+  protected _sync() {
+    this._loadPermissionCollection()
+      .catch((error: AjaxError) => {
```
bschwarzent, line 72: Does it need to be an AjaxError?

eclipse-scout-core/src/security/AccessControl.ts
```diff
@@ -0,0 +1,124 @@
...
+        // handle error and keep last permission collection (or none collection if no permission collection is present)
+        scout.create(ErrorHandler, {displayError: false}).handle(error);
+        return [this._permissionCollection || {type: PermissionCollectionType.NONE}, error] as [PermissionCollectionModel, AjaxError];
```
bschwarzent: Maybe just return null to indicate error? Could be easier instead of this complicated "tuple" construct

eclipse-scout-core/src/App.ts
```diff
@@ -60,6 +60,11 @@ export interface AppBootstrapOptions {
    *  URL pointing to a json resources containing codes that will be available through {@link codes}.
    */
   codesUrl?: string;
+
+  /**
+   * URL pointing to a json resource containing permission information used by {@link access}.
```
bschwarzent: Mention `PermissionCollectionModel` (probably use the same comment as in AccessControlModel )

eclipse-scout-core/src/security/AccessControl.ts
```diff
@@ -0,0 +1,124 @@
...
+  check(permission: Permission, quick?: boolean): boolean | JQuery.Promise<boolean>;
+  check(permission: Permission, quick?: boolean): boolean | JQuery.Promise<boolean> {
+    if (!this._permissionCollection) {
+      return quick ? false : $.resolvedPromise(false);
+    }
+    return this._permissionCollection.implies(permission, quick);
+  }
+}
+
+export interface AccessControlModel extends ObjectModel<AccessControl> {
```
bschwarzent, line 122: Do we put model and event map in the same file in Scout? Or is there no convention? The migration guide seems to suggest separate files.

org.eclipse.scout.rt.security.data/src/main/java/org/eclipse/scout/rt/security/data/PermissionDo.java
```diff
@@ -0,0 +1,74 @@
...
+import javax.annotation.Generated;
+
+import org.eclipse.scout.rt.dataobject.DoEntity;
+import org.eclipse.scout.rt.dataobject.DoValue;
+import org.eclipse.scout.rt.dataobject.TypeName;
+import org.eclipse.scout.rt.dataobject.TypeVersion;
+import org.eclipse.scout.rt.security.IPermission;
+
+/**
+ * <b>IMPORTANT</b>: Never persist this data object or add a {@link TypeVersion}.
```
bschwarzent: Is this comment necessary? This applies to all data objects without @TypeVersion and is a general rule.

org.eclipse.scout.rt.security.data/src/main/java/org/eclipse/scout/rt/security/data/PermissionDo.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+@TypeName("scout.Permission")
+public class PermissionDo extends DoEntity {
+
+  public PermissionDo() {
+    withObjectType("scout.Permission");
+  }
+
+  public DoValue<String> objectType() {
+    return doValue("objectType");
```
bschwarzent: Needs documentation

eclipse-scout-core/src/session/Session.ts
```diff
@@ -45,6 +45,7 @@ export class Session extends EventEmitter implements SessionModel, ModelAdapterL
   /** assigned by server on session startup (OWASP recommendation, see https://www.owasp.org/index.php/Cross-Site_Request_Forgery_%28CSRF%29_Prevention_Cheat_Sheet#General_Recommendation:_Synchronizer_Token_Pattern). */
   uiSessionId: string;
   clientSessionId: string;
+  userId: string;
```
bschwarzent: Ok, but why is this included in this change? Does this have anything to do with the permission checks?

org.eclipse.scout.rt.security.data/src/main/java/org/eclipse/scout/rt/security/data/ISecurityDoMapper.java
```diff
@@ -0,0 +1,43 @@
...
+
+  boolean accept(SOURCE source);
+
+  TARGET map(SOURCE source);
+
+  /**
+   * Look for the first (see {@link BEANS#all}) mapper implementing the given mapperClass that {@link #accept}s the
+   * source and execute it.
+   */
+  static <S, T extends IDoEntity> T toDo(S source, Class<? extends ISecurityDoMapper<S, T>> mapperClass) {
```
bschwarzent: Shouldn't this be a helper class? (Static methods cannot be overriden, but maybe it's not necessary here?)

eclipse-scout-core/src/security/AccessControl.ts
fschinkel: The following code needs to know whether needs to know whether the model is the result of the call or the error-case-fallback. This is not derivable from the model itself, therefore we need to pass the error as well.

org.eclipse.scout.rt.security.data/src/main/java/org/eclipse/scout/rt/security/data/PermissionDo.java
fschinkel: In addition to this general rule, a @TypeVersion MUST never be added, as these DOs need to support subclassing.

eclipse-scout-core/src/session/Session.ts
fschinkel: Same Features Request, but has nothing to do with the permission checks => separated into different commits

eclipse-scout-core/src/security/AccessControl.ts
fschinkel: Normally we put them in separate files, but there are also a couple of examples with very small models that are contained in the same file (e.g. ErrorHandler, Device, BusyIndicator, ...).

org.eclipse.scout.rt.security.data/src/main/java/org/eclipse/scout/rt/security/data/ISecurityDoMapper.java
fschinkel: For now this is not meant to be overridden.

eclipse-scout-core/src/ajax/ajax.ts
```diff
@@ -19,90 +19,96 @@ export const ajax = {
    * Performs an HTTP GET request.
    * @param [options] additional settings for the request.
    *        Since jQuery is used to perform the request, all the jQuery Ajax settings are accepted.
+   * @param [model] additional properties for the AjaxCall.
```
cguglielmo: Please always use @link when referencing classes

eclipse-scout-core/src/ajax/ajax.ts
```diff
@@ -131,77 +138,81 @@ export const ajax = {
...
     }, options);
-    return ajax.createCall(opts);
+    return ajax.createCall(opts, model);
   },
 
   /**
    * Prepares an Ajax call, but does not execute it yet.
    * @param [options] additional settings for the request.
-   *        Since jQuery is used to perform the request, all the jQuery Ajax settings are accepted.
+   *        Since jQuery is used to perform the request, all the jQuery Url Ajax settings are accepted.
```
cguglielmo: I would omit the word Url

eclipse-scout-core/src/ajax/ajax.ts
```diff
@@ -131,77 +138,81 @@ export const ajax = {
...
+   * @param [model] additional properties for the AjaxCall.
    * @returns the prepared Ajax call object. Execute it with the call() function.
    */
-  createCall(options?: JQuery.AjaxSettings): AjaxCall {
-    let opts = $.extend({}, {
+  createCall(options: JQuery.UrlAjaxSettings, model?: Optional<AjaxCallModel, 'ajaxOptions'>): AjaxCall {
+    const ajaxOptions = $.extend({}, {
       cache: false
     }, options);
+    const ajaxCallModel = $.extend(true, {}, model, {ajaxOptions});
```
cguglielmo: I think this should be the other way round: If model contains the ajaxOptions, it should win over the first parameter. This would be consistent with the above methods: if ajax settings contain the url, it wins over the parameter url.

eclipse-scout-core/src/ajax/AjaxCall.ts
```diff
@@ -111,7 +111,7 @@ export class AjaxCall extends Call implements AjaxCallModel {
 
 export interface AjaxCallModel extends CallModel {
   /**
-   * Options for the jquery ajax call. At least the {@link CallModel.url} is required.
+   * Options for the jquery ajax call. At least the {@link JQuery.UrlAjaxSettings.url} is required.
    */
-  ajaxOptions: JQuery.AjaxSettings;
+  ajaxOptions: JQuery.UrlAjaxSettings;
```
cguglielmo: After some thoughts I think it is better to make it optional here and required only for the init case -> create declare initModel with SomeRequired. This would be the same pattern as for Widget and other objects.

eclipse-scout-core/src/security/access.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * Quick check (only local checks) <code>permission</code> against granted permissions of the current user.
+   */
+  quickCheck(permission: string | Permission): boolean {
+    return check(permission, true);
+  },
+
+  /**
+   * Check <code>permission</code> against granted permissions of the current user.
```
cguglielmo: Please use markdown syntax where possible in JsDoc ->\`permission\` instead of &lt;code&gt;permission&lt;/code&gt;

eclipse-scout-core/src/security/Permission.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+    return quick ? false : $.resolvedPromise(false);
+  }
+
+  matches(permission: Permission): boolean {
+    return permission instanceof Permission && this.name === permission.name;
+  }
+
+  /**
+   * Precondition: <code>matches(permission)</code>
```
cguglielmo: see markdown comment

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/IDoMapper.java
```diff
@@ -0,0 +1,43 @@
...
+import org.eclipse.scout.rt.platform.ApplicationScoped;
+import org.eclipse.scout.rt.platform.BEANS;
+import org.eclipse.scout.rt.platform.util.Assertions;
+import org.eclipse.scout.rt.platform.util.Assertions.AssertionException;
+
+/**
+ * A mapper that creates a data object and applies values from an arbitrary source to it.
+ */
+@ApplicationScoped
+public interface IDoMapper<SOURCE, TARGET extends IDoEntity> {
```
paolobazzi: very similar class to org.eclipse.scout.rt.dataobject.mapping.IDoEntityMapping
-> consolidate or extend each other?

a mapper which maps only from javaclass -> data object could also implement the reverse operation using a `throw new UnsupportedOperationException()`

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/IDoMapper.java
```diff
@@ -0,0 +1,43 @@
...
+
+  boolean accept(SOURCE source);
```
paolobazzi: Add javadoc

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/IDoMapper.java
```diff
@@ -0,0 +1,43 @@
...
+
+  TARGET map(SOURCE source);
+
+  /**
+   * Look for the first (see {@link BEANS#all}) mapper implementing the given mapperClass that {@link #accept}s the
+   * source and execute it.
+   */
+  static <S, T extends IDoEntity> T toDo(S source, Class<? extends IDoMapper<S, ? extends T>> mapperClass) {
```
paolobazzi: move to abstract implementation class as public method, remove "static"
-> static method is not customizable by subclasses/projects
-> static methods are not often used in scout -> prefer app-scoped helper
-> usage similar to all other Scout beans:  BEANS.get(XYMapper.class).toDo(obj)

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/AbstractDoMapper.java
```diff
@@ -0,0 +1,61 @@
...
+import static org.eclipse.scout.rt.platform.util.TypeCastUtility.getGenericsParameterClass;
+
+import org.eclipse.scout.rt.dataobject.IDoEntity;
+import org.eclipse.scout.rt.platform.BEANS;
+
+/**
+ * Base implementation for {@link IDoMapper}. This mapper accepts a source if it is an instance of EXPLICIT_SOURCE and
+ * creates a target data object of the type EXPLICIT_TARGET. Subclasses can change this behaviour.
+ */
+public abstract class AbstractDoMapper<EXPLICIT_SOURCE extends SOURCE, EXPLICIT_TARGET extends TARGET, SOURCE, TARGET extends IDoEntity> implements IDoMapper<SOURCE, TARGET> {
```
paolobazzi: class is not directly related to api.data, move to `org.eclipse.scout.rt.dataobject.mapping` ?

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/security/PermissionDo.java
```diff
@@ -0,0 +1,77 @@
...
+import javax.annotation.Generated;
+
+import org.eclipse.scout.rt.dataobject.DoEntity;
+import org.eclipse.scout.rt.dataobject.DoValue;
+import org.eclipse.scout.rt.dataobject.TypeName;
+import org.eclipse.scout.rt.dataobject.TypeVersion;
+import org.eclipse.scout.rt.security.IPermission;
+
+/**
+ * <b>IMPORTANT</b>: Never persist this data object or add a {@link TypeVersion}.
```
paolobazzi: This comment is not really mandatory.... there are unit tests in place which takes care of this rule for DOs without type version and all DOs without TypeVersion shouldn't be persisted.

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/security/IPermissionDoMapper.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+package org.eclipse.scout.rt.api.data.security;
+
+import org.eclipse.scout.rt.api.data.IDoMapper;
+import org.eclipse.scout.rt.security.IPermission;
+
+public interface IPermissionDoMapper extends IDoMapper<IPermission, PermissionDo> {
+
+  static PermissionDo toDo(IPermission permission) {
+    return IDoMapper.toDo(permission, IPermissionDoMapper.class);
```
paolobazzi: Suggestion: Remove this interface and move this function to helper or AbstractPermissionCollectionDoMapper
--> avoid using static methods in framework code

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/security/PermissionCollectionType.java
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.scout.rt.api.data.security;
+
+public enum PermissionCollectionType {
```
paolobazzi: implement IEnum  (enforcing "static" names for enum type)

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/security/PermissionDo.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+@TypeName("scout.Permission")
+public class PermissionDo extends DoEntity {
+
+  public PermissionDo() {
+    withObjectType("Permission");
```
paolobazzi: Suggestion: Add constant for this:  protected static final String DEFAULT_PERMISSION_OBJECT_TYPE = "Permission";

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/AbstractDoMapper.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  protected EXPLICIT_TARGET createDo() {
+    return BEANS.get(getExplicitTargetClass());
+  }
+
+  @Override
+  public TARGET map(SOURCE source) {
+    EXPLICIT_TARGET doEntity = createDo();
+    if (getExplicitSourceClass().isInstance(source)) {
```
paolobazzi: use accept() here? or remove duplicated check and assume that caller already checked accept before invoking map?

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/security/PermissionCollectionDo.java
```diff
@@ -0,0 +1,122 @@
...
+   * *************************************************************************/
+
+  public static class PermissionCollectionDoMapper extends AbstractPermissionCollectionDoMapper<IPermissionCollection, PermissionCollectionDo> {
+    @Override
+    public void map(IPermissionCollection permissionCollection, PermissionCollectionDo permissionCollectionDo) {
+      permissionCollectionDo
+          .withPermissions(permissionCollection.stream()
+              .map(IPermissionDoMapper::toDo)
+              .filter(Objects::nonNull)
+              .collect(Collectors.groupingBy(PermissionDo::getName, Collectors.toCollection(HashSet::new))))
```
paolobazzi: use Collectors.toSet()? Or is explicit a HashSet required?

eclipse-scout-core/src/security/access.ts
```diff
@@ -0,0 +1,64 @@
...
+  check(permission: string | Permission): JQuery.Promise<boolean> {
+    return check(permission);
+  },
+
+  getPermissionsUrl(): string {
+    return accessControl && accessControl.permissionsUrl;
+  }
+};
+
+function check(permission: string | Permission, quick: true): boolean;
```
cguglielmo, line 54: maybe use conditional return type, but only if it makes it easier:
function check<TQuick extends boolean>(permission: string, quick?: TQuick): TQuick extends true ? boolean : JQuery.Promise<boolean>

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/security/PermissionCollectionDo.java
```diff
@@ -0,0 +1,122 @@
...
+import org.eclipse.scout.rt.security.NonePermissionCollection;
+
+/**
+ * <b>IMPORTANT</b>: Never persist this data object or add a {@link TypeVersion}.
+ */
+@TypeName("scout.PermissionCollection")
+public class PermissionCollectionDo extends DoEntity {
+
+  public PermissionCollectionDo() {
+    withObjectType("PermissionCollection");
```
paolobazzi: - are there any other object types or is this one fixed for PermissionCollectionDo?
- why is a ScoutJS-object type part of the PermissionDo? -> could the UI layer map DO-type-name to a specific ScoutJS class?

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/security/PermissionCollectionDo.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  public DoValue<String> objectType() {
```
paolobazzi, line 42: add javadoc, explaining objectType vs. type

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/security/PermissionDo.java
```diff
@@ -0,0 +1,77 @@
...
+import org.eclipse.scout.rt.dataobject.DoValue;
+import org.eclipse.scout.rt.dataobject.TypeName;
+import org.eclipse.scout.rt.dataobject.TypeVersion;
+import org.eclipse.scout.rt.security.IPermission;
+
+/**
+ * <b>IMPORTANT</b>: Never persist this data object or add a {@link TypeVersion}.
+ */
+@TypeName("scout.Permission")
+public class PermissionDo extends DoEntity {
```
paolobazzi, line 25: IPermission contains a level, would this be useful in DO-layer as well?

scout-hellojs-app/src/main/resources/archetype-resources/__rootArtifactId__.app/src/main/js/__simpleArtifactName__.ts
```diff
@@ -1,7 +1,9 @@
 import {App} from '@${simpleArtifactName}/ui';
+
 new App().init({
   bootstrap: {
-    textsUrl: 'texts.json'
+    textsUrl: 'texts.json',
+    permissionsUrl: 'api/permissions'
```
paolobazzi, line 6: convention over configuration? allow to customize URL but use a default (e.g. hardcoded resource url) if no specific URL is given?

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/security/PermissionDo.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public PermissionDo() {
+    withObjectType("Permission");
+  }
+
+  /**
+   * String used to identify the Scout JS object created by this object model.
+   */
+  public DoValue<String> objectType() {
```
paolobazzi, line 32: ScoutJS-specific type as part of API? Should the ScoutJS UI not be able to determine the correct JS class based on typename of DO? Actually DO subclass implementations seems to match 1:1 to a ScoutJS class

paolobazzi (commented):
Looks good, see comments for improvements.

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/security/PermissionDo.java
fschinkel: coming soon

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/security/PermissionCollectionType.java
```diff
@@ -0,0 +1,46 @@
...
+  PermissionCollectionType(String stringValue) {
+    m_stringValue = stringValue;
+  }
+
+  @Override
+  public String stringValue() {
+    return m_stringValue;
+  }
+
+  public static PermissionCollectionType resolve(String value) {
```
paolobazzi: resolve implementation may be omitted, as long as enum names and id-strings correspond
the possibility to add a resolve funtion is just in case of a later change in values which requires a on-the-fly migration by a custom resolve method implementation

org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/mapping/IToDoFunction.java
```diff
@@ -0,0 +1,26 @@
...
+import org.eclipse.scout.rt.dataobject.IDoEntity;
+import org.eclipse.scout.rt.platform.ApplicationScoped;
+
+/**
+ * A function that creates a data object and applies values from an arbitrary source to it.<br>
+ * This one way mapping mechanism might be useful when dealing with subclassed objects that need to be mapped to
+ * subclassed data objects. If there are different {@link IToDoFunction}s for different types of sources
+ * {@link ToDoFunctionHelper#toDo(Object, Class)} will find the first one that is able to handle a given source and
+ * return its result.
+ */
```
paolobazzi, line 23: maybee add ```@see ToDoFunctionHelper```

org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/mapping/IToDoFunction.java
```diff
@@ -0,0 +1,26 @@
...
+@ApplicationScoped
+public interface IToDoFunction<SOURCE, TARGET extends IDoEntity> extends Function<SOURCE, TARGET> {
```
paolobazzi, line 25: Naming sounds like a TODO for me ;)  -> maybe use the full name IToDataObjectFunction here? This naming pattern is used elsewhere in case when 'Do` is a bit to short: IDataObjectXY

paolobazzi (approved):

fschinkel:
run tests

fschinkel:
run tests

fschinkel:
run tests

PR #664, opened 2023-07-06
CLI: ignore warnings from esbuild minifier

cguglielmo:
5.3.7 from terser-webpack-plugin fixed esbuild warning output: https://github.com/webpack-contrib/terser-webpack-plugin/releases/tag/v5.3.7

-> the check for [object Object] can be removed.
-> since the warnings from esbuild are not of interest they need to be
 ignored now

Also don't print empty warning objects.
hasWarnings() may return true even if info.warnings is empty. Maybe happens because some warnings are ignored.


mvilliger (approved):

cguglielmo:
Jenkins, run tests

PR #665, opened 2023-07-06
Improve some types

cguglielmo:
Optional may be confusing for Java devs.
Abortable should not implement closable by default so it can be reused better.
Interfaces seem to be the better choice for these kind of types.

334151


mvilliger (approved):
Looks good

PR #671, opened 2023-07-10
TileGrid: add generic

cguglielmo:
If placeholders are enabled, the tiles may also contain placeholder tiles. The user cannot interact with these tiles and the developer working with the tile grid should not either -> Placeholder tiles are internal and should not be exposed.

- TileGrid.tiles and TileGrid.filteredTiles don't return the placeholder tiles anymore
- TileGrid.selectTiles now filter the placeholder tiles, so they cannot be selected anymore.

334151

### DIFF
--- a/eclipse-scout-core/src/tile/TileGrid.ts
+++ b/eclipse-scout-core/src/tile/TileGrid.ts
@@ -21,21 +21,20 @@ import $ from 'jquery';
  */
 const TILE_SELECTOR = '> .tile';
 
-export class TileGrid extends Widget implements TileGridModel {
+export class TileGrid<TTile extends Tile = Tile> extends Widget implements TileGridModel {
   declare model: TileGridModel;
   declare eventMap: TileGridEventMap;
-  declare self: TileGrid;
+  declare self: TileGrid<TTile>;
   declare logicalGrid: AbstractGrid;
 
   animateTileRemoval: boolean;
   animateTileInsertion: boolean;
-  comparator: Comparator<Tile>;
+  comparator: Comparator<TTile>;
   contextMenu: ContextMenuPopup;
   empty: boolean;
-  filters: Filter<Tile>[];
-  filteredTiles: Tile[];
+  filters: Filter<TTile>[];
   filteredElementsDirty: boolean;
-  focusedTile: Tile;
+  focusedTile: TTile;
   // GridColumnCount will be modified by the layout, prefGridColumnCount remains unchanged
   gridColumnCount: number;
   prefGridColumnCount: number;
@@ -44,12 +43,11 @@ export class TileGrid extends Widget implements TileGridModel {
   multiSelect: boolean;
   renderAnimationEnabled: boolean;
   selectable: boolean;
-  selectedTiles: Tile[];
+  selectedTiles: TTile[];
   selectionHandler: TileGridSelectionHandler;
   scrollable: boolean;
   startupAnimationDone: boolean;
   startupAnimationEnabled: boolean;
-  tiles: Tile[];
   tileRemovalPendingCount: number;
   viewRangeSize: number;
   viewRangeRendered: Range;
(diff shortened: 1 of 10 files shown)


eclipse-scout-core/src/tile/TileGrid.ts
```diff
@@ -252,71 +253,83 @@ export class TileGrid extends Widget implements TileGridModel {
...
+  deleteTile(tile: TTile) {
     this.deleteTiles([tile]);
   }
 
-  deleteTiles(tilesToDelete: Tile | Tile[], appendPlaceholders?: boolean) {
+  deleteTiles(tilesToDelete: TTile | (TTile)[]) {
+    this.deleteTilesInternal(tilesToDelete);
+  }
+
+  protected deleteTilesInternal(tilesToDelete: (TTile | PlaceholderTile) | (TTile | PlaceholderTile)[], appendPlaceholders?: boolean) {
```
mvilliger: Is the first parenthesis necessary?
"TTile | PlaceholderTile | (TTile | PlaceholderTile)[]" looks a bit cleaner

eclipse-scout-core/src/tile/TileGrid.ts
```diff
@@ -252,71 +253,83 @@ export class TileGrid extends Widget implements TileGridModel {
...
+  insertTile(tile: ObjectOrChildModel<TTile>) {
     this.insertTiles([tile]);
   }
 
-  insertTiles(tilesToInsert: ObjectOrChildModel<Tile> | ObjectOrChildModel<Tile>[], appendPlaceholders?: boolean) {
+  insertTiles(tilesToInsert: ObjectOrChildModel<TTile> | ObjectOrChildModel<TTile>[]) {
+    this.insertTilesInternal(tilesToInsert);
+  }
+
+  protected insertTilesInternal(tilesToInsert: ObjectOrChildModel<TTile | PlaceholderTile> | ObjectOrChildModel<TTile | PlaceholderTile>[], appendPlaceholders?: boolean) {
```
mvilliger: "_" in some protected method names missing

eclipse-scout-core/src/tile/TileGrid.ts
```diff
@@ -1165,18 +1207,18 @@ export class TileGrid extends Widget implements TileGridModel {
...
+    return arrays.findIndexFrom(this._filteredTiles, startIndex, (tile, i) => {
       return tile.gridData.x === x && tile.gridData.y === y;
     }, reverse);
   }
 
   /**
    * If the max range is used, the live list of filtered tiles is returned, because every tile has to be in the range.
    */
-  findTilesInRange(viewRange: Range, filter?: Predicate<Tile>): Tile[] {
+  findTilesInRange(viewRange: Range, filter?: Predicate<TTile | PlaceholderTile>): (TTile | PlaceholderTile)[] {
```
mvilliger: Should the predicate really get placeholders? For other public methods the placeholders are like not existing. But here developers have to deal with them. Maybe never include placeholders in the find and therefore also not add them to the predicate and return API?
Same maybe also applies to method "eachTileInRow" below.

eclipse-scout-core/src/tile/TileGrid.ts
```diff
@@ -1574,12 +1616,12 @@ export class TileGrid extends Widget implements TileGridModel {
    * If virtual is false, the live list of filtered tiles is returned, because every tile has to be rendered.
    * If virtual is true, the rendered tiles are collected and returned.
    */
-  renderedTiles(): Tile[] {
+  renderedTiles(): (TTile | PlaceholderTile)[] {
```
mvilliger: Do you really want to return placeholders here? Are they of any interest to developers? Maybe they should be completely hidden for them and therefore also not be part of the public API?

mvilliger (commented):
I am missing some Specs for the new behavior.
E.g. to test that no placeholders are returned, that no placeholders can be selected, etc?
And maybe also modify the Commit message title, to indicate that not only a generic was added?

PR #677, opened 2023-07-12
System properties must be set as VM parameters

matthiaso:
And not as program parameters, compare with [webapp] dev ui.run_.xml file in the same directory (which does set system properties correctly as VM parameters).


matthiaso:
Go jenkins

matthiaso:
Jenkins go

PR #682, opened 2023-07-14
Add support for IJsPage

matthiaso:
355584


org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/desktop/JsonOutline.java
```diff
@@ -194,6 +199,9 @@ protected void putNodeType(JSONObject json, ITreeNode node) {
     else if (node instanceof IPageWithTable) {
       nodeType = "table";
     }
+    if (node instanceof IJsPage) {
```
matthiaso, line 202: else?

stephan-merkli:
Integrated in https://github.com/eclipse-scout/scout.rt/pull/702, thus this one is closed.

PR #702, opened 2023-07-21
Add JsPage

fschinkel:
The JsPage is a Java-page that wraps a page implemented in JS. This
allows a Scout Classic application to open a ScoutJS page.
Add destroy events to page.
Add remote-property to Scout Classic pages.

326014

### DIFF
--- a/eclipse-scout-core/src/desktop/outline/Outline.ts
+++ b/eclipse-scout-core/src/desktop/outline/Outline.ts
@@ -156,6 +156,7 @@ export class Outline extends Tree implements DisplayParent, OutlineModel {
     this._setSelectedViewTabs(this.selectedViewTabs);
     this._setMenus(this.menus);
     this.updateDetailContent();
+    this._nodesSelectedInternal(this.selectedNodes);
   }
 
   /**
@@ -1203,6 +1204,10 @@ export class Outline extends Tree implements DisplayParent, OutlineModel {
   }
 
   pageChanged(page: Page) {
+    if (page?.pageChanging) {
+      return;
+    }
+
     if (page) {
       this._initDetailTableAndForm(page);
     }
--- a/eclipse-scout-core/src/desktop/outline/OutlineAdapter.ts
+++ b/eclipse-scout-core/src/desktop/outline/OutlineAdapter.ts
@@ -7,7 +7,9 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-import {App, EventHandler, Form, objects, Outline, Page, RemoteEvent, scout, Table, TableAdapter, TableFilterRemovedEvent, TableRow, TableRowInitEvent, TableRowsInsertedEvent, TreeAdapter} from '../../index';
+import {
+  App, ChildModelOf, EventHandler, Form, objects, Outline, Page, RemoteEvent, scout, Table, TableAdapter, TableFilterRemovedEvent, TableRow, TableRowInitEvent, TableRowsInsertedEvent, Tree, TreeAdapter, TreeNode, TreeNodeModel
+} from '../../index';
 
 export class OutlineAdapter extends TreeAdapter {
   declare widget: Outline;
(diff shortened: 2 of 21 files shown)


eclipse-scout-core/src/desktop/outline/pages/Page.ts
```diff
@@ -126,13 +126,15 @@ export class Page extends TreeNode implements PageModel {
   }
 
   protected override _destroy() {
+    this.trigger('destroying');
```
cguglielmo, line 136: where are these events used?

eclipse-scout-core/src/table/columns/ColumnModel.ts
```diff
@@ -46,6 +46,14 @@ export interface ColumnModel<TValue = string> extends ObjectModel<Column<TValue>
    */
   index?: number;
 
+  /**
```
cguglielmo, line 49: Maybe add "This is mainly used to restore the selection when rows are replaced using {@link table.replaceRows}"

eclipse-scout-core/src/table/TableAdapter.ts
```diff
@@ -651,6 +660,7 @@ export class TableAdapter extends ModelAdapter {
     }
 
     objects.replacePrototypeFunction(Table, '_createRow', TableAdapter._createRowRemote, true);
+    objects.replacePrototypeFunction(Table, 'replaceRows', TableAdapter.replaceRows, true);
```
cguglielmo: Why is this replaced for classic? This is only used by PageWithTable which does not exist in 
classic mode

eclipse-scout-core/src/table/Table.ts
```diff
@@ -2885,6 +2885,13 @@ export class Table extends Widget implements TableModel {
     this._sort();
   }
 
+  replaceRows(rows: ObjectOrModel<TableRow> | ObjectOrModel<TableRow>[]) {
+    const selectedKeys = this.getSelectedKeys();
+    this.deleteAllRows();
+    this.insertRows(rows);
+    this.restoreSelection(selectedKeys);
```
cguglielmo, line 2895: Please add some js doc explaining how the selection is restored.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/outline/pages/js/AbstractJsPage.java
```diff
@@ -0,0 +1,82 @@
...
+    return m_jsPageModel;
+  }
+
+  @Override
+  public void setJsPageModel(IDoEntity jsPageModel) {
+    m_jsPageModel = jsPageModel;
+  }
+
+  @Override
+  protected ITable createTable() {
```
cguglielmo: maybe make it final?

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/desktop/JsonOutline.java
```diff
@@ -176,12 +184,16 @@ protected JSONObject treeNodeToJson(ITreeNode node, IChildNodeIndexLookup childI
...
     IPage<?> page = (IPage<?>) node;
     JSONObject json = super.treeNodeToJson(node, childIndexes, acceptedNodes);
+    json.put("remote", true);
     putDetailFormAndTable(json, page);
     putNodeType(json, node);
     putProperty(json, PROP_NAVIGATE_BUTTONS_VISIBLE, page.isNavigateButtonsVisible());
     json.put(PROP_OVERVIEW_ICON_ID, page.getOverviewIconId());
     json.put(PROP_SHOW_TILE_OVERVIEW, page.isShowTileOverview());
     json.put(PROP_COMPACT_ROOT, page.isCompactRoot());
+    if (node instanceof IJsPage) {
```
cguglielmo, line 188: Is it necessary to send all the properties above for jsPages? Do they have any effect?

eclipse-scout-core/src/desktop/outline/Outline.ts
```diff
@@ -174,7 +175,9 @@ export class Outline extends Tree implements DisplayParent, OutlineModel {
     nodeModel = nodeModel || {};
     nodeModel.objectType = scout.nvl(nodeModel.objectType, Page);
     nodeModel.parent = this;
-    return scout.create(nodeModel as FullModelOf<Page>);
+    let page = scout.create(nodeModel as FullModelOf<Page>);
+    page.on('propertyChange:detailForm propertyChange:detailTable', e => this.pageChanged(page));
```
cguglielmo: pageChanged is now called multiple times in classic mode which is unnecessary

eclipse-scout-core/src/desktop/outline/Outline.ts
```diff
@@ -156,6 +156,7 @@ export class Outline extends Tree implements DisplayParent, OutlineModel {
     this._setSelectedViewTabs(this.selectedViewTabs);
     this._setMenus(this.menus);
     this.updateDetailContent();
+    this._nodesSelectedInternal([]);
```
cguglielmo: empty array is actually not correct, either pass this.selectedNodes or maybe better remove the parameter completely because it is not used

cguglielmo (changes requested):
- Please describe in commit message, why it is necessary to send the current data in case of a form reset.

eclipse-scout-core/src/desktop/outline/pages/Page.ts
fschinkel: If a Scout JS page uses the `HybridManager` to set a Scout Classic form as its detail form or search form, this form needs to be closed when the page is destroyed. Therefore, the page listens on its own destroying-event to trigger a `form.close()`.

eclipse-scout-core/src/desktop/outline/OutlineAdapter.ts
```diff
@@ -28,26 +30,30 @@ export class OutlineAdapter extends TreeAdapter {
...
+          this._destroyDetailTable(page);
+        }
+        page.setDetailTable(detailTable);
+        if (page.detailTable) {
+          this._initDetailTable(page);
+        }
       }
+    } finally {
+      page.setPageChanging(false);
+      this.widget.pageChanged(page);
```
cguglielmo, line 59: If there was an error, pageChanged event should probably not be triggered?

eclipse-scout-core/src/table/Table.ts
```diff
@@ -3526,6 +3536,66 @@ export class Table extends Widget implements TableModel {
...
+        if (!keys.length) {
+          break;
+        }
+      }
+    }
+    return rows;
+  }
+
+  /**
+   * (see {@link Table.getRowsByKey})
```
cguglielmo: Use @see to just reference another method

eclipse-scout-core/src/table/Table.ts
```diff
@@ -3526,6 +3536,66 @@ export class Table extends Widget implements TableModel {
     return this.selectedRows.indexOf(row) > -1;
   }
 
+  /**
+   * Restores the selection by the given selectedKeys. Rows matching these given values (see {@link Table.getRowsByKey}) will be selected.
+   *
+   * @param selectedKeys array of key values (see {@link TableRow.getKeyValues}) of the rows to select.
+   */
+  restoreSelection(selectedKeys: any[][]) {
```
cguglielmo, line 3544: Is it expected that the selection will be removed if selectedKeys is empty or null? It maybe correct, I don't know

cguglielmo (commented):
Looks good!

eclipse-scout-core/src/table/Table.ts
fschinkel: That behavior is intended.

fschinkel:
run tests

cguglielmo (approved):

PR #707, opened 2023-07-26
Rework ActionUtility and filter menus recursively

fschinkel:
Most methods of ActionUtility are built for menus but accept IAction.
Therefore, they do not handle child menus as IAction does not know child
actions. Move those methods to MenuUtility, change their signature to
IMenu and migrate all usages. As a consequence also the signature of
MenuWrapper, OutlineMenuWrapper and OutlineFormMenuWrapper are changed
accordingly.
The behaviour of MenuUtility.filterMenusRec (moved from
ActionUtility.getActions) was changed slightly. A menu accepted by the
filter that has child menus will now be wrapped and its child menus are
filtered by the given filter as well.
Consider a menu A with child menus B and C and the filter accepts A and
B. The result of MenuUtility.filterMenusRec will be a wrapped menu A
with only one child B.

240403

### DIFF
--- a/org.eclipse.scout.rt.client.test/src/test/java/org/eclipse/scout/rt/client/ui/action/ActionUtilityTest.java
+++ b/org.eclipse.scout.rt.client.test/src/test/java/org/eclipse/scout/rt/client/ui/action/ActionUtilityTest.java
@@ -9,8 +9,7 @@
  */
 package org.eclipse.scout.rt.client.ui.action;
 
-import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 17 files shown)


cguglielmo (commented):
The title suggests that only refactoring has taken place. But there are actual changes of the functionality. Please describe, what and why was changed.
Also rename the MenuUtility.getMenus to make it more clear, that wrappers are created. E.g. filterMenusAndWrapChildren
There is also a type in the commit message: are build -> are built
Please push to 24.1 only

PR #712, opened 2023-08-04
Multidimensional properties: support truthy and falsy

cguglielmo:
In earlier versions, widget.setVisible(null) or widget.setEnabled('hi') worked as expected.
Now, if the parameter is an object, it is treated as dimension object, which means the key is used as dimension and the value as dimension value.
For existing code, this is wrong and creates errors that are hard to find.
To mitigate it, the parameter is now checked for the correct structure. If the structure is wrong, it will be converted to a boolean and used for the default dimension.

This should work except for 2 cases because the structure is correct:
- If an empty object was used ({})
- If an object was used whose values had the correct data type ({dim: false})

### DIFF
--- a/eclipse-scout-core/test/events/PropertyEventEmitterSpec.ts
+++ b/eclipse-scout-core/test/events/PropertyEventEmitterSpec.ts
@@ -371,6 +371,73 @@ describe('PropertyEventEmitter', () => {
         expect(emitter.getPropertyDimension('multiProp', 'dim1')).toBe(false);
       });
 
+      it('only accepts objects with the correct structure for legacy reasons', () => {
+        // Prior to the multidimensional properties, not only real booleans were accepted
+        // but also falsy (null, undefined, empty string, 0) and truthy values (strings, objects, numbers etc.)
+        // Even though it is bad practise, people used it and without using TypeScript, such errors are hard to find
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/test/events/PropertyEventEmitterSpec.ts
```diff
@@ -371,6 +371,64 @@ describe('PropertyEventEmitter', () => {
...
+        expect(emitter.getPropertyDimension('multiProp', 'default')).toBe(false);
+        expect(emitter.getPropertyDimension('multiProp', 'dim1')).toBe(true);
+        expect(emitter.getProperty('multiProp')).toBe(false);
+
+        emitter.setProperty('multiProp', {});
+        expect(emitter.getProperty('multiProp')).toBe(true);
+
+        // Multidimensional properties only work with booleans
+        // If not all values in the given object are booleans, the object will be used as value for the default dimension
+        emitter.setProperty('multiProp', {dim1: 'asdf', default: false});
```
fschinkel, line 426: Maybe add `emitter.setProperty('multiProp', {dim1: false, default: true});`-block before to show that even if `dim1` is truthy it will not be changed by this call.

fschinkel (approved):

PR #717, opened 2023-08-09
Reposition tooltips after layout validation

fschinkel:
Add a permanent (adds itself again an execution) post validation function if there are tooltips present on the desktop. This solves several tooltip position problems:
* collapse a group box containing a tooltip
* collapse a group box so that a sibling group box containing a tooltip moves to a new position
* collapse an accordion group containing a tooltip
* collapse an accordion group so that a sibling accordion group containing a tooltip moves to a new position
* move splitter in order to resize content containing a tooltip And it makes it unnecessary to explicitly call tooltip.position() during layout.
Revalidate layout tree after the resize of a form or widget popup to ensure that all tooltips are correctly positioned. After a popups open-animation ends and after a column resize reposition all tooltips.

308905, 317648, 318045

### DIFF
--- a/eclipse-scout-core/src/popup/Popup.ts
+++ b/eclipse-scout-core/src/popup/Popup.ts
@@ -256,6 +256,7 @@ export class Popup extends Widget implements PopupModel {
         return;
       }
       this.$container.addClassForAnimation('animate-open');
+      this.$container.oneAnimationEnd(() => this.findDesktop().repositionTooltips());
     });
   }
 
--- a/eclipse-scout-core/src/table/Table.ts
+++ b/eclipse-scout-core/src/table/Table.ts
@@ -4073,6 +4073,8 @@ export class Table extends Widget implements TableModel {
         this._resizeAggregateCell(this.$cell(column, aggregateRow.$row));
       }
     });
+
+    this.findDesktop().repositionTooltips();
   }
 
   /** @internal */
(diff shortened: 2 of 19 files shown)


eclipse-scout-core/src/popup/Popup.ts
```diff
@@ -256,6 +256,7 @@ export class Popup extends Widget implements PopupModel {
         return;
       }
       this.$container.addClassForAnimation('animate-open');
+      this.$container.oneAnimationEnd(() => this.findDesktop()?.repositionTooltips());
```
cguglielmo: There is always a desktop, the ? should not be necessary.

eclipse-scout-core/src/table/Table.ts
```diff
@@ -4073,6 +4073,8 @@ export class Table extends Widget implements TableModel {
         this._resizeAggregateCell(this.$cell(column, aggregateRow.$row));
       }
     });
+
+    this.findDesktop()?.repositionTooltips();
```
cguglielmo: ? should not be necessary

eclipse-scout-core/src/tooltip/Tooltip.ts
```diff
@@ -117,14 +117,19 @@ export class Tooltip extends Widget implements TooltipModel {
       scrollbars.onScroll(this.$anchor, this._anchorScrollHandler);
     }
 
-    // If the tooltip is rendered inside a (popup) dialog, get a reference to the dialog.
-    this.dialog = this.findParent(p => p instanceof Form && p.isDialog()) as Form;
+    // If the tooltip is rendered inside a popup (dialog form or popup), get a reference to the popup.
+    this._popup = this.findParent(p => p instanceof WidgetPopup);
```
cguglielmo: You can use the shorthand findParent(WidgetPopup)

PR #726, opened 2023-08-14
TreeBox: ignore invalid lookup rows

bschwarzent:
- Prevent endless loop when lookup rows reference themselves
- Ignore lookup rows with key "null" (otherwise, the meaning of parentKey=null would no longer be defined)
- Add Jasmine tests

358497


bschwarzent:
Unknown state of PR, seems to be obsolete -> close

PR #734, opened 2023-08-28
Create ValueDo in TypeScript as equivalent to IValueDo in Java

Knicksaas:
In Java, we have a separate implementation for each data type. This is because the Jackson object mapper cannot map a generic object. In TypeScript, we do not have this issue, so there is just this generic interface.

348303

### DIFF
--- a/eclipse-scout-core/src/dataobject/ValueDo.ts
+++ b/eclipse-scout-core/src/dataobject/ValueDo.ts
@@ -1,0 +1,24 @@
+/*
+ * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {DoEntity} from './DoEntity';
+
+/**
+ * <strong>Wrapper data object for a generic value.</strong>
+ *
+ * In Java, we have a separate implementation for each data type. This
+ * is because the Jackson object mapper cannot map a generic object.
+ * In TypeScript, we do not have this issue, so there is just this
+ * generic interface.
+ *
+ * @see "org.eclipse.scout.rt.dataobject.value.IValueDo"
+ */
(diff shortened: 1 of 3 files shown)


eclipse-scout-core/src/util/BinaryResource.ts
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+/**
+ * Warper for binary content with metadata.
```
cguglielmo: Typo: Wrapper instead of Warper
In General: Use Markdown syntax in JsDoc instead of HTML -> <p> can be removed.

eclipse-scout-core/src/dataobject/ValueDo.ts
```diff
@@ -0,0 +1,24 @@
...
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {DoEntity} from './DoEntity';
+
+/**
+ * <strong>Wrapper data object for a generic value.</strong>
```
cguglielmo: Please don't create titles using strong, Seperating the first line with an empty line is sufficient.

eclipse-scout-core/src/dataobject/ValueDo.ts
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {DoEntity} from './DoEntity';
```
cguglielmo: Always import from index.

PR #742, opened 2023-09-18
Page: add summary support

fschinkel:
Use all columns marked as summary column while computing the text for a
row. The texts of all cells of summary columns are joined with a
whitespace separating them. If no column is marked as summary column use
the text of the first cell.
Add primaryKeyColumns() and summaryColumns() to Table.
Do not send the summary-flag from the server to the browser as it is not
used in the Scout Classic case.

346791

### DIFF
--- a/eclipse-scout-core/src/table/Table.ts
+++ b/eclipse-scout-core/src/table/Table.ts
@@ -5671,6 +5671,20 @@ export class Table extends Widget implements TableModel {
   }
 
   /**
+   * @returns all columns marked as primaryKey column (see {@link Column.primaryKey})
+   */
+  primaryKeyColumns(): Column[] {
+    return this.filterColumns(column => column.primaryKey);
+  }
+
+  /**
+   * @returns all columns marked as summary column (see {@link Column.summary})
+   */
+  summaryColumns(): Column[] {
+    return this.filterColumns(column => column.summary);
+  }
+
+  /**
    * @param includeGuiColumns true to also include columns created by the UI (e.g. {@link rowIconColumn} or {@link checkableColumn}). Default is true.
    */
   filterColumns(filter: Predicate<Column<any>>, includeGuiColumns?: boolean): Column<any>[] {
--- a/eclipse-scout-core/test/desktop/outline/pages/PageSpec.ts
+++ b/eclipse-scout-core/test/desktop/outline/pages/PageSpec.ts
@@ -161,6 +161,35 @@ describe('Page', () => {
     }]);
   });
 
+  describe('computeTextForRow', () => {
+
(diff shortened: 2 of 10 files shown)


eclipse-scout-core/src/table/Table.ts
```diff
@@ -3658,6 +3658,24 @@ export class Table extends Widget implements TableModel {
     return $row.children('.table-cell').eq(columnIndex);
   }
 
+  /**
+   * @returns all columns marked as primaryKey column (see {@link Column.primaryKey})
+   */
+  getPrimaryKeyColumns(): Column[] {
```
cguglielmo: move to visibleColumns and name it primaryKeyColumns().

eclipse-scout-core/src/table/Table.ts
```diff
@@ -3658,6 +3658,24 @@ export class Table extends Widget implements TableModel {
...
+    return this._getColumnsByPredicate(column => column.primaryKey);
+  }
+
+  /**
+   * @returns all columns marked as summary column (see {@link Column.summary})
+   */
+  getSummaryColumns(): Column[] {
```
cguglielmo: move to visibleColumns and name it summaryColumns

eclipse-scout-core/src/table/Table.ts
```diff
@@ -3658,6 +3658,24 @@ export class Table extends Widget implements TableModel {
...
+    return this._getColumnsByPredicate(column => column.summary);
```
cguglielmo: use this.filterColumns instead

eclipse-scout-core/test/desktop/outline/pages/PageSpec.ts
```diff
@@ -161,6 +161,35 @@ describe('Page', () => {
...
+      const table = tableHelper.createTable(tableHelper.createModel(
+        tableHelper.createModelColumns(5),
+        $.extend(true, [], tableHelper.createModelRows(5, 2), [{cells: ['a', 'b', 'c', 'd', 'e']}, {cells: ['1', '2', '3', '4', '5']}])
+      ));
+      const [rowAbc, row123] = table.rows;
+
+      expect(page.computeTextForRow(rowAbc)).toBe('a');
+      expect(page.computeTextForRow(row123)).toBe('1');
+
+      table.columns[1].summary = true;
```
cguglielmo: Can you add setSummary on Column.ts similar to setCssClass? updateRows should trigger the page update, so changing it dynamically should work

cguglielmo (changes requested):

cguglielmo (approved):

PR #743, opened 2023-09-18
FormController: do not reactivate the removed dialog

fschinkel:
Consider the recently removed dialog while looking for a form to activate after the dialog removal to prevent it from being reactivated. If there is a dialog and its displayParent is a detailForm of a page and another page is being selected, the detailForm is removed which triggers the removal of the dialog. As the dialog is contained in the detailForms dialogs it needs to be excluded from this list while looking for a new form to activate.

356016

### DIFF
--- a/eclipse-scout-core/src/form/FormController.js
+++ b/eclipse-scout-core/src/form/FormController.js
@@ -1,9 +1,9 @@
 /*
- * Copyright (c) 2014-2018 BSI Business Systems Integration AG.
+ * Copyright (c) 2010-2023 BSI Business Systems Integration AG.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/src/form/FormController.js
```diff
@@ -221,9 +221,10 @@ export default class FormController {
     }
   }
 
-  _findFormToActivateAfterDialogRemove() {
-    if (this.displayParent.dialogs.length > 0) {
-      return this.displayParent.dialogs[this.displayParent.dialogs.length - 1];
+  _findFormToActivateAfterDialogRemove(removedDialog) {
+    const dialogs = this.displayParent.dialogs.filter(d => d !== removedDialog);
+    if (dialogs > 0) {
```
cguglielmo: this should probably be dialogs.length > 0. It is interesting, that no spec fails. At least "will be set to the display parent form if dialog closes" should fail, but maybe there is a different logic that handles that case.

cguglielmo (changes requested):

cguglielmo (approved):

PR #746, opened 2023-09-22
Scout JS Form: use any instead of object in _load

stephan-merkli:
Because this.data is typed any too.


fschinkel (changes requested):
Please also change this for `Form._save` and `FormModel.data`.

fschinkel (approved):

PR #748, opened 2023-09-28
A11y: Add ARIA attributes to Scout

jwinkens:
Setting ARIA attributes enables screen readers to work with the framework.


eclipse-scout-chart/src/chart/ChartJsRenderer.ts
```diff
@@ -260,6 +262,55 @@ export class ChartJsRenderer extends AbstractChartRenderer {
     this.chartJs.update();
   }
 
+  /**
+   * Renders data as an invisible table instead of the chart so screen reader users can navigate that data
+   */
+  protected _renderTableForScreenReader(config: ChartConfig) {
```
jwinkens: We may want to remove this and add ChartJS descriptions to the set of open tasks

eclipse-scout-core/src/action/Action.ts
```diff
@@ -369,6 +384,9 @@ export class Action extends Widget implements ActionModel {
 
   protected _renderSelected() {
     this.$container.toggleClass('selected', this.selected);
+    aria.togglePressed(this.$container, this.isToggleAction() && this.actionStyle === Action.ActionStyle.BUTTON, this.selected);
+    aria.toggleExpanded(this.$container, this.isToggleAction() && this.actionStyle === Action.ActionStyle.DEFAULT, this.selected);
```
jwinkens: How can we determine if an action opens a pop up or if it is just a toggle button?

cguglielmo:
Jenkins run tests

org.eclipse.scout.rt.chart.ui.html/src/main/resources/org/eclipse/scout/rt/chart/ui/html/texts/Texts.properties
```diff
@@ -10,8 +10,10 @@
 ui.Bio=Tri.
 ui.Brd=quad
 ui.Chart=Chart
+ui.FulfillmentChartAriaDescription=Fulfillment chart current value {0} %
 ui.Mio=Mio.
 ui.Mrd=Bio.
+ui.SpeedoChartAriaDescription=placeholder {0} {1} {2}
```
cguglielmo: Replace placeholder with actual text

org.eclipse.scout.rt.chart.ui.html/src/main/resources/org/eclipse/scout/rt/chart/ui/html/texts/Texts_de.properties
```diff
@@ -10,8 +10,10 @@
 ui.Bio=Bio.
 ui.Brd=Brd.
 ui.Chart=Diagramm
+ui.FulfillmentChartAriaDescription=Erfüllungsgrad Diagramm Aktueller Wert {0} %
```
cguglielmo: Spielen die Satzzeichen eine Rolle für den Screenreader? Vielleicht besser ändern auf: Erfüllungsgrad-Diagramm. Aktueller Wert: {0}%.

org.eclipse.scout.rt.chart.ui.html/src/main/resources/org/eclipse/scout/rt/chart/ui/html/texts/Texts_de.properties
```diff
@@ -10,8 +10,10 @@
...
 ui.Mio=Mio.
 ui.Mrd=Mrd.
+ui.SpeedoChartAriaDescription=Zeigerdiagramm Aktueller Wert {0} Minimaler Wert {1} Maximaler Wert {2}
```
cguglielmo: Auch hier Satzzeichen ergänzen, siehe anderen Kommentar. Das Gleiche beim englischen Text.

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+  treegrid($elem: JQuery<Element>) {
+    aria.role($elem, 'treegrid');
+  },
+
+  /******************************************************************************************************************
+   * Attributes
+   ******************************************************************************************************************/
+
+  /**
+   * Determines weather a labelledby id is inserted at the front or the back of current aria-labelledby value.
```
cguglielmo: weather -> whether

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+   */
+  LABELLED_BY_INSERT_POSITION: {
```
cguglielmo: Create a type for this object as well, so it can be used in the method signature. Use can either use an enum: export enum AriaLabelledByInsertPosition, or EnumObject which creates a type for this object. A think I would use a real enum here. Also use CamelCase for the enum name and UPPERCASE only for the values of the enum

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+    FRONT: 'front',
+    BACK: 'back'
+  },
+
+  /**
+   * Links the element to the given target element by giving an id to the target element (if needed) and prepending this id to the
+   * given aria attribute. If replace is set to true, completely replaces the attribute.
+   *
+   * For insert position see {@link LABELLED_BY_INSERT_POSITION}, insert position has no effect if the labelledby property is replaced.
```
cguglielmo: Use @param position

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+      }
+    }
+    $elem.attr(ariaAttribute, targetId);
+  },
+
+  /**
+   * Links the given element with the given label by setting aria-labelledby.
+   * This allows screen readers to build a catalog of the elements on the screen and their relationships, for example, to read the label when the input is focused.
+   *
+   * Per default linked labels are added to existing linked labels separated by space, if you want to completely replace the linked label, set replace to true
```
cguglielmo: Make a point after "by space" and after "set replace to true".

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+   * @see <a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby">ARIA: aria-labelledby</a>
+   */
+  linkElementWithLabel($elem: JQuery<Element>, $label: JQuery<Element>, position = aria.LABELLED_BY_INSERT_POSITION.FRONT, replace = false) {
+    aria._linkElementWithTargetElement($elem, $label, 'aria-labelledby', 'lbl', position, replace);
+  },
+
+  /**
+   * Links the given element with the given description by setting aria-describedBy.
+   *
+   * Per default linked descriptions are added to existing linked descriptions separated by space, if you want to completely replace the linked description, set replace to true
```
cguglielmo: See comment above regarding points.

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+    aria.linkElementWithDescription($elem, $descriptionElement, position, replace);
+    return $descriptionElement;
+  },
+
+  /**
+   * Links chart with its field label so the field name is read when entering the chart
+   *
+   * @see <a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby">ARIA: aria-labelledby</a>
+   */
+  linkChartWithFieldLabel($chart: JQuery<Element>) {
```
cguglielmo: This belongs into the chart-module. eclipse-scout-core doesn't know about charts.

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+    if (!$chart) {
+      return;
+    }
+    let $field = $chart.parents('.chart-field');
```
cguglielmo: Charts can be used without a chart-field. I think it should be the other way round: The chart-field should link the label with the chart. Or is it only necessary for svg based charts?

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+    }
+
+    if (currentLevel) {
+      aria.heading($header, currentLevel);
+      aria._addHeaderLevelToElement($elem, currentLevel);
+    }
+  },
+
+  /**
+   * <pre>
```
cguglielmo: Use `<pre>` only for the snippets, not the whole js doc

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+   * In most cases you should just use {@link linkElementWithHeader} which automatically creates an aria heading for you and assigns levels correctly.
+   *
+   * Use this to implicitly link your heading with its container by adding the header level to the container.
+   * This will be used to try and ensure the heading structure is consistent. Normally your DOM structure looks something like this:
```
cguglielmo: Maybe rewrite to: This should ensure the heading structure is consistent..

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+  },
+
+  /**
+   * In most cases you should just use {@link linkElementWithHeader} which automatically creates an aria heading for you and assigns levels correctly.
+   * Derives the current header level by going upwards in the DOM structure and finding the last header level used.
+   * If no parent with a heading is found, returns undefined.
+   */
+  _computeHeaderLevel($elem: JQuery<Element>): number {
+    if (!$elem) {
+      return undefined;
```
cguglielmo: return undefined; is the same as return;, so no need to write undefined. Better return null instead.

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+    }
+    let $parentHeader = $elem.parents('div[data-aria-header-level]');
```
cguglielmo: Does it have to be a div? Don't make it more specific than necessary.

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+   * @see <a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-required">ARIA: aria-required</a>
+   */
+  required($elem: JQuery<Element>, value: boolean) {
+    if (!$elem) {
+      return;
+    }
+    $elem.attr('aria-required', String(value === true));
+  },
+
+  toggleRequired($elem: JQuery<Element>, condition: boolean, value: boolean) {
```
cguglielmo: The usage of this method feels a bit cumbersome. I propose to replace these 3 methods with just one method: required($elem, value), which either sets the value to true or removes it

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+    if (!$elem) {
+      return;
+    }
+    $elem.removeAttr('aria-required');
+  },
+
+  /**
+   * @see <a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label">ARIA: aria-label</a>
+   */
+  labelElement($elem: JQuery<Element>, label: string, allowEmpty = false) {
```
cguglielmo: Why don't you just call it label instead of labelElement?

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+    if (!$elem) {
+      return;
+    }
+    $elem.removeAttr('aria-description');
+  },
+
+  /**
+   * @see <a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked">ARIA: aria-checked</a>
+   */
+  checked($elem: JQuery<Element>, value: boolean, triStateAllowed = false) {
```
cguglielmo, line 404: See comment for required

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+   * @see <a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded">ARIA: aria-expanded</a>
+   */
+  expanded($elem: JQuery<Element>, value: boolean) {
+    if (!$elem) {
+      return;
+    }
+    $elem.attr('aria-expanded', String(value === true));
+  },
+
+  toggleExpanded($elem: JQuery<Element>, condition: boolean, value: boolean) {
```
cguglielmo: See comment for required

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+   * @see <a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected">ARIA: aria-selected</a>
+   */
+  selected($elem: JQuery<Element>, value: boolean) {
+    if (!$elem) {
+      return;
+    }
+    $elem.attr('aria-selected', String(value === true));
+  },
+
+  toggleSelected($elem: JQuery<Element>, condition: boolean, value: boolean) {
```
cguglielmo: See comment for required

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+   * @see <a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed">ARIA: aria-pressed</a>
+   */
+  pressed($elem: JQuery<Element>, value: boolean) {
+    if (!$elem) {
+      return;
+    }
+    $elem.attr('aria-pressed', String(value === true));
+  },
+
+  togglePressed($elem: JQuery<Element>, condition: boolean, value: boolean) {
```
cguglielmo: See comment for required

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+   * @see <a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden">ARIA: aria-hidden</a>
+   */
+  hidden($elem: JQuery<Element>, value: boolean) {
+    if (!$elem) {
+      return;
+    }
+    $elem.attr('aria-hidden', String(value === true));
+  },
+
+  toggleHidden($elem: JQuery<Element>, condition: boolean, value: boolean) {
```
cguglielmo: See comment for required

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+   * @see <a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-multiselectable">ARIA: aria-multiselectable</a>
+   */
+  multiselectable($elem: JQuery<Element>, value: boolean) {
+    if (!$elem) {
+      return;
+    }
+    $elem.attr('aria-multiselectable', String(value === true));
+  },
+
+  toggleMultiselectable($elem: JQuery<Element>, condition: boolean, value: boolean) {
```
cguglielmo: See comment for required

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+   * @see <a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled">ARIA: aria-disabled</a>
+   */
+  disabled($elem: JQuery<Element>, value: boolean) {
+    if (!$elem) {
+      return;
+    }
+    $elem.attr('aria-disabled', String(value === true));
+  },
+
+  toggleDisabled($elem: JQuery<Element>, condition: boolean, value: boolean) {
```
cguglielmo: See comment for required

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
+/*
+ * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {widgets} from './widget/widgets';
```
cguglielmo: Import everything from index

eclipse-scout-core/src/aria.ts
```diff
@@ -0,0 +1,945 @@
...
+    if (!$elem) {
+      return;
+    }
+    $elem.removeAttr('role');
+  },
+
+  /**
+   * @see <a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role">ARIA: alert role</a>
+   */
+  alert($elem: JQuery<Element>) {
```
cguglielmo: We should discuss whether it would be better to create an enum for role and always use aria.role instead of having a method for every role. I think this would make the api of aria object more clear and simpler. 
Instead of an enum, I would even suggest a string literal type:  
`type AriaRole = 'button' | 'alert' | ...`
This makes it simple to use and you still get type errors if a wrong string is used.

eclipse-scout-core/src/menu/Menu.ts
```diff
@@ -322,6 +331,8 @@ export class Menu extends Action implements MenuModel {
         this.$submenuIcon = this.$container
           .appendSpan('submenu-icon')
           .text(icon.iconCharacter);
+        aria.img(this.$submenuIcon);
+        aria.labelElement(this.$submenuIcon, '');
```
cguglielmo: submenuIcon is a new element, setting label to '' will probably do nothing. Should replace be set to true?

eclipse-scout-core/src/menu/Menu.ts
```diff
@@ -519,11 +531,16 @@ export class Menu extends Action implements MenuModel {
...
+    aria.linkElementWithControls(this.$container, this.popup.$container);
     this.popup.one('destroy', event => {
       this.popup = null;
+      aria.removeControls(this.$container);
     });
     // Unselect on close which comes earlier than destroy (before the animation), to give more immediate feedback
-    this.popup.on('close', event => this.setSelected(false));
+    this.popup.on('close', event => {
+      this.setSelected(false);
+      aria.removeControls(this.$container);
```
cguglielmo: Should not be necessary here, destroy will be triggered eventually so it is sufficient to call removeControls in the destroy handler

eclipse-scout-core/src/modeselector/ModeSelector.ts
```diff
@@ -57,6 +73,38 @@ export class ModeSelector<TModeRef = any> extends Widget implements ModeSelector
...
 
+  protected override _renderEnabled() {
+    super._renderEnabled();
+    this._provideTabIndex();
+  }
+
+  protected _provideTabIndex() {
+    let tabSet;
+    this.modes.forEach(function(mode) {
+      if (mode.enabledComputed && this.enabledComputed && !tabSet) {
```
cguglielmo: mode.enabledComputed considers the enabled state of the parents, so checking for this.enabledComputed should not be necessary

eclipse-scout-core/src/modeselector/ModeSelector.ts
```diff
@@ -57,6 +73,38 @@ export class ModeSelector<TModeRef = any> extends Widget implements ModeSelector
...
   }
 
+  protected override _renderEnabled() {
+    super._renderEnabled();
+    this._provideTabIndex();
+  }
+
+  protected _provideTabIndex() {
+    let tabSet;
+    this.modes.forEach(function(mode) {
```
cguglielmo: prefer using arrow functions

eclipse-scout-core/src/modeselector/ModeSelector.ts
```diff
@@ -57,6 +73,38 @@ export class ModeSelector<TModeRef = any> extends Widget implements ModeSelector
     this._renderModes();
   }
 
+  protected override _renderEnabled() {
+    super._renderEnabled();
+    this._provideTabIndex();
+  }
+
+  protected _provideTabIndex() {
```
cguglielmo, line 81: If a selected mode is disabled dynamically, it loses the tabindex -> mode selector is not tabbable anymore -> We should do the same as for radio buttons: keep tabindex even if selected mode is disabled. Can be reproduced with jswidgets app

eclipse-scout-core/src/modeselector/Mode.less
```diff
@@ -68,6 +68,10 @@
       margin-right: 8px;
     }
   }
+
+  &:focus {
+    box-shadow: none;
```
cguglielmo, line 73: To make the focus visible, you can use the following snippet:
```
    &.keyboard-navigation {
      #scout.focus-inset-box-shadow();
    }
```
You also need to set the class keyboard-navigation in your new keystrokes. With this pattern, the focus border is only visible when using the keyboard and not when clicking with the mouse.

eclipse-scout-core/src/modeselector/Mode.ts
```diff
@@ -47,6 +49,7 @@ export class Mode<TRef = any> extends Action implements ModeModel<TRef> {
 
     if (!this.selected) {
       this.setSelected(true);
+      this.focus();
```
cguglielmo, line 57: Is this necessary? The browser sets the focus automatically when a tabbable element is clicked.

eclipse-scout-core/src/modeselector/ModeSelectorLeftOrUpKeyStroke.ts
```diff
@@ -0,0 +1,39 @@
...
+      if (fieldBefore && mode === focusedButton) {
+        fieldBefore.setSelected(true);
+        fieldBefore.focus();
+        return true;
+      }
+      if (mode.enabledComputed && mode.visible) {
+        fieldBefore = mode;
+      }
+      return false;
+    }, this);
```
cguglielmo: `this` parameter is not necessary when using arrow functions, the scope will be correct already

eclipse-scout-core/src/popup/WidgetPopup.ts
```diff
@@ -120,7 +120,8 @@ export class WidgetPopup<TContent extends Widget = Widget> extends Popup impleme
     return scout.create(Action, {
       parent: this,
       cssClass: 'close-action menu-item',
-      iconId: icons.REMOVE
+      iconId: icons.REMOVE,
+      text: this.session.text('ui.Close')
```
cguglielmo: text is now shown in the ui, see jswidgets-app > Popup

eclipse-scout-core/src/session/BusyIndicator.ts
```diff
@@ -194,6 +196,9 @@ export class BusyIndicator extends Widget implements BusyIndicatorModel {
     this.$details
```
cguglielmo: I think linkElementWithDescription should be done in _render() independently of the value of this.details because the property can change dynamically

eclipse-scout-core/src/session/BusyIndicator.ts
```diff
@@ -183,6 +184,7 @@ export class BusyIndicator extends Widget implements BusyIndicatorModel {
 
   protected _renderLabel() {
     this.$label.text(this.label || '');
+    aria.linkElementWithLabel(this.$container, this.$label);
```
cguglielmo: See comment in renderDetails

eclipse-scout-core/src/table/Table.ts
```diff
@@ -576,7 +579,10 @@ export class Table extends Widget implements TableModel {
   }
 
   protected override _render() {
-    this.$container = this.$parent.appendDiv('table').addDeviceClass();
+    this.$container = this.$parent.appendDiv('table')
+      .addDeviceClass();
+    this.accessibilityRenderer.renderTable(this.$container);
+    aria.multiselectable(this.$container, this.multiSelect || this.multiCheck);
```
cguglielmo: These properties can be changed dynamically -> move to _renderMultiSelect and _renderMultiCheck

eclipse-scout-core/src/table/Table.ts
```diff
@@ -3411,6 +3439,16 @@ export class Table extends Widget implements TableModel {
...
       }
 
+      aria.selected(row.$row, thisRowSelected);
+      row.$row.children('.table-cell').each((index, cell) => {
+        aria.selected($(cell), thisRowSelected);
+      });
+
+      // if single row is selected make it the active descendant for screen readers
+      if (singleRowSelected && thisRowSelected) {
+        aria.linkElementWithActiveDescendant(this.$container, row.$row);
```
cguglielmo: If selection is removed with ctrl-click, activedescendant won't be removed but should. Also if multiple rows are selected, is it correct that activedescendant still points to a single row?

eclipse-scout-core/src/table/TableHeader.ts
```diff
@@ -64,8 +64,10 @@ export class TableHeader extends Widget implements TableHeaderModel {
   }
 
   protected override _render() {
-    this.$container = this.table.$data.beforeDiv('table-header')
-      .cssBorderLeftWidth(this.table.rowBorders.left || '');
+    let $tableHeaderContainer = this.table.$data.beforeDiv('table-header-container');
```
cguglielmo: Adding this new container element breaks existing css rules, for example the one in DesktopBench. Can we avoid that?

eclipse-scout-core/src/table/TableHeader.ts
```diff
@@ -392,6 +397,11 @@ export class TableHeader extends Widget implements TableHeaderModel {
     // Make sure empty header is as height as the others to make it properly clickable
     $headerText.htmlOrNbsp(text, 'empty');
     this._updateColumnIconAndTextStyle(column);
+    // add tooltip as invisible text for screen readers
```
cguglielmo, line 408: Item should be removed if tooltipText changes to empty. Also extract to a new method called _renderTooltipText. And call updateHeader in Column#setHeaderTooltipText to ensure element is updated when tooltip changes

eclipse-scout-core/src/table/TableHeader.ts
```diff
@@ -450,18 +462,26 @@ export class TableHeader extends Widget implements TableHeaderModel {
...
     }
+    let $accessibleState = $state.appendDiv().addClass('text').text(accessibleStateText);
+    aria.screenReaderOnly($accessibleState);
     // Contains sort arrow
-    $state.appendDiv('right');
+    let sortArrow = $state.appendDiv('right');
+    aria.hidden(sortArrow, true);
 
     this._adjustColumnMinWidth(column);
+    this._adjustColumnMinWidth(column);
```
cguglielmo: Duplicate method call looks like a mistake

eclipse-scout-core/src/table/TableHeaderMenuButton.ts
```diff
@@ -36,6 +37,7 @@ export class TableHeaderMenuButton extends Action implements TableHeaderMenuButt
     super._render();
     this.$container = this.$container.addClass('table-header-menu-command button')
       .unfocusable()
+      .on('focusOut', this._onFocusOut.bind(this))
```
cguglielmo: This does not work: focusin is missing, focusOut should be named focusout

eclipse-scout-core/src/table/columns/Column.ts
```diff
@@ -335,7 +335,24 @@ export class Column<TValue = string> extends PropertyEventEmitter implements Col
...
-    return '<div class="' + cssClass + '" style="' + style + '">' + content + '</div>';
+    let ariaAttributes = '';
+    if (this.table.accessibilityRenderer && strings.hasText(this.table.accessibilityRenderer.cellRole)) {
+      ariaAttributes = ' role="' + this.table.accessibilityRenderer.cellRole + '" aria-selected="false"';
+    }
+    // set the label of the cell to header name + cell content
+    if (this.$header) {
+      let $headerText = this.$header.children('.table-header-item-text');
+      if ($headerText.length > 0) {
+        let headerLabelId = $headerText.eq(0).attr('id') as string;
```
cguglielmo: Setting the id should be moved to TableHeader.ts, modifying the header at this point seems wrong.

eclipse-scout-core/src/tile/TileGrid.ts
```diff
@@ -863,6 +864,19 @@ export class TileGrid<TTile extends Tile = Tile> extends Widget implements TileG
     });
   }
 
+  setFocusedTileAsActiveDescendant(tile: TTile) {
+    if (!tile) {
+      return;
+    }
+    if (tile instanceof FormFieldTile
```
cguglielmo: TileGrid does not know what tiles it contains -> move method to Tile and FormFieldTile, maybe name it markAsActiveDescendantFor($container)

eclipse-scout-core/src/tree/Tree.ts
```diff
@@ -399,6 +399,8 @@ export class Tree extends Widget implements TreeModel {
 
   protected override _render() {
     this.$container = this.$parent.appendDiv('tree');
+    aria.tree(this.$container);
+    aria.multiselectable(this.$container, this.multiCheck);
```
cguglielmo: Move multiselectable  to _renderMultiCheck

eclipse-scout-core/src/tree/TreeNode.ts
```diff
@@ -327,11 +336,19 @@ export class TreeNode implements TreeNodeModel, ObjectWithType {
...
 
+    aria.disabled($node, $node.hasClass('disabled'));
+    aria.toggleExpanded($node, $node.hasClass('expanded'), $node.hasClass('expanded'));
+
     if (!this.parentNode && tree.selectedNodes.length === 0 || // root nodes have class child-of-selected if no node is selected
       tree.isChildOfSelectedNodes(this)) {
       $node.addClass('child-of-selected');
     }
 
+    if (this.parentNode && this.parentNode.childNodes) {
```
cguglielmo: this.parentNode.childNodes cannot be null

eclipse-scout-core/src/index.ts
```diff
@@ -1076,7 +1078,12 @@ export * from './uuidpool/UuidPoolAdapter';
 export * from './uuidpool/UuidPoolEventMap';
 export * from './uuidpool/UuidPoolModel';
 export * from './types';
+export * from './aria';
 export * from './jquery/jquery-scout-types';
+export * from './table/AbstractTableAccessibilityRenderer';
```
cguglielmo: Move to table section

eclipse-scout-chart/src/chart/ChartJsRenderer.ts
cguglielmo: Maybe we should do that, lets discuss it with Fritz.

eclipse-scout-core/src/action/Action.ts
cguglielmo: Actions cannot open popups, only specific actions like menus can do it, so we do not know it here. I would just set togglePressed. Also, it should not be necessary to check isToggleAction and actionStyle, _renderSelected is only called for toggleActions -> 
`aria.pressed(this.$container, this.selected)`

eclipse-scout-core/src/desktop/navigation/DesktopNavigation.ts
```diff
@@ -79,9 +85,47 @@ export class DesktopNavigation extends Widget implements DesktopNavigationModel
...
   }
 
+  protected _addScreenReaderStatus() {
+    this.$screenReaderStatus = this.$container.appendDiv();
+    aria.status(this.$screenReaderStatus);
+    aria.screenReaderOnly(this.$screenReaderStatus);
+  }
+
+  _renderScreenReaderStatus() {
+    if (this.$screenReaderStatus && this.outline) {
```
cguglielmo: Early return would make it a little bit more clear

eclipse-scout-core/src/desktop/viewbutton/ViewMenuPopup.ts
```diff
@@ -97,13 +101,38 @@ export class ViewMenuPopup extends WidgetPopup implements ViewMenuPopupModel {
...
+
+  protected _renderTiles() {
+    this.content.tiles.forEach(tile => {
+      aria.menuitem(tile.$container);
+      this._linkWidgetLabels(tile);
+    });
+  }
+
+  protected _linkWidgetLabels(tile: ViewButtonTile) {
+    let labels = tile.$container.children('.label');
```
cguglielmo: labels should be called $labels because it is a jQuery element

eclipse-scout-core/src/form/fields/button/Button.ts
```diff
@@ -166,13 +167,18 @@ export class Button extends FormField implements ButtonModel {
...
     if (this.menus && this.menus.length > 0) {
+      aria.hasPopup($button);
       this.menus.forEach(menu => this.keyStrokeContext.registerKeyStroke(menu));
       if (this.label || !this.iconId) { // no indicator when _only_ the icon is visible
         let icon = icons.parseIconId(Button.SUBMENU_ICON);
         this.$submenuIcon = $button
           .appendSpan('submenu-icon')
           .text(icon.iconCharacter);
+        aria.img(this.$submenuIcon);
+        aria.labelElement(this.$submenuIcon, '');
```
cguglielmo: Shouldn't replace parameter be true?

eclipse-scout-core/src/form/fields/button/Button.ts
```diff
@@ -239,11 +252,14 @@ export class Button extends FormField implements ButtonModel {
   togglePopup() {
     if (this.popup) {
       this.popup.close();
+      aria.expanded(this.$field, false);
```
cguglielmo: Should not be necessary, will be done by the destroy handler

eclipse-scout-core/src/form/fields/groupbox/GroupBox.ts
```diff
@@ -424,13 +425,21 @@ export class GroupBox extends CompositeField implements GroupBoxModel {
       return;
     }
     this.$label = this.$title.appendDiv('label');
+    if (this._computeTitleVisible()) { // add it as a heading if its not the invisible main box
```
cguglielmo, line 429: Actually, the label can be change dynamically, but then we would have to recompute the levels for all headers, so we decided not to handle it, right?

eclipse-scout-core/src/form/fields/smartfield/SmartField.ts
```diff
@@ -222,6 +242,27 @@ export class SmartField<TValue> extends ValueField<TValue> implements SmartField
...
+      if (result.numLookupRows) {
+        this.$screenReaderStatus.appendSpan().addClass('sr-lookup-row-count').text(this.session.text('ui.NumProposals', result.numLookupRows));
+      } else {
+        this.$screenReaderStatus.appendSpan().addClass('sr-lookup-row-count').text(this.session.text('ui.NoProposals'));
+      }
+    }
+  }
+
+  /**
+   * screen reader status should always be cleared if chooser is not visible, so everytime the chooser shows the status is read to the
```
cguglielmo: "shows the status is read" -> "shows the status, it will be read"

eclipse-scout-core/src/form/fields/smartfield/SmartField.ts
```diff
@@ -1021,9 +1065,20 @@ export class SmartField<TValue> extends ValueField<TValue> implements SmartField
     this._pendingOpenPopup = false;
     if (this.popup) {
       this.popup.close();
+      this._resetPopupAriaProperties();
```
cguglielmo: No need to do it here, the remove handler will be called and will do it

eclipse-scout-core/src/form/fields/smartfield/SmartField.ts
```diff
@@ -1021,9 +1065,20 @@ export class SmartField<TValue> extends ValueField<TValue> implements SmartField
...
     }
   }
 
+  /**
+   * resets the aria properties that were changed by the pop up to indicate that the pop up is not open
```
cguglielmo: Start with Uppercase letter

eclipse-scout-core/src/form/fields/stringfield/StringField.ts
```diff
@@ -191,6 +191,8 @@ export class StringField extends BasicField<string> {
   override addIcon($parent?: JQuery) {
     this.$icon = fields.appendIcon(this.$container)
       .on('click', this._onIconClick.bind(this));
+    aria.img(this.$icon);
```
cguglielmo: This is actually a button, it will be shown if hasAction is true

eclipse-scout-core/src/form/fields/wizard/WizardProgressField.ts
```diff
@@ -141,6 +151,34 @@ export class WizardProgressField extends FormField implements WizardProgressFiel
     this.invalidateLayoutTree(false);
   }
 
+  protected _renderScreenReaderStatus(statusContainer: JQuery) {
```
cguglielmo: statusContainer should be called $statusContainer

eclipse-scout-core/src/form/fields/FormField.ts
```diff
@@ -1107,11 +1110,15 @@ export class FormField extends Widget implements FormFieldModel {
...
    */
   protected _linkWithLabel($element: JQuery) {
-    if (!this.$label || !$element) {
+    if (strings.empty(this.label)) { // no label, do not link field to nbsp
       return;
     }
 
-    fields.linkElementWithLabel($element, this.$label);
+    if (this.labelPosition === FormField.LabelPosition.ON_FIELD) {
+      aria.labelElement($element, this.label);
```
cguglielmo: label can be set later. If that happens, the label won't be linked

eclipse-scout-core/src/form/Form.ts
```diff
@@ -1278,6 +1291,7 @@ export class Form extends Widget implements FormModel, DisplayParent {
     if (this.$header) {
       this.$title.text(this.title);
       this.$header.toggleClass('no-title', !this.title && !this.subTitle);
+      aria.linkElementWithLabel(this.$container, this.$title);
```
cguglielmo: This belongs to _renderHeader. _renderTitle is called whenever the title property changes, but the linking can be done initially when this.$title is created

eclipse-scout-core/src/form/Form.ts
```diff
@@ -1292,6 +1306,7 @@ export class Form extends Widget implements FormModel, DisplayParent {
     if (this.$header) {
       this.$subTitle.text(this.subTitle);
       this.$header.toggleClass('no-title', !this.title && !this.subTitle);
+      aria.linkElementWithLabel(this.$container, this.$subTitle, aria.LABELLED_BY_INSERT_POSITION.BACK);
```
cguglielmo: See comment in renderTitle

cguglielmo (changes requested):

eclipse-scout-chart/src/chart/ChartJsRenderer.ts
jwinkens: I removed the part and stored it to our share. A meeting with Fritz is scheduled to sketch a solution.

eclipse-scout-core/src/action/Action.ts
jwinkens: Code was split up to subclasses to handle expanded/pressed individually if it is a menu or a button.

eclipse-scout-core/src/form/fields/FormField.ts
jwinkens: label moved to _renderLabel

eclipse-scout-core/src/aria.ts
jwinkens: moved to AbstractSvgRenderer, Comment added

eclipse-scout-core/src/aria.ts
jwinkens: As discussed, methods are replaced with a single method. If null is inserted as value, the attribute is removed.

eclipse-scout-core/src/aria.ts
jwinkens: String literal type added, individual role methods removed.

eclipse-scout-core/src/modeselector/ModeSelector.ts
jwinkens: Fixed by overriding _renderTabbable in Mode.ts and not checking for enabled there

eclipse-scout-core/src/modeselector/Mode.ts
jwinkens: It seems to be necessary. Without it, after clicking a mode, Keyboard navigation does not work. In my case if the tabs below the mode selector were focused before, and then a mode is clicked, keyboard navigation will act on the tabs and not the mode.

eclipse-scout-core/src/table/Table.ts
jwinkens: Active-descendant logic was moved to keystrokes/mousehandler to correctly reflect the active row. It is correct that it only points to a single row on multi-select - however until now there were cases where it was pointing to the wrong row.

I did not implement removal on ctrl-click because it seemed wrong. The active descendant is the current action row, regardless if it is selected or not, because any subsequent keyboard navigation will start there. E.g. if 3 rows are selected and the middle one is ctrl+clicked and 2 remain selected. Is there realy no active descendant? Keyboard navigation will start at the unselected row. Further, low-vision user may profit from the unselected row being announced. I also can open the context menu on that unselected row -> I think we should stick with not removing active-descendant.

eclipse-scout-core/src/table/TableHeader.ts
jwinkens: removed

eclipse-scout-core/src/form/fields/groupbox/GroupBox.ts
jwinkens: Correct

eclipse-scout-core/src/menu/Menu.ts
```diff
@@ -481,6 +488,7 @@ export class Menu extends Action implements MenuModel {
   _closePopup() {
     if (this.popup && !this.popup.isRemovalPending()) {
       this.popup.close();
+      aria.removeControls(this.$container);
```
cguglielmo: This can be removed because it is done in the destroy handler

eclipse-scout-core/src/table/columns/Column.ts
```diff
@@ -335,7 +335,19 @@ export class Column<TValue = string> extends PropertyEventEmitter implements Col
...
-    return '<div class="' + cssClass + '" style="' + style + '">' + content + '</div>';
+    let ariaAttributes = '';
+    if (this.table.accessibilityRenderer && strings.hasText(this.table.accessibilityRenderer.cellRole)) {
+      ariaAttributes = ' role="' + this.table.accessibilityRenderer.cellRole + '"';
+    }
+    // Set the label of the cell to header name + cell content. The reference to cell content is needed, because
+    // without it screen readers may only announce the header name without the cell content. If there is no header
+    // to reference, we do not need to reference the cell either, because screen readers will announce the cell
+    // content naturally if there is no aria-labelledby
+    if (strings.hasText(this.table.header.headerLabelId)) {
```
cguglielmo: this throws an error if there is no header (e..g when opening a smartfield)

eclipse-scout-core/src/menu/Menu.ts
```diff
@@ -153,8 +153,9 @@ export class Menu extends Action implements MenuModel {
     this._renderSubMenuIcon();
   }
 
-  protected _renderActionStyle() {
+  protected override _renderActionStyle() {
     this.$container.toggleClass('menu-button', this.isButton() && !this.overflown);
+    aria.role(this.$container, this.separator ? 'separator' : 'menuitem');
```
cguglielmo: Maybe we should use menuitemcheckbox if it is a toggle action, then we could use aria-checked?

eclipse-scout-core/src/table/TableHeader.ts
```diff
@@ -371,6 +380,7 @@ export class TableHeader extends Widget implements TableHeaderModel {
     this._renderColumnLegacyStyle(column);
     this._renderColumnHeaderMenuEnabled(column);
     this._renderColumnHorizontalAlignment(column);
+    this._renderTooltipText(column);
```
cguglielmo: Please name it _renderColumnTooltipText

eclipse-scout-core/src/form/Form.ts
```diff
@@ -281,6 +286,9 @@ export class Form extends Widget implements FormModel, DisplayParent {
       return;
     }
     let modal = this.modal;
+    if (modal) {
+      aria.role(this.$container, 'dialog');
```
cguglielmo: remove dialog role if modal changes to false

eclipse-scout-core/src/jquery/jquery-scout.js
```diff
@@ -521,12 +524,16 @@ $.fn.appendIcon = function(iconId, cssClass) {
     icon = icons.parseIconId(iconId);
   }
   if (icon.isFontIcon()) {
-    return this.makeSpan(cssClass, icon.iconCharacter)
+    let $icon = this.makeSpan(cssClass, icon.iconCharacter)
       .addClass('icon')
       .addClass(icon.appendCssClass('font-icon'))
       .appendTo(this);
+    aria.role($icon, 'img');
+    aria.label($icon, iconId.altText, true);
```
cguglielmo: Shouldn't this be icon.altText instead of iconId.altTexT?

eclipse-scout-core/src/menu/form/field/FormFieldMenu.ts
```diff
@@ -24,6 +24,7 @@ export class FormFieldMenu extends Menu implements FormFieldMenuModel {
 
   protected override _render() {
     this.$container = this.$parent.appendDiv('menu-item');
+    aria.role(this.$container, 'menuitem');
```
cguglielmo: This should not be necessary because it is set by renderActionStyle

eclipse-scout-core/src/menu/EllipsisMenu.ts
```diff
@@ -46,6 +46,12 @@ export class EllipsisMenu extends Menu implements EllipsisMenuModel {
     this._renderHidden();
   }
 
+  protected override _renderSelected() {
```
cguglielmo, line 49: Is this necessary? Menu._renderSelection already does it correctly imho

eclipse-scout-core/src/modeselector/Mode.ts
cguglielmo: ok, it is because there is no tabindex set yet.

eclipse-scout-core/src/modeselector/ModeSelector.ts
cguglielmo: so we can remove this method, can't we?

eclipse-scout-core/src/modeselector/ModeSelector.ts
```diff
@@ -108,6 +176,8 @@ export class ModeSelector<TModeRef = any> extends Widget implements ModeSelector
       this._updateMarkers();
     } else if (event.propertyName === 'enabled') {
       this._updateSlider();
+    } else if (event.propertyName === 'focused') {
+      this.setFocused(event.newValue);
```
cguglielmo: Really necessary?

eclipse-scout-core/src/modeselector/ModeSelector.ts
```diff
@@ -68,12 +116,16 @@ export class ModeSelector<TModeRef = any> extends Widget implements ModeSelector
       mode.on('propertyChange', this._modePropertyChangeHandler);
       if (mode.selected) {
         this.setSelectedMode(mode);
+        if (mode.focused) {
```
cguglielmo: Really necessary?

eclipse-scout-core/src/modeselector/ModeSelectorRightOrDownKeyStroke.ts
```diff
@@ -0,0 +1,40 @@
...
+      focusedMode = $(event.target).data('mode');
+
+    this.field.modes.some(mode => {
+      if (fieldBefore && mode.enabledComputed && mode.visible) {
+        mode.setSelected(true);
+        mode.focus();
+        mode.$container.addClass('keyboard-navigation');
+        return true;
+      }
+      if (mode === focusedMode && mode.enabledComputed && mode.visible) {
```
cguglielmo: This keystroke does not work if selected mode is disabled

eclipse-scout-core/src/table/controls/TableControl.ts
```diff
@@ -143,11 +143,15 @@ export class TableControl extends Action implements TableControlModel {
...
-  protected override _renderSelected(selected?: boolean, options?: { closeWhenUnselected?: boolean; animate?: boolean }) {
+  protected override _renderSelected(selected?: boolean, options?: {
+    closeWhenUnselected?: boolean;
+    animate?: boolean;
+  }) {
     selected = scout.nvl(selected, this.selected);
     options = $.extend({}, {closeWhenUnselected: true}, options);
 
     this.$container.select(selected);
+    aria.pressed(this.$container, selected);
```
cguglielmo: Is pressed allowed if it is not a button?

eclipse-scout-core/src/table/Table.ts
```diff
@@ -2783,6 +2820,7 @@ export class Table extends Widget implements TableModel {
     if (changedRows.length === 0) {
       return;
     }
+    changedRows.forEach(row => aria.expanded(row.$row, row.expanded));
```
cguglielmo: This should not be necessary, viewport will be rendered again in renderRowDelta

eclipse-scout-core/src/table/TableHeaderMenuButton.ts
```diff
@@ -46,20 +49,47 @@ export class TableHeaderMenuButton extends Action implements TableHeaderMenuButt
     this._renderToggleAction();
   }
 
-  // Show 'remove' text when button is already selected
+  protected override _renderActionStyle() {
+    aria.role(this.$container, 'button');
```
cguglielmo: I think we should move this to Action.ts because it seems to be a reasonable default

eclipse-scout-core/src/table/TableHeaderMenuButton.ts
```diff
@@ -46,20 +49,47 @@ export class TableHeaderMenuButton extends Action implements TableHeaderMenuButt
...
+  }
+
+  protected override _renderSelected() {
+    super._renderSelected();
+    aria.pressed(this.$container, this.selected || null);
```
cguglielmo: See comment before -> move to action.ts. Why does expanded need to be set to null explicitly?

cguglielmo (changes requested):
It looks like something went wrong with the rebase on 24.1. The change contains unrelated commits with the wrong committer

eclipse-scout-core/src/menu/EllipsisMenu.ts
jwinkens: Is necessary because "hasPopup" in super is false for ellipsis menu.

eclipse-scout-core/src/modeselector/ModeSelector.ts
jwinkens: As you mentioned, this method was not working correctly because the child actions enabled state was not computed yet. It worked for radio group because the method was called twice, which is not done for modeselector. I moved calling of the method from (i think it was) selectEnabled to recomputeEnabled. At this place, child action enabled state is already computed

cguglielmo (approved):

PR #752, opened 2023-10-05
Form: move error handling out of busy handling

fschinkel:
The busy handling of the form will set the desktop busy until the given promise completes (resolved or rejected). This might lead to deadlocks if errors occur inside this given promise that are handled by a message box which prevents the promise from completing until a button is clicked. If such a message box is rendered before the busy indicator is rendered, the glass pane of the busy indicator will make it impossible to click any button of the message box, and therefore it is not possible to leave the busy state.
To solve this problem move the error handling out of the scope that is covered by the busy handling. This leads to the busy handling being stopped when an error occurs and therefore the message box is not blocked by a glass pane.

362418

### DIFF
--- a/eclipse-scout-core/src/form/Form.ts
+++ b/eclipse-scout-core/src/form/Form.ts
@@ -363,7 +363,14 @@ export class Form extends Widget implements FormModel, DisplayParent {
     if (!allowReload && this.formLoaded) {
       return $.resolvedPromise();
     }
-    return this._withBusyHandling(() => this.lifecycle.load());
+    try {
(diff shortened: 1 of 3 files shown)


eclipse-scout-core/src/form/Form.ts
```diff
@@ -403,11 +411,11 @@ export class Form extends Widget implements FormModel, DisplayParent {
         })
         .catch(error => {
```
mvilliger: Maybe use a finally instead of a catch with a rethrow?

mvilliger (approved):

PR #761, opened 2023-10-20
Notification: Adjust CSS-handling of empty message string

gabrielstalder:
To avoid a collapsing notification box, a zero-width whitespace is inserted into the notification's after pseudo-element. This leads to an empty new line if the notification message is new element as-well. A min-height instead of a zero-width  whitespace is now used.


bschwarzent (approved):
Thanks for the fix!

PR #774, opened 2023-10-30
Add overflow-styles to calendar popups

egloff:
Set ellipsis and no-wrap for title and ellipsis for description.


egloff:
Before:

image

After (title on one line with ellipsis, description with break-word):

image

PR #781, opened 2023-11-06
FocusContext: automatically select text content of input fields

bschwarzent:
When tabbing through a form, the browser automatically natively selects the text content of <input> fields. When we intercept the default tabbing order (e.g. because we have reached the end of a focus context and want to wrap around to the beginning), the same behavior should be applied. This is achieved by calling select() after focus().

365935

### DIFF
--- a/eclipse-scout-core/src/focus/FocusContext.js
+++ b/eclipse-scout-core/src/focus/FocusContext.js
@@ -8,7 +8,7 @@
  * Contributors:
  *     BSI Business Systems Integration AG - initial API and implementation
  */
-import {filters, focusUtils, graphics, keys, Point, scrollbars} from '../index';
+import {filters, focusUtils, graphics, keys, objects, Point, scrollbars} from '../index';
 import $ from 'jquery';
 
 /**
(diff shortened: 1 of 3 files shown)


eclipse-scout-core/src/focus/FocusContext.js
```diff
@@ -286,6 +290,10 @@ export default class FocusContext {
     elementToFocus.focus({
       preventScroll: scout.nvl(options.preventScroll, false)
     });
+    // If possible and requested, select the text content
+    if (elementToFocus instanceof HTMLInputElement && options.selectText) {
```
cguglielmo: Unfortunately, instanceof does not work for these HTML elements if used in a popup window and probably also in iframes.

cguglielmo (changes requested):

eclipse-scout-core/src/focus/FocusContext.js
bschwarzent: The code now checks if a functions called "select" exists on the element instead.

cguglielmo (approved):

PR #782, opened 2023-11-07
Fix Layout for ModeSelector

nsteger:
Fix ModeSelector width calculation when modes have a marigin set

318434


cguglielmo (commented):
There is a typo in the commit comment:
marigin -> margin

nsteger:
> There is a typo in the commit comment: marigin -> margin

Fixed

PR #783, opened 2023-11-07
DefaultFilesystemWebResourceRootContributor: improve detection of root module

stephan-merkli:
338662


mvilliger (approved):
Looks good.
Instead of first removing .dev and then .app it would also be possible to declare the suffixes in a list or array and then removing them in a loop. This would reduce the duplicate code of the current implementation and would be better extendable in the future if more suffixes are added. Just an idea.

stephan-merkli:
Thank you for the review. That's an improvement that came into my mind too, but didn't implement it in the end because only having these two suffixes yet (.dev and .app). In case there will another case to support, the code should be improved further, I agree.

PR #785, opened 2023-11-13
Form: add FormValidators

fschinkel:
A FormValidator is a function that gets the Form and returns a Promise
containing a Status. The Form waits for all validators to be completed
and considers their results while determining if it is valid.
This allows to add asynchronous validators to the Form that e.g. do a
REST call.

344799

### DIFF
--- a/eclipse-scout-core/src/form/Form.ts
+++ b/eclipse-scout-core/src/form/Form.ts
@@ -8,9 +8,9 @@
  * SPDX-License-Identifier: EPL-2.0
  */
 import {
-  AbortKeyStroke, App, aria, AriaLabelledByInsertPosition, BusyIndicatorOptions, Button, ButtonSystemType, DialogLayout, DisabledStyle, DisplayParent, DisplayViewId, EnumObject, ErrorHandler, Event, EventHandler, FileChooser,
+  AbortKeyStroke, App, aria, AriaLabelledByInsertPosition, arrays, BusyIndicatorOptions, Button, ButtonSystemType, DialogLayout, DisabledStyle, DisplayParent, DisplayViewId, EnumObject, ErrorHandler, Event, EventHandler, FileChooser,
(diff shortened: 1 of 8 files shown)


eclipse-scout-core/src/form/Form.ts
```diff
@@ -613,7 +616,31 @@ export class Form extends Widget implements FormModel, DisplayParent {
    * @internal
    */
   _validate(): Status | JQuery.Promise<Status> {
```
cguglielmo: I think it would be better to move the default logic to another function, e.g. _onLifecycleValidate Projects may already override this method without calling super, which breaks the new behavior. Also, js doc needs to be adjusted.

eclipse-scout-core/src/form/Form.ts
```diff
@@ -613,7 +616,31 @@ export class Form extends Widget implements FormModel, DisplayParent {
...
-    return Status.ok();
+    if (!this.validators?.length) {
```
cguglielmo: Don't overly use ?. If you initialize with [], it is good enough. If a developer replaces the variable with null it's his fault and fine if his app breaks.

eclipse-scout-core/src/form/Form.ts
```diff
@@ -613,7 +616,31 @@ export class Form extends Widget implements FormModel, DisplayParent {
...
+    this.setValidators(validators);
+  }
+
+  removeValidator(validator: FormValidator) {
+    let validators = this.validators.slice();
+    arrays.remove(validators, validator);
+    this.setValidators(validators);
+  }
+
+  setValidators(validators: FormValidator | FormValidator[]) {
```
cguglielmo, line 680: add   /** @see FormModel.validators */

cguglielmo (changes requested):

mvilliger:
run tests

PR #790, opened 2023-11-22
SmartFieldTouchPopup: Delegate lookup events to source smart field

gabrielstalder:
The SmartField has its own touch popup implementation. Internally, the popup creates a clone of the original SmartField. Until now, any lookup event listeners on the original field have been ignored, as the clone does not delegate the events to the original smart field. This change adjusts this behaviour. The SmartFieldTouchPopup now delegates the events 'prepareLookupCall' and 'lookupCallDone' to the original  smart field.

366351

### DIFF
--- a/eclipse-scout-core/test/form/fields/smartfield/SmartFieldSpec.ts
+++ b/eclipse-scout-core/test/form/fields/smartfield/SmartFieldSpec.ts
@@ -8,7 +8,8 @@
  * SPDX-License-Identifier: EPL-2.0
  */
 import {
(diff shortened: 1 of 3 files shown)


eclipse-scout-core/test/form/fields/smartfield/SmartFieldSpec.ts
```diff
@@ -387,6 +388,36 @@ describe('SmartField', () => {
...
+      };
+      field.on('prepareLookupCall', onPrepareLookupCall.bind(field));
+      field.on('lookupCallDone', onLookupCallDone.bind(field));
+      field.render();
+      jasmine.clock().tick(500);
+      JQueryTesting.triggerClick(field.$field);
+      jasmine.clock().tick(500);
+
+      let popup = field.popup as SpecSmartFieldTouchPopup;
+      const oldPrepareLookupCallCounter = prepareLookupCallCounter;
```
awegmueller: I would prefer "let" here. More because of consistency, than technical correctness.

PR #791, opened 2023-11-22
FormController: do not register forms twice on their displayParent

fschinkel:
A form that is already contained in its displayParents views or dialogs must not be registered again when it is shown.
This happens for example if a JsForm is open and the browser is reloaded. The JsForm will be part of the desktops views on session startup. As the JsForm calls Form.show() after it is loaded the FormController tries to register it a second time on the desktop. If a form is registered multiple times on its displayParent only one occurrence is removed when the form is destroyed. This will break e.g. open exclusive as there might be a matching form which can not be displayed as it is already destroyed.
Add the same logic to FileChooserController and MessageBoxController even if there are no known errors so far.

367169

### DIFF
--- a/eclipse-scout-core/src/filechooser/FileChooserController.ts
+++ b/eclipse-scout-core/src/filechooser/FileChooserController.ts
@@ -67,7 +67,7 @@ export class FileChooserController {
     if (fileChooser.rendered) {
       return;
     }
-    if (register) {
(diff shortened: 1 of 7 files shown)


mvilliger (commented):
Looks good to me. 
Maybe @cguglielmo would like to have a look as well?

PR #798, opened 2023-11-27
Table: error status is being display on wrong cell

lardser:
If a table has invisible columns, cells' error status might be displayed on the wrong cell.

367747


cguglielmo:
Jenkins run tests

PR #799, opened 2023-11-27
Find pages method on outline

christiandoebele:
Provide a new outline method for finding all pages of a given page class

363930


egloff (approved):
Looks good

PR #809, opened 2023-12-05
A11y: Fix tests after merge

jwinkens:


mvilliger:
run tests

eclipse-scout-core/src/action/Action.ts
```diff
@@ -184,10 +184,9 @@ export class Action extends Widget implements ActionModel {
       } else {
         this.$text.text(text);
       }
-    } else if (text) {
+    } else {
       // add as label for screen readers
       aria.label(this.$container, text);
```
cguglielmo, line 191: If text visibility changes dynamically, aria label is set if text will be invisible but not removed when text gets visible again. I don't know how screen readers react to this, but I think there should never be an aria-label if text is visible.

eclipse-scout-core/src/table/TableHeader.ts
```diff
@@ -456,7 +456,7 @@ export class TableHeader extends Widget implements TableHeaderModel {
     let accessibleStateText = '';
     if (column.sortActive) {
       let sortDirection = column.sortAscending ? 'asc' : 'desc';
-      accessibleStateText += this.session.text('ui.Sorting') + ' ' + (column.sortAscending ? this.session.text('ui.ascending') : this.session.text('ui.descending'));
+      accessibleStateText += ' ' + this.session.text('ui.Sorting') + ' ' + (column.sortAscending ? this.session.text('ui.ascending') : this.session.text('ui.descending'));
```
cguglielmo: If this is only necessary for the spec, I would rather adjust the spec. If screen readers needs it, its ok.

eclipse-scout-core/src/action/Action.ts
jwinkens: aria-label is now removed if text becomes visible

eclipse-scout-core/src/table/TableHeader.ts
jwinkens: removed, spec changed to test `${column1.text}sorting ascending`.

Screen reader does not need it, it will treat each DOM node as its own navigatable text. This may or may not change if the table header ever becomes tabbable.

PR #816, opened 2023-12-12
Activate outline without active form check

christiandoebele:
The active ScoutJS form is not tracked on the UI server. To allow for the outline activation with active ScoutJS form the check for an existing active form is removed.


cguglielmo:
run tests

PR #817, opened 2023-12-13
UploadRequestHandler: Switch multipart request handling to servlet api

paolobazzi:
Java EE servlet API offers access to parts of a multipart/form-data request since Servlet API 3.0. Remove Apache commons-fileupload library and switch UploadRequestHandler to Java EE API.

354734

### DIFF
--- a/org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/UploadRequestHandler.java
+++ b/org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/UploadRequestHandler.java
@@ -28,10 +28,6 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import org.apache.commons.fileupload2.core.FileItemInput;
-import org.apache.commons.fileupload2.core.FileItemInputIterator;
-import org.apache.commons.fileupload2.core.FileUploadException;
-import org.apache.commons.fileupload2.jakarta.JakartaServletFileUpload;
 import org.eclipse.scout.rt.platform.BEANS;
 import org.eclipse.scout.rt.platform.Order;
 import org.eclipse.scout.rt.platform.config.CONFIG;
@@ -61,12 +57,15 @@ import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import jakarta.servlet.ServletException;
 import jakarta.servlet.ServletResponse;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
+import jakarta.servlet.http.Part;
 
 /**
- * This handler contributes to the {@link UiServlet} as the POST handler for /upload
+ * This handler contributes to the {@link UiServlet} as the POST handler for /upload.<br>
+ * See Session.uploadFiles() method in {@code Session.ts}.
  */
 @Order(4520)
 public class UploadRequestHandler extends AbstractUiServletRequestHandler {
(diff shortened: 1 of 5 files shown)


org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/UploadRequestHandler.java
```diff
@@ -139,7 +138,18 @@ public boolean handlePost(final HttpServletRequest req, final HttpServletRespons
     return true;
   }
 
-  protected void handleUploadFileRequest(IUiSession uiSession, HttpServletRequest httpServletRequest, HttpServletResponse httpServletResponse, String targetAdapterId) throws IOException, FileUploadException {
+  protected boolean isMultipartContent(HttpServletRequest request) {
+    if (!"POST".equalsIgnoreCase(request.getMethod())) {
```
bschwarzent: If this method is only called from handlePost(), this first check could be omitted.

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/UploadRequestHandler.java
```diff
@@ -243,21 +249,25 @@ protected void readUploadData(HttpServletRequest httpReq, IUploadable uploadable
...
+          .withContentType(detectContentType(filename, part, content))
           .withContent(content)
           .build();
       verifyFileSafety(res);
       verifyFileIntegrity(res);
 
-      if (item.isFormField()) {
+      // properties are sent as form fields without file name by UI (see Session.ts)
+      boolean formField = StringUtility.isNullOrEmpty(part.getSubmittedFileName());
+      if (formField) {
```
bschwarzent: Could be inlined

```suggestion
      if (StringUtility.isNullOrEmpty(part.getSubmittedFileName())) {
```

bschwarzent (commented):
Looks good!

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/UploadRequestHandler.java
```diff
@@ -359,6 +369,18 @@ protected void verifyFileName(Set<String> validFileExtensions, String filename,
...
 
+  /**
+   * Checks if the uploaded file exceeds the maximum allowed upload size for given {@code uploadable}.
+   *
+   * @throws RejectedResourceException
+   *           when size of part exceeds the allowed upload size
+   */
+  protected void verifyMaximumUploadSize(IUploadable uploadable, Part part) {
+    if (part.getSize() > uploadable.getMaximumUploadSize()) {
+      throw new RejectedResourceException("The field {} exceeds its maximum permitted size of {} bytes.", part.getName(), uploadable.getMaximumUploadSize());
```
matthiaso, line 376: Can we rely on part.getSize? The previous implementation did check the provided size but also created a LimitedInputStream which limited the actual number of bytes read from the stream to the maximum size.

matthiaso (approved):

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/UploadRequestHandler.java
paolobazzi: Done

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/UploadRequestHandler.java
paolobazzi: part.getSize() seems to be reliable, at least within my local tests.

Additionally we could also limit the IOUtility call: 

```
content = IOUtility.readBytes(in, (int) uploadable.getMaximumUploadSize());
```
This would lead to a similar limitation as the former LimitedInputStream.
Drawback: Implementation in IOUtilty seems a little bit less performant when a limit size is used.

PR #818, opened 2023-12-14
367178 Improve BEANS.get performance, fix 2 concurrency issues

aschwanden:
- After setup, PlatformImplentor.getBeanManager no longer needs a lock
- Removed lock from BeanHierarchy which is guarded and only accessed by BeanManagerImplementor (initialize could now be called concurrently, but cannot cause any harm)
- Add missing lock to BeanManagerImplementor#getAllBeans
- Remove synchronized keyword from BeanManagerImplementor#unregisterBean (does not make sense to use 2 locking variations, other methods use RRWL)

=> under non-initializing circumstances, only 1 lock now remains for BEANS.get, instead of 3. Which saves about 20% access time, says the profiler at least.


mvilliger (approved):
Looks good. Thanks a lot!

PR #820, opened 2023-12-19
P_GlobalTrustManager: Install using platform listener

matthiaso:
Platform listener must run early during start-up to ensure trust manager
 is available to every following call.

369717

### DIFF
--- a/org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/GlobalTrustManager.java
+++ b/org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/GlobalTrustManager.java
@@ -37,6 +37,10 @@ import javax.net.ssl.X509TrustManager;
 
 import org.eclipse.scout.rt.platform.ApplicationScoped;
 import org.eclipse.scout.rt.platform.BEANS;
+import org.eclipse.scout.rt.platform.IPlatform.State;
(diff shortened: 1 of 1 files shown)


org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/GlobalTrustManager.java
```diff
@@ -224,7 +228,8 @@ protected List<X509Certificate> getTrustedCertificatesInRemoteFiles() {
       return remoteFilesToCertificates(certRemoteFiles);
     }
     catch (RuntimeException e) {
-      LOG.error("Could not access folder '{}' to import trusted certificates.", PATH_CERTS, e);
+      RuntimeException e0 = LOG.isDebugEnabled() ? e : null;
```
paolobazzi, line 231: Is this "ok" to happen? If remote file service is available it should return a value?

paolobazzi (approved):

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/GlobalTrustManager.java
matthiaso: Yes, but remote file server may not be configured properly or there may be not PATH_CERTS path on remote file system.

PR #821, opened 2023-12-19
Add test: nested raw do

fschinkel:
- serialization removes _type of inner do if attribute is of type DoEntity

- serialization uses unqualified keys if part of an inner do that is set as attribute of type DoEntity/IDoEntity


paolobazzi:
Attempt to fix for release 23/2, see https://github.com/eclipse-scout/scout.rt/pull/835

paolobazzi:
Solved by 
* 23.2 https://github.com/eclipse-scout/scout.rt/pull/835
* 24.1 https://github.com/eclipse-scout/scout.rt/pull/844

PR #835, opened 2024-01-04
Data Objects: Fix serialization of nested raw attributes

paolobazzi:
This change fixes serialization of data object with attributes declared using raw data object types (e.g. IDoEntity, DoEntity or IDataObject). Those attributes where serialized without type information before this fix.

This change also introduces a DataObjectSerializer used when an object of declared type IDataObject is serialized and itself delegates to an appropriate serializer.

370640

### DIFF
--- a/org.eclipse.scout.rt.jackson.test/src/test/java/org/eclipse/scout/rt/jackson/dataobject/JsonDataObjectsSerializationTest.java
+++ b/org.eclipse.scout.rt.jackson.test/src/test/java/org/eclipse/scout/rt/jackson/dataobject/JsonDataObjectsSerializationTest.java
@@ -123,7 +123,10 @@ import org.eclipse.scout.rt.jackson.dataobject.fixture.TestStringHolderPojo;
 import org.eclipse.scout.rt.jackson.dataobject.fixture.TestStringPojo;
 import org.eclipse.scout.rt.jackson.dataobject.fixture.TestSubPojo;
 import org.eclipse.scout.rt.jackson.dataobject.fixture.TestThrowableDo;
+import org.eclipse.scout.rt.jackson.dataobject.fixture.TestTypedUntypedInnerAbsDo;
+import org.eclipse.scout.rt.jackson.dataobject.fixture.TestTypedUntypedInnerDataObjectIfcDo;
 import org.eclipse.scout.rt.jackson.dataobject.fixture.TestTypedUntypedInnerDo;
+import org.eclipse.scout.rt.jackson.dataobject.fixture.TestTypedUntypedInnerIfcDo;
 import org.eclipse.scout.rt.jackson.dataobject.fixture.TestTypedUntypedOuterDo;
 import org.eclipse.scout.rt.jackson.dataobject.fixture.TestVersionedDo;
 import org.eclipse.scout.rt.jackson.dataobject.fixture.TestWithEmptyTypeNameDo;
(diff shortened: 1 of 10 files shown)


org.eclipse.scout.rt.jackson.test/src/test/java/org/eclipse/scout/rt/jackson/dataobject/JsonDataObjectsSerializationTest.java
```diff
@@ -3798,6 +3801,65 @@ public void testDeserializeTypedPojo() throws Exception {
...
+  @Test
+  public void testSerializeDeserializeNestedRawDoTypes() throws Exception {
+    String inputJson = readResourceAsString("TestNestedRawDo.json");
+    TestNestedRawDo testDo = s_dataObjectMapper.readValue(inputJson, TestNestedRawDo.class);
+
+    TestNestedRawDo expected = BEANS.get(TestNestedRawDo.class)
+        .withDoEntity(BEANS.get(TestTypedUntypedInnerDo.class)
+            .withIId(FixtureStringId.of("qualified-string-id-0"))
+            .withStringId(FixtureStringId.of("unqualified-string-id-0")))
+        .withIDataObject(BEANS.get(TestTypedUntypedInnerDo.class)
```
stephan-merkli: Do we require a test case where the IDataObject is not a DoEntity but a collection instead? We shouldn't do that anyway, usually a DoList, DoCollection... will be defined and not a DoValue, so would be okay for me to leave that one out as a test case.

org.eclipse.scout.rt.jackson/src/main/java/org/eclipse/scout/rt/jackson/dataobject/ScoutDataObjectSerializerProvider.java
```diff
@@ -64,6 +64,9 @@ public JsonSerializer<?> findSerializer(ScoutDataObjectModuleContext moduleConte
     else if (ObjectUtility.isOneOf(rawClass, DoList.class, DoSet.class, DoCollection.class)) {
       return new DoCollectionSerializer<>(moduleContext, type);
     }
+    else if (IDataObject.class.isAssignableFrom(rawClass)) {
```
stephan-merkli: Why is the own serializer required?
When gen.writeObjectField(attributeName, obj); is called in DoEntitySerializer (due to the modified if clause), isn't the serializer determined based on the value type anyway, which is either IDoEntity or one of the collections above?

stephan-merkli:
https://github.com/eclipse-scout/scout.rt/pull/821 may be closed afterwards, correct?

org.eclipse.scout.rt.jackson/src/main/java/org/eclipse/scout/rt/jackson/dataobject/ScoutDataObjectSerializerProvider.java
paolobazzi: The DataObjectSerializer is required if a mapper for IDataObject.class is requested explicitly.
See test: 
org.eclipse.scout.rt.jackson.dataobject.JsonDataObjectsSerializationTest#testSerializeDeserializeDataObject_DoEntity
org.eclipse.scout.rt.jackson.dataobject.JsonDataObjectsSerializationTest#testSerializeDeserializeDataObject_DoList

Unfortunately I was not able to write a test to force an error if this serializer is missing using a DoEntity and a DoValue<IDataObject> attribute, since in this case your input is correct and DoEntitySerializer handles the case.

-> Therefore removed serializer

org.eclipse.scout.rt.jackson.test/src/test/java/org/eclipse/scout/rt/jackson/dataobject/JsonDataObjectsSerializationTest.java
paolobazzi: I tried this case, but since DoList is final, there is no option create a typed object which extends IDataObject/DoList.
Added some test cases with raw DoList (and other raw DOs) as example for non-supported cases

org.eclipse.scout.rt.jackson.test/src/test/java/org/eclipse/scout/rt/jackson/dataobject/JsonDataObjectsSerializationTest.java
```diff
@@ -3806,41 +3806,66 @@ public void testSerializeDeserializeNestedRawDoTypes() throws Exception {
     String inputJson = readResourceAsString("TestNestedRawDo.json");
     TestNestedRawDo testDo = s_dataObjectMapper.readValue(inputJson, TestNestedRawDo.class);
 
+    DoEntity entity2 = BEANS.get(DoEntity.class);
+    // NOTE: This test-case is not a supposed real-world example. An IId as part of an untyped DoEntity is serialized unqualified and may never be deserialized correctly.
```
stephan-merkli, line 3839: DoEntity ws. DoEntityBuilder below, use one or the other, but not both?

paolobazzi: Used both on purpose: DoEntityBuilder returns IDoEntity, therefore for the DoValue<DoEntity> I need the  BEANS.get(DoEntity.class) directly.

PR #838, opened 2024-01-09
HttpProxy: Improve error and cancel handling

matthiaso:
* Cancel outgoing requests if incoming request time out.
* Ignore error for isCommitted if response is already invalidated.

### DIFF
--- a/org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/AlreadyInvalidatedException.java
+++ b/org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/AlreadyInvalidatedException.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.scout.rt.server.commons.servlet;
+
+import java.lang.reflect.Method;
+
+import org.eclipse.scout.rt.platform.exception.PlatformException;
+
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+
+/**
+ * Exception thrown by {@link AbstractHttpServlet} when trying to access {@link HttpServletRequest} or
+ * {@link HttpServletResponse} methods after the response has already been completed.
+ */
+public class AlreadyInvalidatedException extends PlatformException {
+
+  private static final long serialVersionUID = -1;
+
+  public AlreadyInvalidatedException(Method method, Object origin) {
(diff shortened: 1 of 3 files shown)


org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/AlreadyInvalidatedException.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Exception thrown by {@link AbstractHttpServlet} when trying to access {@link HttpServletRequest} or
+ * {@link HttpServletResponse} methods after the response has already been completed.
+ */
+public class AlreadyInvalidatedException extends PlatformException {
+
+  private static final long serialVersionUID = -1;
+
+  public AlreadyInvalidatedException(Method method, Object origin) {
+    super("Access to '{}' is not allowed because {} is no longer valid (request has been completed).", method, (origin instanceof HttpServletRequest ? "HTTP servlet request" : "HTTP servlet response"));
```
paolobazzi: At this place we have no guarantee for the origin typing. 
Suggestion: Just use origin.getClass() as log message instead of hard coded request/response strings

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
```diff
@@ -221,7 +224,10 @@ public void proxy(HttpServletRequest req, HttpServletResponse resp, HttpProxyReq
     }
 
     LOG.trace("Executing request for {}", url);
-    getHttpClientManager().getClient().execute(asyncRequestBuilder.build(), createAsyncResponseConsumer(resp), createHttpContext(), createExecuteCallback(resp, asyncContext));
+    // Execute request and add a listener potentially cancelling the proxy request if async context times out/fails (if incoming request is not available anymore)
```
paolobazzi, line 227: Suggestion: Increase readability by using two lines:
```
    Future<Boolean> future = getHttpClientManager().getClient().execute(asyncRequestBuilder.build(), createAsyncResponseConsumer(resp), createHttpContext(), createExecuteCallback(resp, asyncContext));
    addCancelListener(asyncContext, future);
```

paolobazzi (approved):

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
matthiaso: Applied your suggestion

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/AlreadyInvalidatedException.java
matthiaso: Applied your suggestion (with additional null-check - even though origin also should not be null).

PR #843, opened 2024-01-11
Install ConfigurableProxySelector as default

matthiaso:
371252

### DIFF
--- a/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/proxy/ConfigurableProxySelector.java
+++ b/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/proxy/ConfigurableProxySelector.java
@@ -24,7 +24,11 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.regex.Pattern;
 
+import org.eclipse.scout.rt.platform.BEANS;
 import org.eclipse.scout.rt.platform.Bean;
+import org.eclipse.scout.rt.platform.IPlatform.State;
+import org.eclipse.scout.rt.platform.IPlatformListener;
+import org.eclipse.scout.rt.platform.PlatformEvent;
 import org.eclipse.scout.rt.platform.config.AbstractStringListConfigProperty;
 import org.eclipse.scout.rt.platform.config.CONFIG;
 import org.eclipse.scout.rt.shared.http.ApacheHttpTransportFactory;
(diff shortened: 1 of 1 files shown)


org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/proxy/ConfigurableProxySelector.java
```diff
@@ -188,4 +192,24 @@ public void connectFailed(URI uri, SocketAddress sa, IOException ioe) {
...
+  /**
+   * Install an instance of {@link ConfigurableProxySelector} as default using
+   * {@link ProxySelector#setDefault(ProxySelector)}
+   */
+  public static class ProxySelectorInstaller implements IPlatformListener {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ProxySelectorInstaller.class);
+
+    @Override
+    public void stateChanged(PlatformEvent event) {
```
matthiaso, line 204: Inspired by [MultiSessionCookieStoreInstallListener ](https://github.com/eclipse-scout/scout.rt/blob/c67fb70429f99c3334992bfe596885d65ab9f51c/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/servicetunnel/http/MultiSessionCookieStoreInstallListener.java#L18) which is just [always inactive](https://github.com/eclipse-scout/scout.rt/blob/c67fb70429f99c3334992bfe596885d65ab9f51c/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/servicetunnel/http/MultiSessionCookieStoreInstallListener.java#L23) (non-configurable).

However I decided not to implement the uninstall (restore previous state) on bean manager stopped as with Jetty I suppose I application cannot be restarted anymore (the platform cannot be restarted)? -> Goal: Faster application shutdown.

The [MultiSessionCookieStoreInstaller](https://github.com/eclipse-scout/scout.rt/blob/c67fb70429f99c3334992bfe596885d65ab9f51c/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/servicetunnel/http/MultiSessionCookieStoreInstaller.java#L59) even includes code to ensure there is only one installation which is probably also not necessary if all our applications are run on our own using Jetty?

Back to topic (proxies): If the proxy selector is now installed as default it is even arguable if we should create other instances for our other HTTP clients (e.g. Apache/OkHttp) - or if we should just use system default here?

paolobazzi (approved):

PR #847, opened 2024-01-12
Add streaming support to hashing through DigestInput/OutputStream

brunokoeferli:
Additionally move SecurityUtility.hash(byte[] data, byte[] salt) to new ILegacySecurityProvider. This method exists only for backward compatibility reasons.

370862


mvilliger (approved):
Looks good. Thanks you!

PR #856, opened 2024-01-18
Tree: fix setting of inital tree values

Knicksaas:
When initial values are set and autoCheckChildNodes is enabled, children were not updated. In the init function, no events can be triggered, because the TreeAdapter has not yet attached its widget. Because of this, the updates are performed in the _postCreateWidget function.

363033

### DIFF
--- a/eclipse-scout-core/src/tree/Tree.ts
+++ b/eclipse-scout-core/src/tree/Tree.ts
@@ -301,19 +301,6 @@ export class Tree extends Widget implements TreeModel {
   }
 
   /**
-   * @internal
-   */
-  _postInitNodes() {
-    Tree.visitNodes((node: TreeNode, parentNode: TreeNode) => {
-      // Check all children, in when autoCheckChildren is true
-      if (node.checked && this.autoCheckChildren) {
-        this.checkNodes(node.childNodes, {checked: true});
-      }
-      this._updateMarkChildrenChecked(node);
-    }, this.nodes, null);
-  }
-
-  /**
    * @param autoCheckChildren
    */
   setAutoCheckChildren(autoCheckChildren: boolean) {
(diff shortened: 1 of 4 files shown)


eclipse-scout-core/src/tree/Tree.ts
```diff
@@ -300,6 +300,19 @@ export class Tree extends Widget implements TreeModel {
     ]);
   }
 
+  /**
+   * @internal
+   */
+  _postInitNodes() {
```
cguglielmo: - Because this is only necessary for remote case, it should be implemented in TreeAdapter.ts
- It is only necessary if autoCheckChildren is true -> Check must be outside of visitor
- Is calling updateMarkChildrenChecked necessary? Is this not done by checkNodes itself?
- Is it required to visit further after checking nodes? I think we can return false in that case which will abort visiting.
- Remove the comment or explain why it is necessary to check all children

eclipse-scout-core/src/tree/Tree.ts
```diff
@@ -368,7 +381,10 @@ export class Tree extends Widget implements TreeModel {
       this.checkedNodes.push(node);
     }
     this._initTreeNodeInternal(node, parentNode);
-    this._updateMarkChildrenChecked(node);
+    if (this.initialized) {
+      // Only execute, when Tree has already been initialized
```
cguglielmo: - Why is this necessary? The comment should explain the why. "if (this.initialized)" already shows that it will only be done when tree is initialized.
- There should be no comma after execute

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/tree/ITree.java
```diff
@@ -406,6 +406,8 @@ default void addUITreeListener(TreeListener listener, Integer... eventTypes) {
 
   void setNodesChecked(List<ITreeNode> nodes, boolean checked, boolean enabledNodesOnly, boolean forceUpdateNode);
 
+  void setAllNodesChecked(boolean checked);
```
cguglielmo, line 409: Maybe also add checkAllNodes() and uncheckAllNodes() which call setAllNodesChecked

cguglielmo:
Thank you, I've adjusted it a bit and merged it: https://github.com/eclipse-scout/scout.rt/pull/858

PR #857, opened 2024-01-18
Environment map properties may lead to deadlock on platform start

matthiaso:
370937


org.eclipse.scout.rt.platform/pom.xml
```diff
@@ -40,7 +40,12 @@
     <dependency>
       <groupId>org.quartz-scheduler</groupId>
       <artifactId>quartz</artifactId>
-   </dependency>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.scout.rt</groupId>
```
stephan-merkli, line 50: Add comment why this dependency is added (i.e. "Dependency is onl< required to enable JSON parsing within PropertiesHelper without using IDataObjectMapper, org.eclipse.scout.json must not be used for other scenarios). Something like this.

org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/config/PropertiesHelperTest.java
```diff
@@ -899,36 +784,126 @@ else if ("dbSystemProperties".equals(configUrl)) {
...
 
-    private final Map<String, Map<String, String>> m_propertyValueToMap = new HashMap<>();
+    HashMap<String, String> expectedMap = new HashMap<>();
+    expectedMap.put("key1", "value1");
+    expectedMap.put("key2", null);
+    expectedMap.put("key3", "");
+    expectedMap.put("key4", "1");
+    expectedMap.put("key5", "1");
+    expectedMap.put("key6", "true");
+    expectedMap.put("key7", "{\"key71\":\"value71\",\"key72\":\"value72\",\"key73\":\"value73\"}"); // changed?
```
stephan-merkli: Add reason why not completly identical, e.g. JSON as a value might be reformatted, thus spaces might change...
Add this reason to on the JavaDoc of the parse method

org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/config/PropertiesHelperTest.java
```diff
@@ -899,36 +784,126 @@ else if ("dbSystemProperties".equals(configUrl)) {
...
+  @Test
+  public void testParseJsonReadJsonObjectWithJsonStringProperty() {
+    assertEquals(Map.of(
+            "testKey1", "{\"subKey1\":\"subValue1\",\"subKey2\":\"subValue2\"}",
+            "testKey2", "[\"testValue2Sub1\",\"testValue2Sub2\",\"testValue2Sub3\"]"),
+        testParseJsonInternal("{\"testKey1\": {\"subKey1\": \"subValue1\", \"subKey2\": \"subValue2\"}, \"testKey2\": [\"testValue2Sub1\", \"testValue2Sub2\", \"testValue2Sub3\"]}"));
+  }
+
+  protected Map<String, String> testParseJsonInternal(String propertyValue) {
+    return new PropertiesHelper(Mockito.mock(IPropertyProvider.class)).parseJson(propertyValue);
```
stephan-merkli: Why not just use new PropertiesHelper(null).parseJson(propertyValue)? No reason to use a mock here

stephan-merkli (approved):

PR #861, opened 2024-01-19
Calendar: add resource view

Knicksaas:
349852


Knicksaas:
run test

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -62,6 +38,9 @@ export class Calendar extends Widget implements CalendarModel {
   selectorEnd: Date;
   showDisplayModeSelection: boolean;
   rangeSelectionAllowed: boolean;
+  defaultCalendar: CalendarDescriptor;
+  calendars: CalendarDescriptor[];
+  selectedCalendar: CalendarDescriptor;
```
nsteger: use "Descriptor" in the name of the 3 variables?

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -160,8 +143,9 @@ export class Calendar extends Widget implements CalendarModel {
     this.$list = null;
     this.$progress = null;
 
-    this._showYearPanel = false;
-    this._showListPanel = false;
+    this.showCalendarSidebar = false;
+    this.showListPanel = false;
+    this.showCalendarsPanel = false;
```
nsteger, line 153: why no longer protected?

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -229,11 +209,28 @@ export class Calendar extends Widget implements CalendarModel {
     return new KeyStrokeContext();
   }
 
+  protected _createDefaultCalendar(): CalendarDescriptor {
```
nsteger: Method name with "Descriptor"

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -229,11 +209,28 @@ export class Calendar extends Widget implements CalendarModel {
...
+    return {
+      calendarId: UuidPool.ZERO_UUID,
+      selectable: true,
+      visible: true
+    };
+  }
+
+  override init(model: InitModelOf<this>) {
+    super.init(model);
+  }
```
nsteger: should be left at the old position (after constructor)

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -229,11 +209,28 @@ export class Calendar extends Widget implements CalendarModel {
...
     super._init(model);
-    this.yearPanel = scout.create(YearPanel, {
+    this.calendarSidebar = scout.create(CalendarSidebar, {
       parent: this
+
     });
+    this.yearPanel = this.calendarSidebar.yearPanel;
+    this.calendarsPanel = this.calendarSidebar.calendarsPanel;
+
+    this.calendarsPanel.treeBox.on('propertyChange:value', this._onCalendarTreeBoxValueChanged.bind(this));
```
nsteger: naming.. _onCalendarsPanelTreeBoxValueChanged?

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -286,7 +288,9 @@ export class Calendar extends Widget implements CalendarModel {
       }
     }
     this.selectedRange = null;
+    this.selectedCalendar = null;
     this.trigger('selectedRangeChange');
+    this.trigger('selectedCalendarChange', {calendarId: null});
```
nsteger: event naming 'selectedCalendarDescriptorChange' ?

nsteger:
squash commits to 1 commit

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -296,9 +300,9 @@ export class Calendar extends Widget implements CalendarModel {
     }
     this._updateModel(true);
 
+    this._renderComponents();
     // only render if components have another layout
```
nsteger, line 304: comment doesn't match code... wouln't it be enough to render the components when

`!(oldDisplayMode === Calendar.DisplayMode.WEEK && this.displayMode === Calendar.DisplayMode.WORK_WEEK) && !(oldDisplayMode === Calendar.DisplayMode.WORK_WEEK && this.displayMode === Calendar.DisplayMode.WEEK)`

i.e. when switched between workweek and week, no renderComponents is needed...

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -373,6 +377,18 @@ export class Calendar extends Widget implements CalendarModel {
...
+  setShowCalendarSidebar(showCalendarSidebar: boolean) {
+    this.setProperty('showCalendarSidebar', showCalendarSidebar);
+  }
+
+  setShowCalendarsPanel(showCalendarsPanel: boolean) {
+    this.setProperty('showCalendarsPanel', showCalendarsPanel);
+  }
+
+  setShowListPanel(showCalendarsPanel: boolean) {
+    this.setProperty('showListPanel', showCalendarsPanel);
```
nsteger: rename showListPanel

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -481,12 +493,85 @@ export class Calendar extends Widget implements CalendarModel {
...
 
+  setCalendars(calendars: CalendarDescriptor[]) {
+    this.setProperty('calendars', calendars);
+  }
+
+  protected _setCalendars(calendars: CalendarDescriptor[]) {
+    this._setProperty('calendars', calendars);
+    this._updateCalendarsPanelDisplayable();
+    this._updateCalendarsPanel();
+    this._validateSelectedCalendar();
```
nsteger: naming with "calendarDescriptor" instead "calendar" in these two methods..

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -481,12 +493,85 @@ export class Calendar extends Widget implements CalendarModel {
...
+        .data('calendarId', calendar.calendarId);
+    });
+
+    // click event on all day and children elements
+    let mousedownCallbackWithTime = this._onDayColumnMouseDown.bind(this, true);
+    this.$grid.find('.calendar-column').on('mousedown', mousedownCallbackWithTime);
+    let mousedownCallback = this._onDayColumnMouseDown.bind(this, false);
+    this.$topGrid.find('.calendar-column').on('mousedown', mousedownCallback);
+
+    this._renderComponents();
```
nsteger: already called in _renderProperties, shouldn't be needed here

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -481,12 +493,85 @@ export class Calendar extends Widget implements CalendarModel {
...
+
+  protected _renderCalendars() {
+    let $dayName = this.$topGrid.find('.calendar-week-header > .calendar-day-name');
+    let $fullDay = this.$topGrid.find('.calendar-week-allday-container > .calendar-day');
+    let $day = this.$grid.find('.calendar-week > .calendar-day');
+
+    // Remove old calendar columns
+    $dayName.find('.calendar-column').remove();
+    $fullDay.find('.calendar-column').remove();
+    $day.find('.calendar-column').remove();
```
nsteger, line 525: why are these removes needed? Shouldn't be found anything by these "find"...

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -481,12 +493,85 @@ export class Calendar extends Widget implements CalendarModel {
...
+    let mousedownCallbackWithTime = this._onDayColumnMouseDown.bind(this, true);
+    this.$grid.find('.calendar-column').on('mousedown', mousedownCallbackWithTime);
+    let mousedownCallback = this._onDayColumnMouseDown.bind(this, false);
+    this.$topGrid.find('.calendar-column').on('mousedown', mousedownCallback);
+
+    this._renderComponents();
+  }
+
+  protected _updateCalendarsPanel() {
+    this.calendarsPanel.treeBox.lookupCall.setCalendars(this.calendars);
```
nsteger: everyehere calendarDescriptors instead calendars etc, I won't comment every occurence ;)

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -774,6 +877,21 @@ export class Calendar extends Widget implements CalendarModel {
       this.selectedDate = selectedDate;
     }
 
+    // Set selected calendar
+    let newSelectedCalendar = this.selectedCalendar;
+    if (typeof selectedCalendar === 'string' || selectedCalendar instanceof String) {
```
nsteger: why both needed? is one of the checks not enough?

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -845,19 +963,20 @@ export class Calendar extends Widget implements CalendarModel {
...
+    if (this.showCalendarSidebar) {
+      this.calendarSidebar.$container.data('new-width', this.calendarToggleYearWidth);
       gridW -= this.calendarToggleYearWidth;
       containerW -= this.calendarToggleYearWidth;
     } else {
-      this.yearPanel.$container.data('new-width', 0);
+      this.calendarSidebar.$container.data('new-width', 0);
     }
 
+    this.calendarSidebar.setCalendarsPanelExpanded(this.showCalendarsPanel);
```
nsteger, line 994: move this to if above? (if calendarSidebar not visible nothing has to be changed here..)

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -1062,7 +1249,8 @@ export class Calendar extends Widget implements CalendarModel {
         date = from;
       }
 
-      let usedIndices = arrays.ensure(usedIndicesMap.get(date.valueOf()));
+      let key = this._calculateFullDayIndiciesKey(component, date);
```
nsteger: rename _calculateFullDayIndiciesKey => _calculateFullDayIndexKey ?

use better variable name instead "key" (for example fullDayIndexKey)

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -1519,6 +1770,18 @@ export class Calendar extends Widget implements CalendarModel {
     }
   }
 
+  protected _calculateStackKey(date: Date, calendarId?: string): string {
+    let stackKey = this.defaultCalendar.calendarId;
```
nsteger: date is not used here

nsteger: ok, seems that date does not have to be necessarily part of the key, because _arrangeComponentSetPlacement is called directly after _arrange, but why do you need calendarId as part of the stack key then?

key will always be the same except when mode is day. But in day mode, the columns are the different colendarDescriptors, which are handled in separately in a loop execution. As far as I can see, the whole logic with the key can be deleted..

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -1787,6 +2050,10 @@ export class Calendar extends Widget implements CalendarModel {
     return null;
   }
 
+  protected _getCalendarsWithoutGroups(): CalendarDescriptor[] {
```
nsteger: maybe rename to _getLeafCalendarDescriptors?

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -1898,17 +2165,27 @@ export class Calendar extends Widget implements CalendarModel {
...
-    let selectedDateTime = this._getSelectedDateTime(event);
+    let selectedDateTime = this._getSelectedDateTime(event),
+      newSelectedCalendarId = $(event.delegateTarget).data('calendarId');
+
     if (event.which === 3 &&
       this.selectedRange && selectedDateTime &&
       dates.compare(selectedDateTime, this.selectedRange.from) >= 0 &&
-      dates.compare(selectedDateTime, this.selectedRange.to) < 0) {
+      dates.compare(selectedDateTime, this.selectedRange.to) < 0 &&
+      this.selectedCalendar.calendarId === newSelectedCalendarId) {
```
nsteger: this doesn't work as expected, it seems that this.selectedCalendar is already updated! (I assume the following was intended: when in day view and when a range in calendarDescriptor-A is selected and then a right click in the time-range but NOT in the calendarDescriptor-A but for example in the calendarDescriptor-B is done, then the right click sould not be ignored, instead a range in calendarDescriptor-B should be selected)

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -1945,6 +2222,16 @@ export class Calendar extends Widget implements CalendarModel {
...
   }
 
+  protected _findSelectedCalendarColumn(selectedDate: Date): JQuery {
+    let $day = this._findDayInCalendar(selectedDate);
+    if (!$day) {
+      return null;
+    }
+    return $day
+      .find('.calendar-column')
+      .filter((i, e) => $(e).data('calendarId') === this.selectedCalendar.calendarId);
```
nsteger: this.selectedCalendar can be set to null... nullcheck? or set defaultCalendar instead of null for selectedCalendar?

eclipse-scout-core/src/calendar/CalendarComponent.ts
```diff
@@ -124,20 +137,29 @@ export class CalendarComponent extends Widget implements CalendarComponentModel
...
+        .addClass(calendarCssClass)
         .data('component', this)
         .data('partDay', partDay)
+        .on('mouseenter', this._onMouseEnter.bind(this))
+        .on('mouseleave', this._onMouseLeave.bind(this))
         .on('mouseup', this._onMouseUp.bind(this))
         .on('contextmenu', this._onContextMenu.bind(this));
       $part.appendDiv('calendar-component-leftcolorborder');
       let $partContent = $part.appendDiv('content');
       if (this.item.subjectIconId) {
```
nsteger: use item instead of this.item? Or if its never null, delete variable item in this method..

eclipse-scout-core/src/calendar/CalendarModel.ts
```diff
@@ -22,11 +22,25 @@ export interface CalendarModel extends WidgetModel {
...
+  /**
+   * Indicates, if the sidebar is shown
+   */
+  showCalendarSidebar?: boolean;
+  /**
+   * Indicates, if the calendars panel is shown.
+   */
+  showCalendarsPanel?: boolean;
+  /**
+   * Indicates, if the calendars panel is expanded
```
nsteger: the list panel?

eclipse-scout-core/src/calendar/CalendarSidebar.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    this.splitter.on('positionChange', this._onSplitterPositionChange.bind(this));
+    this.splitter.on('splitterClick', this._onSplitterClick.bind(this));
+  }
+
+  protected override _render() {
+    this.$container = this.$parent.appendDiv('calendar-sidebar');
+    this.htmlComp = HtmlComponent.install(this.$container, this.session);
+    this.htmlComp.setLayout(new CalendarSidebarLayout(this));
+    this.htmlComp.invalidateLayoutTree(false);
```
nsteger: needed here?

eclipse-scout-core/src/calendar/CalendarSidebar.ts
```diff
@@ -0,0 +1,91 @@
...
+  protected _onSplitterPositionChange(event: SplitterPositionChangeEvent) {
+    this.invalidateLayoutTree(false);
+  }
+
+  protected _onSplitterClick(event: Event<CalendarSidebarSplitter>) {
+    this.setCalendarsPanelExpanded(this.splitter.collapsed, true);
+  }
+
+  setCalendarsPanelExpanded(expanded: boolean, animate = false) {
+    if (this.splitter.collapsed !== undefined && !this.splitter.collapsed === expanded) {
```
nsteger: rewrite second part to this.splitter.collapsed !== expanded

(for me this is easier to read)

eclipse-scout-core/src/calendar/CalendarSidebarLayout.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    // Set new sizes
+    this.yearPanel.htmlComp.setBounds(new Rectangle(insets.left, insets.top, availableSize.width, yearPanelHeight));
+    this.calendarsPanel.htmlComp.setBounds(new Rectangle(insets.left, insets.top + calendarsPanelMargin, availableSize.width, calendarsPanelHeight));
+
+    this.calendarSidebar.yearPanel._scrollYear();
+  }
+
+  protected _validateSplitterPosition(htmlComp: HtmlComponent, splitter: Splitter, availableSize: Dimension) {
+    // Window has been resized -> preserve relative splitter position
```
nsteger, line 84: this doesnt work yet. if i resize my window, size calculation of sidebar isnt't adjusted correctly

eclipse-scout-core/src/calendar/CalendarSidebarLayout.ts
```diff
@@ -0,0 +1,150 @@
...
+  protected _silentUpdateSpliterPosition(htmlComp: HtmlComponent, splitter: Splitter, newPosition: number) {
+    htmlComp.suppressInvalidate = true;
+    try {
+      splitter.setPosition(newPosition);
+    } finally {
+      htmlComp.suppressInvalidate = false;
+    }
+  }
+
+  protected _calculateColapsedLableHeight(): number {
```
nsteger: typos =>  _calculateCollapsedLabelHeight

eclipse-scout-core/src/calendar/CalendarSidebarLayout.ts
```diff
@@ -0,0 +1,150 @@
...
+    if (!this.splitter.rendered) {
+      return 40; // arbitrary number
+    }
+    let labelHeight = this.splitter.$collapsedLabel.outerHeight();
+    let splitterHeight = this.splitter.$splitterBox.outerHeight();
+    return labelHeight + splitterHeight;
+  }
+
+  setNewSplitterPosition(pos: number, animate?: boolean) {
+    this._newSplitterPosition = pos;
```
nsteger, line 143: maybe _newRelativeSplitterPosition should be set to null in this case?

eclipse-scout-core/src/calendar/CalendarSidebarLayout.ts
```diff
@@ -0,0 +1,150 @@
...
+    // Applies new absolute position
+    if (this._newSplitterPosition !== null) {
+      let newPos = this._newSplitterPosition;
+      this._newSplitterPosition = null;
+      this._setSplitterPosition(newPos, this._animateNewSplitterPosition);
+      this._validateSplitterPosition(htmlComp, this.splitter, availableSize);
+    }
+
+    // Applies new relative position
+    if (this._newRelativeSplitterPosition !== null) {
```
nsteger, line 59: maybe else if?

eclipse-scout-core/src/calendar/CalendarSidebarModel.ts
```diff
@@ -0,0 +1,32 @@
...
+  splitter?: CalendarSidebarSplitter;
+  /**
+   * Calendars panel widget
+   */
+  calendarsPanel?: CalendarsPanel;
+  /**
+   * Defines, whenether the calendar panel is displayable.
+   * The calendar panel is e.g. not visible When only one calendar
+   * is on the calendar widget.
+   */
```
nsteger, line 29: Rewrite:
```
  /**
   * Defines, whether the calendars panel is displayable.
   * For example, the calendars panel is not visible when the calendar widget has only one calendar descriptor.
   */
```
?

eclipse-scout-core/src/calendar/CalendarSidebarSplitter.less
```diff
@@ -0,0 +1,62 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ */
+.splitter.calendar-sidebar-splitter {
+
+  & > .splitter-box {
+    display: flex;
+    min-width: 206px;
+
+    &:not(.disabled) {
+      cursor: pointer;
```
nsteger: why? I think otherwise the cursor is a "resize" cursor, right? I thing this is more intuitive. Maybe pointer cursor for group-collapse-icon but not when hovered over the line..

eclipse-scout-core/src/calendar/CalendarSidebar.ts
```diff
@@ -0,0 +1,91 @@
...
+      return;
+    }
+    // 62 is the golden ratio
+    (this.htmlComp.layout as CalendarSidebarLayout).setNewSplitterPositionPercentage(expanded ? 62 : 100, animate);
```
nsteger, line 78: not sure if this should be set when calendarsPanel needs less space? in this case the splitter position should be higher...?

eclipse-scout-core/src/calendar/CalendarSidebarSplitter.ts
```diff
@@ -0,0 +1,86 @@
...
+  collapsed: boolean;
+  collapsedLabel: string;
+
+  $splitterBox: JQuery;
+  $collapsedLabel: JQuery;
+  $collapseIcon: JQuery;
+  $collapseBorderLeft: JQuery;
+  $collapseBorderRight: JQuery;
+
+  protected _spliterMoveStartPosition: number;
```
nsteger: typo splitter

eclipse-scout-core/src/calendar/CalendarSidebarSplitter.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  setCollapsed(collapsed: boolean) {
+    this.setProperty('collapsed', collapsed);
+  }
+
+  setCollapsedLabel(label: string) {
+    this.setProperty('collapsedLabel', label);
+  }
+
+  protected override _render() {
```
nsteger, line 35: move _render after _init

eclipse-scout-core/src/calendar/CalendarsPanel.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  protected override _createNode(lookupRow: LookupRow<string>): TreeBoxTreeNode<string> {
+    let node = super._createNode(lookupRow);
+    node.expanded = true;
+    return node;
+  }
+
+  protected override _onTreeNodesChecked(event: TreeNodesCheckedEvent) {
+    // Make impossible to uncheck all nodes
+    if (arrays.hasElements(this.tree.checkedNodes)) {
```
nsteger, line 62: not sure about this logic. I would prefer the usual convenience logic: none selected = behave as all selected

eclipse-scout-core/src/calendar/CalendarsPanelTreeNode.ts
```diff
@@ -0,0 +1,24 @@
...
+export class CalendarsPanelTreeNode extends TreeNode implements CalendarsPanelTreeNodeModel {
+  declare model: CalendarsPanelTreeNodeModel;
+
+  calendarId: string;
+  parentId: string;
+
+  constructor() {
+    super();
+
+    this.calendarId = null;
```
nsteger, line 22: ok. why not parentId = null?

eclipse-scout-core/src/index.ts
```diff
@@ -619,6 +620,16 @@ export * from './tree/keystrokes/CompactTreeUpKeyStroke';
 export * from './tree/keystrokes/CompactTreeDownKeyStroke';
 export * from './tree/keystrokes/CompactTreeLeftKeyStroke';
 export * from './tree/keystrokes/CompactTreeRightKeyStroke';
+export * from './calendar/CalendarsPanelLookupCall';
```
nsteger: order alphabetically

eclipse-scout-core/src/index.ts
```diff
@@ -940,6 +951,7 @@ export * from './tabbox/SimpleTabArea';
 export * from './tabbox/SimpleTabAreaModel';
 export * from './tabbox/SimpleTabAreaEventMap';
 export * from './tabbox/SimpleTabAreaLayout';
+export * from './calendar/CalendarsPanel';
```
nsteger: this too, order alphabetically

eclipse-scout-core/src/tree/TreeNodeUpdate.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * This object is needed during an update in the tree. It
+ * holds the updated nodes to be processed at the end of the update.
+ *
+ * This class holds two lists of {@link TreeNode}:
+ * - TreeNodes which have to be rendered
+ * - TreeNodes which triggered an update event
+ */
+export class TreeNodeUpdate {
```
nsteger: is this class still needed? I can't find any usage

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/calendar/JsonCalendar.java
```diff
@@ -231,6 +267,31 @@ public <C extends CalendarComponent> JsonCalendarComponent<C> resolveCalendarCom
...
+    if (ICalendar.PROP_SHOW_CALENDAR_SIDEBAR.equals(propertyName)) {
+      getModel().getUIFacade().setShowYearPanelFromUI(data.getBoolean(propertyName));
+    }
+    else if (ICalendar.PROP_SHOW_CALENDARS_PANEL.equals(propertyName)) {
+      getModel().getUIFacade().setShowCalendarsPanelFromUI(data.getBoolean(propertyName));
+    }
+    else if (ICalendar.PROP_SHOW_LIST_PANEL.equals(propertyName)) {
+      getModel().getUIFacade().setShowListPanelFromUI(data.getBoolean(propertyName));
+    }
+    super.handleUiPropertyChange(propertyName, data);
```
nsteger: else { ... }

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/tree/AbstractTree.java
```diff
@@ -358,6 +358,26 @@ protected boolean getConfiguredAutoCheckChildNodes() {
...
+   * <li>NONE: No nodes are auto-checked</li>
+   * <li>AUTO_CHECK_CHILD_NODES: All child nodes will be checked/unchecked together with their parent</li>
+   * <li>SYNCH_CHILD_AND_PARENT_STATE: The state of the node is a representation of its children</li>
+   * </ul>
+   * <p>
+   * Only has an effect if the tree is checkable.
+   *
+   * @see AutoCheckStyle
+   * @see #getConfiguredCheckable()
+   * @since 5.1
```
nsteger, line 373: 5.1?

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/calendar/provider/ICalendarItemProvider.java
```diff
@@ -35,6 +36,10 @@ public interface ICalendarItemProvider extends IPropertyObserver {
    * type boolean
    */
   String PROP_LOAD_IN_PROGRESS = "loadInProgress";
+  /**
+   * type ICalendarDescriptor
+   */
+  String PROP_CALENDAR_BELONGING = "calendarBelonging";
```
nsteger: use CALENDAR_DESCRIPTOR?

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/tree/AbstractTree.java
```diff
@@ -358,6 +358,26 @@ protected boolean getConfiguredAutoCheckChildNodes() {
     return false;
   }
 
+  /**
+   * Configures the default auto-check mode of the tree. There are three modes:
+   * <ul>
+   * <li>NONE: No nodes are auto-checked</li>
+   * <li>AUTO_CHECK_CHILD_NODES: All child nodes will be checked/unchecked together with their parent</li>
+   * <li>SYNCH_CHILD_AND_PARENT_STATE: The state of the node is a representation of its children</li>
```
nsteger, line 366: how is the logic in combination with getConfiguredAutoCheckChildNodes? javadoc has to be clearer here. Shouls getConfiguredAutoCheckChildNodes be deprecated and the autocheckstyle AUTO_CHECK_CHILD_NODES should be used instead?

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/tree/AutoCheckStyle.java
```diff
@@ -0,0 +1,33 @@
...
+  /**
+   * The state of the node is a representation of its children.
+   * <ul>
+   * <li>When none of the children are checked, the node is unchecked</li>
+   * <li>When some of the children are checked, the node is partly checked</li>
+   * <li>When all of the children are checked, the node is also checked</li>
+   * </ul>
+   * When a node with children is selected, its children will be auto-checked
+   */
+  SYNCH_CHILD_AND_PARENT_STATE
```
nsteger, line 32: typo: SYNC

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/tree/ITree.java
```diff
@@ -198,6 +199,10 @@ public interface ITree extends IWidget, IDNDSupport, IStyleable, IAppLinkCapable
 
   void setAutoCheckChildNodes(boolean autoCheckChildNodes);
 
+  AutoCheckStyle getAutoCheckStyle();
```
nsteger, line 202: i didnt't find any logic for this property? can it be deleted?

eclipse-scout-core/test/calendar/CalendarSpec.ts
```diff
@@ -41,6 +66,115 @@ describe('Calendar', () => {
       expect(cal.viewRange).toBeDefined();
     });
 
+    describe('expands panels corretly', () => {
+
+      const getContainterWidth = (container: JQuery): number => {
```
nsteger: typo => container (also getContainerHeight)

eclipse-scout-core/test/calendar/CalendarSpec.ts
```diff
@@ -41,6 +66,115 @@ describe('Calendar', () => {
       expect(cal.viewRange).toBeDefined();
     });
 
+    describe('expands panels corretly', () => {
```
nsteger: describe should contain a function or a property, it isn't a test itself. For example "calendar panels expand"

eclipse-scout-core/test/calendar/CalendarSpec.ts
```diff
@@ -41,6 +66,115 @@ describe('Calendar', () => {
...
+        return container.width();
+      };
+
+      const getContainterHeight = (container: JQuery): number => {
+        return container.height();
+      };
+
+      it('should expand the calendar sidebar when showCalendarSidebar is true', () => {
+        // Arrange
+        let model = {parent: session.desktop, showCalendarSidebar: true};
```
nsteger, line 81: inline model

nsteger: also in all the other cases :)

eclipse-scout-core/test/calendar/CalendarSpec.ts
```diff
@@ -41,6 +66,115 @@ describe('Calendar', () => {
...
+          parent: session.desktop,
+          showCalendarSidebar: true,
+          showCalendarsPanel: true,
+          calendars: [{calendarId: 'a'}, {calendarId: 'b'}]
+        };
+        let cal = scout.create(Calendar, model);
+
+        // Act
+        cal.render();
+        cal.calendarSidebar.revalidateLayout();
```
nsteger, line 119: why needed here?

eclipse-scout-core/test/calendar/CalendarSpec.ts
```diff
@@ -488,6 +627,384 @@ describe('Calendar', () => {
...
+
+      // Act
+      calendar._updateCalendarVisibility([[otherCalendar.calendarId, false]]);
+
+      // Assert
+      expect(businessComp.visible).toBe(true);
+      expect(externalComp.visible).toBe(false);
+    });
+
+    // No range selection on disabled column
```
nsteger: comment not needed here

eclipse-scout-core/test/calendar/CalendarSpec.ts
```diff
@@ -488,6 +627,384 @@ describe('Calendar', () => {
...
+
+        // Act
+        calendar._setSelection(new Date(stringDay), selectableCalendar.calendarId, null, false, false);
+
+        // Assert
+        expect(calendar.selectedCalendar).toBe(selectableCalendar);
+      });
+    });
+
+    describe('one calendar should always be selected', () => {
```
nsteger: better someting like "calendars tree box"

nsteger (changes requested):

eclipse-scout-core/src/calendar/Calendar.ts
Knicksaas: They were pulled up to the model, to synch them to java. Only "public" members should be in the model

eclipse-scout-core/src/calendar/Calendar.ts
Knicksaas: Is this override even necessary? I would not know a reason?

eclipse-scout-core/src/calendar/Calendar.ts
Knicksaas: Rename to _onCalendarVisibilityChanged

eclipse-scout-core/src/calendar/Calendar.ts
Knicksaas: This is corret. I simplyfied the statement a little: `arrays.containsAny([this.displayMode, oldDisplayMode], [Calendar.DisplayMode.DAY, Calendar.DisplayMode.MONTH])` It should have the same effect

eclipse-scout-core/src/calendar/Calendar.ts
Knicksaas: The scout naming convention defines the name of the setter as 'set' + propertyName. I would not break this concept here. Other places do also use this convention: see Form#setShowOnOpen (Form.ts).

eclipse-scout-core/src/calendar/Calendar.ts
Knicksaas: It is necessary, because when the setCalendarDescriptors is called, the calendars are re-rendered. So do have the components be re-rendered. It is not nescessary in the initial render process. So I wrapped the render statement in an if-clause, to check if `this.rendering = true`.

eclipse-scout-core/src/calendar/Calendar.ts
Knicksaas: I believe this is necessary. The _render Method is called, whenever the property changes. The default implementation does not know how to remove the calendars, so here it has to be done. Alternatively, the _removeCalendarDescriptor Method could be overridden, to extract this logic to a separate Method. 

What do you mean with 'Shouldn't be found anything by these "find"...'?

eclipse-scout-core/src/calendar/Calendar.ts
Knicksaas: Sadly, no... See https://stackoverflow.com/questions/4059147/check-if-a-variable-is-a-string-in-javascript...

eclipse-scout-core/src/calendar/Calendar.ts
Knicksaas: I think, it makes sense to be able to modify the property, event when the sidebar is not expanded. Because maybe you want, that the Tree is instant visible when you expand the sidebar. Otherwise, you would have to click twice...

eclipse-scout-core/src/calendar/Calendar.ts
Knicksaas: i think, i produce a bug with this stack key calculation... have to check next week

eclipse-scout-core/src/calendar/Calendar.ts
nsteger: I mean rename variable showCalendarsPanel to showListPanel ;)

eclipse-scout-core/src/calendar/Calendar.ts
nsteger: Me neither.. You can also delete it as far as I know

eclipse-scout-core/src/calendar/Calendar.ts
nsteger: Ah, I forgot the reflection-like call of the render functions.. Code is ok here!

eclipse-scout-core/src/calendar/Calendar.ts
nsteger: I agree => resolved

eclipse-scout-core/src/calendar/Calendar.ts
nsteger: since you listen to 'propertyChange:value' event you should follow the naming pattern _onXyzValueChange, where Xyz should be the object to which the listener is added

eclipse-scout-core/src/calendar/Calendar.ts
Knicksaas: this.selectedCalendar(Descriptor) should now never be null. If no calendar is selected, the value should be equal with this.defaultCalendar. For stability reasons, I included a null check.

eclipse-scout-core/src/calendar/CalendarSidebar.ts
Knicksaas: No :), I removed it

eclipse-scout-core/src/calendar/CalendarsPanelTreeNode.ts
Knicksaas: I forgot :)

eclipse-scout-core/src/index.ts
Knicksaas: JavaScript does interpret the imports from top to bottom. because I need the TreeBox in the CalendarsPanel, i have to place the import below the one of the TreeBox

eclipse-scout-core/src/tree/TreeNodeUpdate.ts
Knicksaas: this was used for the change with the auto check child nodes in the Tree widget. But it is not used now.

eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -1468,29 +1469,28 @@ export class Calendar extends Widget implements CalendarModel {
...
+    return dates.compare(selectedDate, component.coveredDaysRange.from) >= 0 &&
+      dates.compare(selectedDate, component.coveredDaysRange.to) <= 0;
+  }
 
-    components.forEach(component => {
-      listComponent = new CalendarListComponent(this.selectedDate, component);
-      listComponent.render(this.$list);
-      this._listComponents.push(listComponent);
-    });
+  protected _filterVisibleComponents(component: CalendarComponent) {
```
nsteger, line 1515: return type :boolean

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/calendar/provider/ICalendarItemProvider.java
```diff
@@ -55,9 +55,9 @@ public interface ICalendarItemProvider extends IPropertyObserver {
 
   void setRefreshIntervalMillis(long millis);
 
-  ICalendarDescriptor getCalendarBelonging();
+  ICalendarDescriptor getAssociatedCalendarDescriptor();
 
-  void setCalendarBelonging(ICalendarDescriptor belongingCalendar);
+  void setAssociatedCalendarDescriptor(ICalendarDescriptor belongingCalendar);
```
nsteger, line 60: rename belongingCalendar => associatedCalendarDescriptor

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/calendar/ICalendar.java
```diff
@@ -112,11 +112,11 @@ public interface ICalendar extends IWidget, IContextMenuOwner {
 
   void setDisplayCondensed(boolean condensed);
 
-  List<ICalendarDescriptor> getCalendars();
+  List<ICalendarDescriptor> getCalendarDescriptors();
 
-  void setCalendars(List<ICalendarDescriptor> calendars);
+  void setCalendarDescriptors(List<ICalendarDescriptor> calendars);
```
nsteger, line 117: rename calendars => calendarDescriptors

Knicksaas:
replaced by #1051

PR #863, opened 2024-01-22
Environment map properties may lead to deadlock on platform start

matthiaso:
370937


matthiaso:
Cherry-pick of reviewed commit, see https://github.com/eclipse-scout/scout.rt/pull/857

PR #866, opened 2024-01-23
HTTP-Proxy: Prevent doubled encoding of query parameters

buehlmic:
Due to a recent change, query parameters were encoded twice inside HttpProxy#rewriteUrl. In this change we undo this doubled encoding.

372285

### DIFF
--- a/org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
+++ b/org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
@@ -252,7 +252,8 @@ public class HttpProxy {
       pathInfo = pathInfo.substring(1);
     }
     uriBuilder.addPath(pathInfo);
-    uriBuilder.queryString(req.getQueryString());
+    // Decoding URL since req.getQueryString method doesn't decode it. The query string gets encoded again when creating the URI.
+    uriBuilder.queryString(IOUtility.urlDecode(req.getQueryString()));
     return uriBuilder.createURI();
   }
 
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
```diff
@@ -252,7 +252,8 @@ protected URI rewriteUrl(HttpServletRequest req, HttpProxyRequestOptions options
       pathInfo = pathInfo.substring(1);
     }
     uriBuilder.addPath(pathInfo);
-    uriBuilder.queryString(req.getQueryString());
+    // Decoding URL since req.getQueryString method doesn't decode it. The query string gets encoded again when creating the URI.
+    uriBuilder.queryString(IOUtility.urlDecode(req.getQueryString()));
```
matthiaso, line 256: The `urlDecode` method may fail if we would get an improper encoded query string; however I as this is already an incoming HTTP request we should probably assume that our input is properly encoded.

`org.eclipse.scout.rt.server.commons.servlet.HttpProxyTest.testRewriteUrl()` is going to fail here (last case - however test needs adjustment here as it is an incorrect case); suggest to remove:

`    testRewriteUrlInternal("http://internal.example.com:1234/api/lorem?foo=%25bar%25",
        "/lorem", "foo=%bar%", new HttpProxyRequestOptions());`

and add

`    testRewriteUrlInternal("http://internal.example.com:1234/api/lorem?foo=%25bar%25+lorem+ipsum",
        "/lorem", "foo=%25bar%25+lorem+ipsum", new HttpProxyRequestOptions());`

as last case?

matthiaso (changes requested):

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
buehlmic: Done as suggested

matthiaso (approved):

PR #869, opened 2024-01-24
Allow to override _lookupRowsByAll

awegmueller:
Some subclasses need to implement _lookupRowsByAll. In that case super (StaticLookupCall) should not limit the result by maxRowCount. The new methods allow to re-use the methods of the base class in the subclass.

363643

### DIFF
--- a/eclipse-scout-core/src/lookup/StaticLookupCall.ts
+++ b/eclipse-scout-core/src/lookup/StaticLookupCall.ts
@@ -66,8 +66,23 @@ export class StaticLookupCall<TKey> extends LookupCall<TKey> implements StaticLo
     });
   }
 
+  /**
+   * Limits the data to 'maxRowCount' results, converts the data to LookupRows and filters
+   * the rows by their 'active' state. If a subclass overrides this method it can omit the
+   * super call, if it must avoid the maxRowCount limit and does its own filtering. In that
+   * case the subclass may choose to call <code>#_mapAndFilterData</code> and
+   * <code>#_limitMaxRows</code> when suitable.
+   */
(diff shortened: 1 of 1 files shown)


eclipse-scout-core/src/lookup/StaticLookupCall.ts
```diff
@@ -66,8 +66,23 @@ export class StaticLookupCall<TKey> extends LookupCall<TKey> implements StaticLo
...
+   * Limits the data to 'maxRowCount' results, converts the data to LookupRows and filters
+   * the rows by their 'active' state. If a subclass overrides this method it can omit the
+   * super call, if it must avoid the maxRowCount limit and does its own filtering. In that
+   * case the subclass may choose to call <code>#_mapAndFilterData</code> and
+   * <code>#_limitMaxRows</code> when suitable.
+   */
   protected _lookupRowsByAll(): LookupRow<TKey>[] {
-    let datas = this.data.slice(0, this.maxRowCount);
+    let datas = this._limitMaxRows(this.data);
+    return this._mapAndFilterData(datas);
```
buehlmic, line 78: Is there a reason why we first limit the results and only then apply our filter? What about doing it the other way around?

awegmueller: I didn't want to change the order, because I don't want to introduce unwanted side effects with my change (like failing unit tests that expect a certain number of results).  But I agree, it would make more sense the other way around. However, in that case you'd also have to change the _filterActiveLookupRow method which works on LookupRow now, but needed to work on the raw data array if you'd do it before the map() call. I'd rather do not change anything here, because the StaticLookupCall is a fairly wide used class.

PR #872, opened 2024-01-25
UiNotificationRegistry: fix ConcurrentModificationException

cguglielmo:
computeIfAbsent may throw an error if accessed concurrently.

ConcurrentHashMap would solve that, but removeListener is not thread safe as well. It could remove a listener list for a topic if addListener was called again right after isEmpty check. ConcurrentHashMap would not help here, so we need to synchronize anyway.

372642


org.eclipse.scout.rt.uinotification/src/main/java/org/eclipse/scout/rt/api/uinotification/UiNotificationRegistry.java
```diff
@@ -352,13 +352,13 @@ protected void updateNotificationCreationTime(UiNotificationDo notification) {
     m_lastCreationTime = notification.getCreationTime();
   }
 
-  public void addListener(String topic, UiNotificationListener listener) {
+  public synchronized void addListener(String topic, UiNotificationListener listener) {
```
fschinkel: Why synchronized over the instance of `UiNotificationRegistry`? Wouldn't it be more suitable to synchronize only via `m_listerners`?

fschinkel (changes requested):

PR #874, opened 2024-01-26
UiSession: warn when model theme is not valid

bschwarzent:
Automatically assigning a theme when the model theme is not set is a valid case (user has no specific setting). Therefore, this is only logged on level DEBUG. However, if a theme is set explicitly but is not valid, a message with level WARNING is logged.

270127


bschwarzent:
Abandoned in favor of #1001

PR #876, opened 2024-01-29
ScoutJS Archetype: fix permissions url

fschinkel:


scout-hellojs-app/src/main/resources/archetype-resources/__rootArtifactId__.app/src/main/js/__simpleArtifactName__.ts
```diff
@@ -1,9 +1,10 @@
+import {App as ScoutApp} from '@eclipse-scout/core';
 import {App} from '@${simpleArtifactName}/ui';
 
 new App().init({
   bootstrap: {
     textsUrl: 'texts.json',
-    permissionsUrl: 'api/permissions'
+    permissionsUrl: (ScoutApp.get() as App).apiUrl + 'permissions'
```
cguglielmo: I would use App.get() as App instead of ScoutApp.get to make the additional import obsolete

PR #877, opened 2024-01-30
HttpProxy: Umlauts must be encoded as well for URLs

matthiaso:
UriBuilder does not seem to encode all characters which should be encoded (which was the hope in b9687800539ccda586625b6a7e542a2c4561b8cd) -> Use different url encoding function.

372949

### DIFF
--- a/org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
+++ b/org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
@@ -14,7 +14,6 @@ import static org.eclipse.scout.rt.platform.util.Assertions.assertTrue;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.net.URI;
 import java.net.URLEncoder;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
```diff
@@ -240,21 +238,20 @@ protected HttpContext createHttpContext() {
...
     if (rewriteRule != null) {
       pathInfo = rewriteRule.rewrite(pathInfo);
     }
-    UriBuilder uriBuilder = new UriBuilder(getRemoteBaseUrl());
-    if (StringUtility.startsWith(pathInfo, "/")) {
-      // addPath on UriBuilder adds a beginning slash on its own; strip our beginning slash to avoid duplicated slashes
+    // pathInfo must be url-encoded; except for forward-slashes they should be kept
+    pathInfo = IOUtility.urlEncode(pathInfo).replaceAll("%2F", "/");
+    String remoteBaseUrl = getRemoteBaseUrl();
+    if (remoteBaseUrl.endsWith("/") && pathInfo.startsWith("/")) {
```
stephan-merkli: As discussed, check both conditions separately and remove the / then.
String url = remoteBaseUrl + "/" + pathInfo;
Use StringUtility.join("?", url, req.getQueryString())

matthiaso: * Removed trailing slash check for remoteBaseUrl completely, according to javadoc on getRemoteBaseUrl it should never include a trailing slash (added however a trailing slash check in withRemoteBaseUrl).
* An URL with an empty pathInfo (and w/o) slash may be valid after all, e.g. set remoteBaseUrl for a specific ressource remoteBaseUrl = http://my-server/my-servlet - in this case one may want to create a proxy accessing this specific resource w/o any slashes.

stephan-merkli (approved):

PR #882, opened 2024-02-02
Expose CodeTypes to Scout JS

mvilliger:


eclipse-scout-core/src/code/CodeType.ts
```diff
@@ -9,17 +9,17 @@
  */
 import {Code, codes, CodeTypeModel, FullModelOf, InitModelOf, Locale, ObjectOrModel, ObjectWithType, scout, texts, TreeVisitor, TreeVisitResult} from '../index';
 
-export class CodeType<TCodeId> implements ObjectWithType {
+export class CodeType<TCodeId = string, TCodeClass extends Code<TCodeId> = Code<TCodeId>, TCodeTypeId = string> implements ObjectWithType {
```
cguglielmo: Instead of TCodeClass just use TCode, No need to overspecify what it is

PR #889, opened 2024-02-07
GroupBoxLayout: ensure notification layout is valid

bschwarzent:
A group box can have a notification which takes up vertical space. While the height of the notification was already subtracted from the available size, the layout of the notification widget was never actually (re-)validated. Because of this, the notification's HtmlComponent always returned the same cached height, even when the group box was resized. This resulted in a wrong height for the group box body.

Example: when a form with a notification with a long text is made smaller, the notification becomes taller (because the browser automatically wraps the text), but the height of the group box body does not change. Consequently, the widgets inside the box "fall out" of the bottom.

367904

### DIFF
--- a/eclipse-scout-core/src/form/fields/groupbox/GroupBoxLayout.js
+++ b/eclipse-scout-core/src/form/fields/groupbox/GroupBoxLayout.js
@@ -27,7 +27,7 @@ export default class GroupBoxLayout extends AbstractLayout {
   }
 
   _initDefaults() {
-    this._statusWidth = HtmlEnvironment.get().fieldStatusWidth;
+    this._fieldStatusWidth = HtmlEnvironment.get().fieldStatusWidth;
   }
 
   _onHtmlEnvironmentPropertyChange() {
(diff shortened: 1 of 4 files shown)


fschinkel (changes requested):
Please also consider the case, that the notification is part of a group box with a visible status. The widthHint used during preferredLayoutSize() contains the width for the group box, but the notification will be smaller if there is a status. If one calculates the notification height using the width hint for the group box without substracting the status width there are cases where the result is incorrect (text wrapped if status is visible, but not wrapped if status is hidden).

eclipse-scout-core/src/form/fields/groupbox/GroupBoxLayout.js
```diff
@@ -27,7 +27,7 @@ export default class GroupBoxLayout extends AbstractLayout {
   }
 
   _initDefaults() {
-    this._statusWidth = HtmlEnvironment.get().fieldStatusWidth;
+    this.fieldStatusWidth = HtmlEnvironment.get().fieldStatusWidth;
```
fschinkel: This is an internal property, please use `_`-prefix.

fschinkel (changes requested):

eclipse-scout-core/src/form/fields/groupbox/GroupBoxLayout.js
bschwarzent: Changed to _fieldStatusWidth

fschinkel (approved):

PR #890, opened 2024-02-07
Throw error when Karma config is invalid

bschwarzent:
When Karma encounters an error during parsing of the configuration, it simply ends the process. Because logs are off by default, it is therefore not easily possible to find the cause of the problem. By passing the argument "throwErrors: true" explicitly, we can instruct Karma to throw an error instead.

351426


mvilliger (approved):
Looks good, thank you.

Please also cherry-pick this change to the branches 23.2, 24.1 & 24.2

PR #892, opened 2024-02-08
BrowserField: fix CSP report-uri

bschwarzent:
If a browser field is used to display a document, the report-uri cannot be relative, because the dynamic URL has a different base than the application. Instead, use the browserUri from the client session to create an absolute URL.

290376


org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/form/fields/browserfield/BrowserFieldContentHttpResponseInterceptor.java
```diff
@@ -52,4 +59,28 @@ public void intercept(HttpServletRequest req, HttpServletResponse resp) {
...
+   * application.
+   * <p>
+   * Examples:
+   * <ul>
+   * <li>http://localhost:8080 → http://localhost:8080/
+   * <li>https://www.example.org/public/my-app/index.html?debug=true → https://www.example.org/public/my-app/
+   * </ul>
+   */
+  @SuppressWarnings("JavadocLinkAsPlainText")
+  protected String toBaseUri(URI browserUri) {
```
cguglielmo: Maybe move to ClientSession or a helper to make it reusable? Maybe replace last regex with lastIndexOf("/") and substr to make it a little simpler? see OfficeAddinUtil.

cguglielmo:
Run tests

PR #893, opened 2024-02-09
Action: show tooltip even if action is disabled

cguglielmo:
Currently, it is not possible to show a tooltip explaining why an action is disabled because tooltips are only shown if the action is enabled. This is different to buttons or form fields where the tooltip is always shown.
This is a limitation and inconsistent, so now the tooltip is always shown.

It could be helpful for existing actions with tooltips as well, to let the user know what a disabled action would do, if it was enabled.

If it turns out that this is not the expected behavior, we could add a property similar to `showTooltipWhenSelected` called
 `showTooltipWhenDisabled`.

362354

### DIFF
--- a/eclipse-scout-core/src/action/Action.ts
+++ b/eclipse-scout-core/src/action/Action.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -290,11 +290,10 @@ export class Action extends Widget implements ActionModel {
   }
 
   protected _shouldInstallTooltip(): boolean {
-    let show = this.tooltipText && this.enabledComputed;
-    if (!this.showTooltipWhenSelected && this.selected) {
-      show = false;
+    if (this.selected && !this.showTooltipWhenSelected) {
+      return false;
     }
-    return show;
+    return !!this.tooltipText;
   }
 
   /** @see ActionModel.tabbable */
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/src/action/Action.ts
```diff
@@ -290,7 +290,7 @@ export class Action extends Widget implements ActionModel {
   }
 
   protected _shouldInstallTooltip(): boolean {
-    let show = this.tooltipText && this.enabledComputed;
+    let show = !!this.tooltipText;
     if (!this.showTooltipWhenSelected && this.selected) {
```
bschwarzent: Method could be simplified to:

```
if (this.selected && !this.showTooltipWhenSelected) {
   return false;
}
return !!this.tooltipText;
```

eclipse-scout-core/test/action/ActionSpec.ts
```diff
@@ -140,6 +140,35 @@ describe('Action', () => {
...
+    it('is shown on mouseover after a short delay ', async () => {
+      let action = scout.create(Action, {
+        parent: session.desktop,
+        tooltipText: 'tooltip'
+      });
+      action.render();
+      JQueryTesting.triggerMouseEnter(action.$container);
+      expect(action.findChild(Tooltip)).toBeFalsy();
+
+      await sleep(tooltips.DEFAULT_TOOLTIP_DELAY);
```
bschwarzent: If I understand it correctly, this sleep() method sleeps in real-time. Therefore these two test cases delay the execution of all jasmine tests by over a second.

Could we use the jasmine clock here to speed this up? Otherwise consider shortening the delay for this "describe", e.g. by changing the value of the DEFAULT_TOOLTIP_DELAY constant or overwriting Action#_configureTooltip.

PR #894, opened 2024-02-09
UiSession: send previous response for already processed poll request

bschwarzent:
When a ?poll request was already processed, the UiSession returns early from waitForBackgroundJobs() and the corresponding response from the history is returned by processJsonRequest(). However, this method cannot be used when the session lock could not be obtained (e.g. because a user request is currently being processed).

Previously, an empty message was returned instead, because it was assumed that the other request will return all necessary events in its own response later. However, if the network is unreliable, the same ?poll request might have already been answered previously. In that case, the previous message would never reach the client.

To prevent an inconsistent response queue, the previous message from the history has to be sent for poll requests as well (instead of an empty message).

363418


eclipse-scout-core/src/session/ResponseQueue.ts
```diff
@@ -27,9 +27,14 @@ export class ResponseQueue {
   }
 
   /**
-   * in milliseconds
+   * Timeout in milliseconds after which the response queue is processed even when a sequence number is still missing.
+   * This should prevent an infinitely blocked UI, but (optimistically) assumes that the missed response was not important.
+   *
+   * This value should be larger than the difference between the polling interval (`scout.ui.backgroundPollingMaxWaitTime`)
+   * and the polling timeout ({@link Session#requestTimeoutPoll}) to give the retry mechanism a change to retrieve a
```
cguglielmo: a change -> a chance? also remove space before last period.

eclipse-scout-core/src/session/ResponseQueue.ts
```diff
@@ -27,9 +27,14 @@ export class ResponseQueue {
...
+   * missed response after a network interruption .
    */
-  static FORCE_TIMEOUT = 10 * 1000;
+  static FORCE_TIMEOUT = 20 * 1000;
```
cguglielmo: Maybe we should extract the 15 of Session:431 into a constant and use it here + 5 ?

PR #896, opened 2024-02-12
Planner: fix cannot read properties of null (reading 'elementFromPoint')

cguglielmo:
Can be reproduced as follows:
1. Press mouse down on a resource
2. Press alt-tab and release mouse button
3. Press mouse down on a resource again
4. Remove planner and move mouse

Problem is that the mouse move handler is not released after alt-tab. Solution is to always remove the move handler before it is added again so that Planner.remove() will remove the latest handler.

Other adjusmtents (not relevant for the bugfix).
- select(true) -> select() because true is the default
- resource title mouse move and up handlers are only necessary in multi select mode

347428

### DIFF
--- a/eclipse-scout-core/src/planner/Planner.ts
+++ b/eclipse-scout-core/src/planner/Planner.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


eclipse-scout-core/src/planner/Planner.ts
```diff
@@ -344,22 +344,25 @@ export class Planner extends Widget implements PlannerModel {
...
+    this._select();
 
     // add event handlers
-    this._resourceTitleMousemoveHandler = this._onResourceTitleMousemove.bind(this, event);
-    this._$body.document()
-      .on('mousemove', this._resourceTitleMousemoveHandler)
-      .one('mouseup', this._onDocumentMouseUp.bind(this));
+    this._removeMouseMoveHandlers();
+    if (this.multiSelect) {
+      this._resourceTitleMousemoveHandler = this._onResourceTitleMousemove.bind(this);
```
bschwarzent, line 352: Wouldn't it be easier if these event handlers were initialized once in the constructor? This would also simplify _removeMouseMoveHandlers().

cguglielmo: Yes, but _onCellMousemove is bound to the mouse down event so it needs to be bound on every mouse down. To keep it consistent I decided to use the same pattern for the other handlers as well.

PR #898, opened 2024-02-14
StringField: reposition 'disabled copy overlay' when scrollbars change

bschwarzent:
When the $disabledCopyOverlay is visible and the input field is a <textarea>, the form field layout positions the overlay such that it does not cover the scrollbars (otherwise they would not be usable). When the scrollbars change visibility when the text content changes, the position of the overlay has to be adjusted dynamically.

368547


bschwarzent:
Run tests

PR #900, opened 2024-02-15
Features/cgu/24.1/table header tooltip

cguglielmo:


eclipse-scout-core/test/layout/RectangleSpec.ts
```diff
@@ -53,4 +53,37 @@ describe('Rectangle', () => {
...
+      expect(rect.intersect(new Rectangle(4, 5, 20, 20))).toEqual(new Rectangle(4, 5, 8, 6));
+
+      // Second rectangle starts further left and top and ends inside first
+      expect(rect.intersect(new Rectangle(0, 1, 3, 4))).toEqual(new Rectangle(2, 3, 1, 2));
+
+      // Second rectangle starts further left and top and touches right border of first
+      expect(rect.intersect(new Rectangle(-4, -5, 17, 10))).toEqual(new Rectangle(2, 3, 10, 2));
+    });
+
+    it('returns an empty rectangle if the rectangles don\'t intersect', () => {
```
bschwarzent, line 80: Add edge cases:
* rect.intersect(null)
* new Rectangle().intersect(new Rectangle())

eclipse-scout-core/src/table/TableHeader.ts
```diff
@@ -310,7 +311,11 @@ export class TableHeader extends Widget implements TableHeaderModel {
       return column.headerTooltipText;
     }
     let $text = $col.children('.table-header-item-text');
-    if ($text.isContentTruncated() || ($col.width() + $col.position().left) > $col.parent().width()) {
+    let textBounds = graphics.offsetBounds($text);
+    let containerBounds = this._offsetBoundsWithoutMenuBar();
+    let textInView = textBounds.right() < containerBounds.right() && textBounds.x >= containerBounds.x;
```
bschwarzent: `textBounds.right() < containerBounds.right()` --> shouldn't this use the `<=` operator as well?

eclipse-scout-core/src/table/TableHeader.ts
```diff
@@ -310,7 +311,11 @@ export class TableHeader extends Widget implements TableHeaderModel {
...
+    if ($text.isContentTruncated() || !textInView) {
```
bschwarzent: The 3 lines above are always executed, even when the first "if" clause returns already true. Either extract them into a _isTextInView() method, swap the if clauses (we can then save the isContentTruncated() call if the text is not in view), or combine them into a "showTooltip" variable.

PR #901, opened 2024-02-15
UiSession: fix threading issues

fschinkel:
UiSession: fix NPE while transforming response

The scope that marks the UiSession with the flag processingJsonRequest
ends before the response is transformed. As this was used to determine
if the UiSession is disposable or not it was possible to dispose the
UiSession after the request was process but before its response was
transformed. Disposing the UiSession will remove the current response
and therefore transforming afterward will fail with NPEs.
The solution is to not use the processingJsonRequest-flag to determine
if the UiSession is disposable or not but to introduce a new
disposable-flag. This flag is set to false right before processing the
request and to true again after transforming the response. If the
UiSession was meant to be disposed in the meantime, it will be disposed
right after the flag is set to true again.

354503

------------------------------------------------------------------------

UiSession: fix dispose while json adapters are not initialized

The UiSession may be disposed at any time because e.g. the browser was
closed or the user navigated to "/logout". This can happen while the
model thread alters the JSON layer. Most of these alternations are
synced using UiSession.uiSessionLock() as they are triggered by user
interaction. But it is possible that the model changes without user
interaction e.g. as a result of a ClientNotification. If this happens
the state of the JSON layer is unreliable, e.g. there might be json
adapters that are created and registered but not yet initialized. If
these json adapters are disposed some might lead to errors, e.g. the
JsonFormField will result in an NPE as its context menu is not yet
created.
The solution is to dispose the JSON layer in a ModelJob. This ensures
that all changes to the model and therefore the JSON layer, which are
not synchronized by UiSession.uiSessionLock(), are finished before
disposing it.
This will not change the behaviour if there are model changes pending,
which are triggered by user interaction, as processJsonRequest already
waits for all model jobs to finish before processing a potential
dispose.

366500

### DIFF
--- a/org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiSession.java
+++ b/org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiSession.java
@@ -9,6 +9,8 @@
  */
 package org.eclipse.scout.rt.ui.html;
 
+import static org.eclipse.scout.rt.platform.util.Assertions.assertNotNull;
+
 import java.security.AccessController;
 import java.util.Collections;
 import java.util.List;
@@ -62,6 +64,7 @@ import org.eclipse.scout.rt.platform.util.LazyValue;
 import org.eclipse.scout.rt.platform.util.ObjectUtility;
 import org.eclipse.scout.rt.platform.util.concurrent.FutureCancelledError;
 import org.eclipse.scout.rt.platform.util.concurrent.ThreadInterruptedError;
+import org.eclipse.scout.rt.platform.util.concurrent.TimedOutError;
 import org.eclipse.scout.rt.server.commons.authentication.IAccessController;
 import org.eclipse.scout.rt.server.commons.servlet.CookieUtility;
 import org.eclipse.scout.rt.server.commons.servlet.HttpClientInfo;
@@ -593,9 +596,14 @@ public class UiSession implements IUiSession {
     uninstallUiDataAvailableListener();
     signalPoller(); // Notify waiting requests - should not delay web-container shutdown
 
-    m_jsonAdapterRegistry.disposeAdapters();
-    m_httpContext.clear();
-    m_currentJsonResponse = null;
+    // The model may run tasks right now, that alter json adapters, and this may lead to an unreliable state of the JSON layer.
+    // Example: creating and initializing a new json adapter is NOT done atomically and disposing an adapter before it is initialized may lead to errors depending on thy type of json adapter.
+    // Therefore, run this in a model job to ensure a consistent state of the JSON layer before disposing it.
+    runInModelJob(() -> {
+      m_jsonAdapterRegistry.disposeAdapters();
+      m_httpContext.clear();
+      m_currentJsonResponse = null;
+    }, "Disposing Ui session", true);
 
     // Inform the desktop that the UI has been detached.
     // The model may trigger events during detaching the desktop, that need to be sent back to the browser if the ui session weren't disposed.
@@ -603,6 +611,60 @@ public class UiSession implements IUiSession {
     detachDesktop();
 
     m_sessionMetrics.sessionDestroyed(SESSION_TYPE);
+  }
+
+  /**
+   * Run the given {@link Runnable} in a model job. If the current thread is the model thread or there is no active
+   * client session (see {@link #getClientSession()}) the {@link Runnable} is simply executed. If wait is set to
+   * <code>true</code>, the method waits for the model job to complete for 5 seconds. If it does not complete the
+   * {@link Runnable} is executed in the current thread.
+   * 
+   * @param runnable
+   *          {@link Runnable} to be executed.
+   * @param name
+   *          name of the model job that is triggered.
+   * @param wait
+   *          If {@code true}, method waits for the model job to complete.
+   */
+  protected void runInModelJob(Runnable runnable, String name, boolean wait) {
+    assertNotNull(runnable, "runnable is required");
+
+    // current thread is the model thread -> run
+    if (ModelJobs.isModelThread()) {
+      runnable.run();
+      return;
+    }
+
+    // current thread is NOT the model thread but there is no active client session -> run
+    final IClientSession clientSession = getClientSession();
+    if (clientSession == null || !clientSession.isActive() || clientSession.isStopping()) {
+      runnable.run();
+      return;
+    }
+
+    // current thread is NOT the model thread and the client session is active -> run in model job
+    final ClientRunContext clientRunContext = ClientRunContexts.copyCurrent(true).withSession(clientSession, true);
+    IFuture<Void> modelJob = ModelJobs.schedule(
+        runnable::run,
+        ModelJobs.newInput(clientRunContext)
+            .withName(name)
+            .withExceptionHandling(null, false)); // Propagate exception to caller
+
+    // no need to wait -> return
+    if (!wait) {
+      return;
+    }
+
+    try {
+      // wait for the model job to complete
+      modelJob.awaitDone(5, TimeUnit.SECONDS);
+    }
+    catch (TimedOutError | ThreadInterruptedError | FutureCancelledError e) {
+      LOG.warn("Unable to run task '{}' in model job -> run it in current thread.", name, e);
+      // the model job did not complete within wait time -> cancel it and run task in current thread
+      modelJob.cancel(true);
+      runnable.run();
+    }
   }
 
   protected void detachDesktop() {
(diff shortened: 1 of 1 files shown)


fschinkel:
run tests

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiSession.java
```diff
@@ -605,6 +611,52 @@ public void dispose() {
     m_sessionMetrics.sessionDestroyed(SESSION_TYPE);
   }
 
+  protected void runInModelJob(Runnable r, String name) {
+    runInModelJob(r, name, false);
+  }
+
+  protected void runInModelJob(Runnable r, String name, boolean wait) {
```
bschwarzent: Short variable names don't make the code run faster: `r` -> `runnable`

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiSession.java
```diff
@@ -605,6 +611,52 @@ public void dispose() {
     m_sessionMetrics.sessionDestroyed(SESSION_TYPE);
   }
 
+  protected void runInModelJob(Runnable r, String name) {
```
bschwarzent: Necessary? There is only one call, I think it would be okay if it explicitly specified the third parameter.

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiSession.java
```diff
@@ -605,6 +611,52 @@ public void dispose() {
...
+    runInModelJob(r, name, false);
+  }
+
+  protected void runInModelJob(Runnable r, String name, boolean wait) {
+    if (r == null) {
```
bschwarzent: assertNotNull?

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiSession.java
```diff
@@ -605,6 +611,52 @@ public void dispose() {
     m_sessionMetrics.sessionDestroyed(SESSION_TYPE);
   }
 
+  protected void runInModelJob(Runnable r, String name) {
+    runInModelJob(r, name, false);
+  }
+
```
bschwarzent: Add JavaDoc and specify the arguments (especially "wait")

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiSession.java
```diff
@@ -605,6 +611,52 @@ public void dispose() {
...
+      }
+      catch (TimedOutError e) {
+        // the model job did not complete within wait time -> cancel it and do NOT return but run task in current thread
+        modelJob.cancel(true);
+        LOG.debug("Unable to run task \"{}\" in model job.", name);
+      }
+    }
+
+    // current thread is NOT the model thread but there is no active client session or the wait time exceeded -> run
+    r.run();
```
bschwarzent: I would add this to the "catch" clause (and instead remove the return in the "try" part.

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiSession.java
```diff
@@ -605,6 +611,52 @@ public void dispose() {
...
+
+      try {
+        // wait for the model job to complete
+        modelJob.awaitDone(2, TimeUnit.SECONDS);
+        return;
+      }
+      catch (TimedOutError e) {
+        // the model job did not complete within wait time -> cancel it and do NOT return but run task in current thread
+        modelJob.cancel(true);
+        LOG.debug("Unable to run task \"{}\" in model job.", name);
```
bschwarzent: - Move log line to start of catch block
- Shouldn't the log level be at least INFO? This should happen very rarely and it would probably be good to see if it doesn't.
- We usually write ' inside log strings to avoid the escape characters

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiSession.java
```diff
@@ -605,6 +611,52 @@ public void dispose() {
...
+      if (!wait) {
+        return;
+      }
+
+      try {
+        // wait for the model job to complete
+        modelJob.awaitDone(2, TimeUnit.SECONDS);
+        return;
+      }
+      catch (TimedOutError e) {
```
bschwarzent: What about FutureCancelledError and ThreadInterruptedError?

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiSession.java
```diff
@@ -605,6 +611,52 @@ public void dispose() {
...
+              .withExceptionHandling(null, false));// Propagate exception to caller (UIServlet)
+
+      // no need to wait -> return
+      if (!wait) {
+        return;
+      }
+
+      try {
+        // wait for the model job to complete
+        modelJob.awaitDone(2, TimeUnit.SECONDS);
```
bschwarzent: Not sure about the timeout duration. Having a timeout is a safety measure against deadlocks or forever blocked threads. But since we are in the process of disposing from outside a model job (i.e. valueUnbound), the user would not notice if the waiting time was longer. Maybe 30 or 60 seconds would be more appropriate?

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiSession.java
```diff
@@ -605,6 +611,52 @@ public void dispose() {
...
+    if (ModelJobs.isModelThread()) {
+      r.run();
+      return;
+    }
+
+    // current thread is NOT the model thread and the client session is active -> run in model job
+    final IClientSession clientSession = getClientSession();
+    if (clientSession != null && clientSession.isActive() && !clientSession.isStopping()) {
+      final ClientRunContext clientRunContext = ClientRunContexts.copyCurrent(true).withSession(clientSession, true);
+      IFuture<Void> modelJob = ModelJobs.schedule(
```
bschwarzent: More readable?

```
      IFuture<Void> modelJob = ModelJobs.schedule(r::run, ModelJobs.newInput(clientRunContext)
              .withName(name)
              .withExceptionHandling(null, false));// Propagate exception to caller (UIServlet)
```

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiSession.java
```diff
@@ -605,6 +611,52 @@ public void dispose() {
...
+          r::run,
+          ModelJobs.newInput(clientRunContext)
+              .withName(name)
+              .withExceptionHandling(null, false));// Propagate exception to caller (UIServlet)
```
bschwarzent: Missing space between code and comment.

Is the comment still accurate? Is it always the UiServlet? If yes, a lowercase "i" should be used. If not, maybe it is better to remove the "(UIServlet)" part entirely.

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiSession.java
```diff
@@ -593,9 +594,14 @@ public void dispose() {
...
-    m_httpContext.clear();
-    m_currentJsonResponse = null;
+    // The model may run tasks right now, that alter json adapters, and this may lead to an unreliable state of the JSON layer.
+    // Example: creating and initializing a new json adapter is NOT done atomically and disposing an adapter before it is initialized may lead to errors depending on thy type of json adapter.
+    // Therefore, run this in a model job to ensure a consistent state of the JSON layer before disposing it.
+    runInModelJob(() -> {
+      m_jsonAdapterRegistry.disposeAdapters();
+      m_httpContext.clear();
+      m_currentJsonResponse = null;
+    }, "Dispose UI session", true);
```
bschwarzent: Jobs names should be more consistent.
- "Dispose UI session" --> "Dispose" or "Disposing"? "UI" or "Ui"?
- "Detaching Gui" --> "Detaching" or "Detach"? "Gui", "GUI" or "UI"?

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiSession.java
fschinkel: I do not think so. I find the indentation misleading in this case, this suggests that `withName` is a method called on the scheduled job not on the input.
Consider a potential `whenDone` call on a new line at the end. `withName` is called on the result of `ModelJobs.newInput` while `whenDone` is called on the result of `ModelJobs.schedule`, but both lines will have the same indentation.
Therefore, I prefer one level of indentation per block that is not closed on the same line it is opened. In this case one for the `schedule`-method-parameters and one for the `input`-method-chaining.

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiSession.java
fschinkel: The user will not notice this delay if he closes the browser or navigates to another url in general, but he will notice something if the new url is a request to `/logout`. If the timeout is e.g. 60 seconds and there is a long running task blocking dispose, the user will wait 60s until the logout screen is displayed.

PR #910, opened 2024-02-19
Tree: improve tooltip position

bschwarzent:
A tree node is always as wide as the entire tree. By default, tooltips are displayed at the center of the anchor. In the case of tree nodes, the tooltip may therefore point to an empty space in the middle of the tree.

By providing a custom "origin producer" function, the tooltip position can be adjusted such that it points to the center of the text (or the icon if the node has no text).

graphics.prefSize() now adds a marker CSS class 'measure' while measuring to allow elements to assume their natural size.

327222

### DIFF
--- a/eclipse-scout-core/src/layout/graphics.ts
+++ b/eclipse-scout-core/src/layout/graphics.ts
@@ -142,8 +142,12 @@ export const graphics = {
 
   /**
    * Returns the preferred size of $elem.
-   * Precondition: $elem and it's parents must not be hidden (display: none. Visibility: hidden would be ok
-   * because in this case the browser reserves the space the element would be using).
+   *
+   * Precondition: $elem and its parents must not be hidden ('display: none' - other styles like 'visibility: hidden'
+   * or 'opacity: 0' would be ok because in this case the browser reserves the space the element would be using).
+   *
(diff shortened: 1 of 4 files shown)


eclipse-scout-core/src/tree/Tree.ts
```diff
@@ -1320,10 +1320,45 @@ export class Tree extends Widget implements TreeModel {
       text: this._nodeTooltipText.bind(this),
       arrowPosition: 50,
       arrowPositionUnit: '%',
-      nativeTooltip: !Device.get().isCustomEllipsisTooltipPossible()
+      nativeTooltip: !Device.get().isCustomEllipsisTooltipPossible(),
+      originProducer: this._treeNodeTooltipOrigin.bind(this)
     });
   }
 
+  _treeNodeTooltipOrigin($node) {
```
cguglielmo: Please add type and protected modifier

eclipse-scout-core/src/layout/graphics.ts
```diff
@@ -208,6 +213,7 @@ export const graphics = {
     }
 
     // modify properties which prevent reading the preferred size
+    $elem.addClass('measure');
```
cguglielmo, line 216: Maybe remove measure class in DesktopNavigationLayout.ts because this makes it obsolete

PR #913, opened 2024-02-22
AppLinkKeyStroke: add enter key and make it work when field is disabled

cguglielmo:
App links can be clicked even if the field is disabled. So it should be possible to activate them using a keyboard as well.

Regular links can be activated by pressing 'enter', app links need to be activated by pressing 'space'. This is confusing -> add enter key to AppLinkKeyStroke.


cguglielmo:
run tests

PR #919, opened 2024-02-28
Harmonize revalidation of layout

bschwarzent:
Widgets that reposition themselves needs to revalidate their layout. This should be done via revalidateLayoutTree(false), because otherwise, scheduled post validate functions are not executed.

369849


bschwarzent:
Run tests

bschwarzent:
Run tests

bschwarzent:
Abandoned due to unpredictable changes in behavior, leading to potential bugs.

PR #941, opened 2024-03-07
Expose ConfigProperties to Scout JS

mvilliger:
- Add new config utility in Scout JS to access config properties from backend systems
- Add REST resource and ResourceLoader to fetch selected config properties from UI server and backend server
- Add contributor to expose config properties
- Add function to convert IConfigProperties to ConfigPropertyDo
- Replace existing properties sent to the browser with new config utility:
  - devMode
  - application.version
  - ui.backgroundPollingMaxWaitTime
  - uinotification.waitTimeout

373066


org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/config/AbstractConfigPropertyToDoFunction.java
```diff
@@ -0,0 +1,42 @@
...
+package org.eclipse.scout.rt.api.data.config;
+
+import org.eclipse.scout.rt.dataobject.mapping.AbstractToDoFunction;
+import org.eclipse.scout.rt.platform.config.IConfigProperty;
+
+/**
+ * Base implementation of a function that converts {@link IConfigProperty} to {@link ConfigPropertyDo}. Only
+ * {@link IConfigProperty#getKey()} and {@link IConfigProperty#getValue()} is exported.
+ */
+public abstract class AbstractConfigPropertyToDoFunction<EXPLICIT_SOURCE extends IConfigProperty<?>, EXPLICIT_TARGET extends ConfigPropertyDo>
```
paolobazzi: I recently moved some other ToDo-functions to a ".mapping" package to separate them from the business objects like permissions or config properties here. Suggestion: move these new functions to a .mapping package as well.

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/config/ConfigPropertyDo.java
```diff
@@ -0,0 +1,75 @@
...
+   * *************************************************************************/
+
+  /**
+   * Converts an {@link IConfigProperty} to a DO.
+   *
+   * @param property
+   *     The property to convert.
+   * @return The corresponding {@link ConfigPropertyDo} or {@code null} if it cannot be converted.
+   */
+  public static ConfigPropertyDo create(IConfigProperty<?> property) {
```
paolobazzi: As long as this method is just called in AbstractApiExposedConfigPropertyContributor.convert() I would move this helper code there and remove this method on this class.
Other API Data Objects which do not know the correspondent business object (like permission) also do not provide such convenience.

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/config/ConfigPropertyDo.java
```diff
@@ -0,0 +1,75 @@
...
+import org.eclipse.scout.rt.dataobject.TypeName;
+import org.eclipse.scout.rt.dataobject.mapping.ToDoFunctionHelper;
+import org.eclipse.scout.rt.platform.BEANS;
+import org.eclipse.scout.rt.platform.config.IConfigProperty;
+
+/**
+ * Do for {@link IConfigProperty}. Use {@link #create(IConfigProperty)} to convert a property to a DO.
+ */
+@TypeName("scout.ConfigProperty")
+public class ConfigPropertyDo extends DoEntity {
```
paolobazzi, line 31: Add disclaimer that this DO should only be used to return config properties as response to a ScoutJS REST Client.
A Java REST layer will (without any further type information about the value) not be able to deserialize an arbitrary object value.
This also will just work for properties having a Jackson-serializable value type. An arbitrary POJO is not serializable as JSON.
As a hint for the caller, adding type information as additional field or using different config property DOs for various base types would be helpful.

eclipse-scout-core/src/code/codes.ts
```diff
@@ -29,6 +29,14 @@ export const codes = {
     return promise.then(codes._preInit.bind(this, url));
   },
 
+  /**
+   * load codes from the main system
```
cguglielmo: Please use third person -> Loads. Same for bootstrap method
Also write it more clearly what it does, as you have done with config.ts

eclipse-scout-core/src/config/config.ts
```diff
@@ -0,0 +1,116 @@
...
+   */
+  bootstrap(urls: string | string[], system?: string): JQuery.Promise<void> {
+    let promises = arrays.ensure(urls)
+      .map(url => $.ajaxJson(url)
+        .then(properties => config.init(properties, system)));
+    return $.promiseAll(promises);
+  },
+
+  /**
+   * Load config properties from the given system using the configuration of that {@link System}.
```
cguglielmo: Third person, same for init

eclipse-scout-core/src/config/config.ts
```diff
@@ -0,0 +1,116 @@
...
+  /**
+   * @internal
+   */
+  _getSystemMap(system?: string): Map<string, ConfigProperty<any>> {
+    return config.configMap.get(system || System.MAIN_SYSTEM);
+  },
+
+  /**
+   * Gets the value of the property with given key. Optionally from a specific backend system.
+   * <p>
```
cguglielmo: Use markdown syntax (not html)

eclipse-scout-core/src/config/config.ts
```diff
@@ -0,0 +1,116 @@
...
+      config.configMap.set(system || System.MAIN_SYSTEM, systemPropertyMap);
+    }
+    let property = {key, value};
+    systemPropertyMap.set(property.key, property);
+    return property;
+  },
+
+  /**
+   * Loads the properties from the system given and returns the property with given key.
+   * <p>
```
cguglielmo: remove html

eclipse-scout-core/src/config/config.ts
```diff
@@ -0,0 +1,116 @@
...
+      systemPropertyMap = new Map<string, ConfigProperty<any>>();
+      config.configMap.set(system || System.MAIN_SYSTEM, systemPropertyMap);
+    }
+    let property = {key, value};
+    systemPropertyMap.set(property.key, property);
+    return property;
+  },
+
+  /**
+   * Loads the properties from the system given and returns the property with given key.
```
cguglielmo: system given -> given system
with given key -> with the given key

eclipse-scout-core/src/config/config.ts
```diff
@@ -0,0 +1,116 @@
...
+   * <p>
+   * To configure the system URL use {@link systems.getOrCreate} and {@link System.setEndpointUrl} for endpointName 'config-properties' if required.
+   * @param key The key of the property.
+   * @param system An optional system to which the property belongs. By default, {@link System.MAIN_SYSTEM} is used.
+   */
+  load<TKey extends keyof ConfigProperties[TSystem] & string, TSystem extends keyof ConfigProperties & string = 'main'>(key: TKey, system?: TSystem): JQuery.Promise<ConfigProperties[TSystem][TKey]> {
```
cguglielmo: the first call writes the properties into the registry. If the property changes and is loaded a second time, there will be a warning, won't it? 
Wouldn't it be better to only load the desired property (not all) without writing it into the registry?

eclipse-scout-core/src/security/access.ts
```diff
@@ -24,6 +24,11 @@ export const access = {
     return accessControl.whenSync();
   },
 
+  bootstrapSystem(): JQuery.Promise<void> {
```
cguglielmo, line 34: Add similar jsdoc as for config.bootstrap

eclipse-scout-core/src/system/System.ts
```diff
@@ -0,0 +1,158 @@
...
+  getEndpointUrl(endpointName: string, defaultEndpoint?: string): string {
+    let customizedEndpointUrl = this._endpointUrls.get(endpointName);
+    let relEndpointUrl = customizedEndpointUrl || defaultEndpoint;
+    if (!relEndpointUrl) {
+      return null;
+    }
+    return this._concatPath(this.baseUrl, relEndpointUrl);
+  }
+
+  protected _concatPath(a: string, b: string): string {
```
cguglielmo, line 116: Maybe simplify using strings.removePrefix, strings.removeSuffix, array.join?

eclipse-scout-core/src/system/systems.ts
```diff
@@ -0,0 +1,41 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {scout, System} from '../index';
+
+/**
+ * Utility to retrieve and create {@link System} instances.
+ * A {@link System} represents a REST backend (typically Scout UI or server backend).
+ */
+export const systems = {
+  _systemsMap: new Map<string, System>(),
```
cguglielmo, line 17: Move outside of object to make it private?

eclipse-scout-core/src/system/systems.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * Retrieves (and creates if not yet existing) the {@link System} with given name.
+   * @param name The optional name of the System. If omitted, {@link System.MAIN_SYSTEM} is used.
+   * @param baseUrl The base URL of the REST resources of the system. This is the (typically relative) URL under which all the REST endpoints are available. If omitted, 'api/' is used as a default.
```
cguglielmo: 'api/' -> 'api'. Otherwise, it looks like a slash is required when registering a system

scout-hellojs-app/src/main/resources/archetype-resources/__rootArtifactId__.app/src/main/js/__simpleArtifactName__.ts
```diff
@@ -1,9 +1,20 @@
+/*
+ * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
```
cguglielmo: remove copyright header

eclipse-scout-core/src/App.ts
```diff
@@ -385,10 +392,7 @@ export class App extends EventEmitter {
   }
 
   protected _initVersion(options: AppModel) {
-    this.version = scout.nvl(
-      this.version,
-      options.version,
-      $('scout-version').data('value'));
+    this.version = scout.nvl(this.version, options.version, config.get('scout.application.version'));
```
cguglielmo: I don't think the version tag in the html can be replaced -> we need to discuss it

eclipse-scout-core/src/config/config.ts
mvilliger: The idea of this solution is that you have the possibility to preload the properties from a system, if you know that you won't need it before and afterwards you can access them sync. Therefore all properties are cached to allow a later sync access. The warning is changed to a info log in a subsequent commit. The old value is then overwritten.

eclipse-scout-core/src/system/systems.ts
mvilliger: It was added to the object as private map intentionally because it simplifies debugging (e.g. easy inspect of the map in the console)

PR #944, opened 2024-03-11
MimeType: add missing mime type for mjs

fschinkel:
377307

### DIFF
--- a/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/MimeType.java
+++ b/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/MimeType.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


mvilliger (changes requested):
My first question would be: how did we loose this mime type?
Which other mime types have we lost the same way? All these should be added again?

General:
application/javascript is obsolete.
Maybe we should update to the new "text/javascript"? Do all browser support that?
See chapter 6 in https://datatracker.ietf.org/doc/rfc9239/ and https://www.iana.org/assignments/media-types/media-types.xhtml and https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

Mabe we should add all of the common types of the MDN to have the most important ones?

fschinkel:
> My first question would be: how did we loose this mime type? Which other mime types have we lost the same way? All these should be added again?
> 
> General: application/javascript is obsolete. Maybe we should update to the new "text/javascript"? Do all browser support that? See chapter 6 in https://datatracker.ietf.org/doc/rfc9239/ and https://www.iana.org/assignments/media-types/media-types.xhtml and https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
> 
> Mabe we should add all of the common types of the MDN to have the most important ones?

We lost the mime type by this [commit](https://github.com/eclipse-scout/scout.rt/commit/8cc7f2321955660f6ba60c769b7910d792e3d73a) that deleted the _ServletContextMimeTypeDetector_.
There is a ticket already for those deleting the MimeTypeDetector to fix all other mime types that were lost by that change (see 377371).

mvilliger:
> 
> We lost the mime type by this [commit](https://github.com/eclipse-scout/scout.rt/commit/8cc7f2321955660f6ba60c769b7910d792e3d73a) that deleted the _ServletContextMimeTypeDetector_. There is a ticket already for those deleting the MimeTypeDetector to fix all other mime types that were lost by that change (see 377371).

I think we should add the missing ones directly now to ensure no other side effects happen instead of delegating that to a later fix?

mvilliger (approved):

PR #945, opened 2024-03-11
Update browser cache on backend Permission & CodeType invalidate

mvilliger:
373498


eclipse-scout-core/src/code/codes.ts
```diff
@@ -41,14 +43,23 @@ export const codes = {
         url: url
       };
     }
-    codes.init(data);
+    codes.init(data)
```
cguglielmo: Subscribe should be done before codes are initialized. Otherwise there is a chance that the ui notification is missed if the codes change between init and subscribe

eclipse-scout-core/src/code/codes.ts
```diff
@@ -25,12 +25,14 @@ export const codes = {
...
-    return promise.then(codes._preInit.bind(this, url));
+    if (!url) {
+      return $.resolvedPromise();
+    }
+    return $.ajaxJson(url).then(codes._handleBootstrapResponse.bind(this, url));
   },
 
   /** @internal */
-  _preInit(url: string, data: any) {
+  _handleBootstrapResponse(url: string, data: any) {
```
cguglielmo: the _preInit name is used for texts.ts and locales.ts as well. Please change these to _handleBootstrapResponse as well

org.eclipse.scout.rt.uinotification/src/main/java/org/eclipse/scout/rt/api/uinotification/UiNotificationRegistry.java
```diff
@@ -372,6 +377,21 @@ public void removeListener(String topic, UiNotificationListener listener) {
     }
   }
 
+  /**
+   * Retrieves how many observer are listening for the topic given.
```
cguglielmo: observer -> observers
topic given -> given topic
 for given topic. -> for the given topic.

org.eclipse.scout.rt.uinotification/src/main/java/org/eclipse/scout/rt/api/uinotification/UiNotificationRegistry.java
```diff
@@ -185,14 +188,14 @@ protected List<UiNotificationDo> get(String topic, String user, final List<UiNot
...
    * <p>
-   * This is necessary to ensure the client receives every notification from now on even if the connection
-   * temporarily drops before the first real notification can be sent.
-   * During that connection drop a notification could be added that needs to be sent as soon as the connection is reestablished again.
+   * This is necessary to ensure the client receives every notification from now on even if the connection temporarily
+   * drops before the first real notification can be sent. During that connection drop a notification could be added
+   * that needs to be sent as soon as the connection is reestablished again.
    * <p>
-   * The result can contain multiple notifications, one for each cluster node.
-   * So, if a cluster node created notifications for the topic, the last one for that node will be returned.
```
cguglielmo, line 193: Why have you adjusted this? I find it more readable, if every sentence is on one line. We do the same for markdown and ascii doc text.

org.eclipse.scout.rt.uinotification/src/test/java/org/eclipse/scout/rt/api/uinotification/UiNotificationRegistryTest.java
```diff
@@ -49,6 +49,23 @@ public void before() {
     m_registry.setIdGenerator(new TestIdGenerator());
   }
 
+  @Test
+  public void testIsListenedForTopic() {
```
cguglielmo: rename to testGetListenerCount

eclipse-scout-core/src/security/AccessControl.ts
```diff
@@ -63,10 +67,41 @@ export class AccessControl extends PropertyEventEmitter implements AccessControl
 
   startSync() {
     this._sync();
+    this._subscribeForNotifications();
```
cguglielmo, line 45: subscribe first, then sync

eclipse-scout-core/src/security/AccessControl.ts
```diff
@@ -25,16 +25,18 @@ export class AccessControl extends PropertyEventEmitter implements AccessControl
...
   protected _retryIntervals: number[];
-  protected _syncTimeoutId: number;
+  protected _permissionUpdateEventHandler: EventHandler<UiNotificationEvent>;
+  protected _reloadTimeoutId;
 
   constructor() {
     super();
     this.permissionsUrl = null;
-    this.interval = 1800000; // 30 * 60 * 1000 (30 minutes)
+    this.interval = 1_800_000; // 30 * 60 * 1000 (30 minutes)
```
cguglielmo: I think we should remove this.interval, because it does not reload every 30 min anymore. This variable now only specifies the limit for the retryIntervals array.
Actually, I don't think it should ever stop retrying, unless the session is invalidated.

eclipse-scout-core/src/security/AccessControl.ts
```diff
@@ -52,9 +54,11 @@ export class AccessControl extends PropertyEventEmitter implements AccessControl
     const retryIntervals = [];
     if (interval) {
       let retryInterval = 1000;
-      while (retryInterval < interval) {
+      let factor = 2;
+      let limit = interval / factor; // retryInterval is the time between retries
+      while (retryInterval < limit) {
         retryIntervals.push(retryInterval);
-        retryInterval = retryInterval * 2;
```
cguglielmo, line 57: Doubling the retryInterval is probably not ideal, because it will soon be a very large interval. I know, it is nothing new, but I think we should change it. Maybe we have to adjust Call.ts to support this requirement, e.g. we could set maxRetries to -1 which means no limit and update this member based on the retryInterval only if maxRetries is null

eclipse-scout-core/src/security/AccessControl.ts
```diff
@@ -63,10 +67,41 @@ export class AccessControl extends PropertyEventEmitter implements AccessControl
...
+  protected _subscribeForNotifications() {
+    uiNotifications.subscribe('permissionsUpdate', this._permissionUpdateEventHandler);
+  }
+
+  protected _unsubscribeFromNotifications() {
+    uiNotifications.unsubscribe('permissionsUpdate', this._permissionUpdateEventHandler);
+  }
+
+  protected _onPermissionUpdateNotify(event: UiNotificationEvent) {
+    let reloadDelayWindow = scout.nvl(event.message['reloadDelayWindow'], 0) as number;
```
cguglielmo: Better create PermissionUpdateNotificationDo.ts which contains reloadDelayWindow as typed member to make clear what notification is expected

eclipse-scout-core/src/uinotification/UiNotificationPoller.ts
```diff
@@ -109,6 +109,7 @@ export class UiNotificationPoller extends PropertyEventEmitter {
   }
 
   protected _poll() {
+    this._call?.abort(); // abort in case there is already a call running
```
cguglielmo, line 125: This should be fine but what makes it necessary? How can it happen?

eclipse-scout-core/src/code/codes.ts
```diff
@@ -84,16 +60,17 @@ export const codes = {
...
+  },
 
-    for (let codeType of codes.registry.values()) {
-      if (codeType instanceof codeTypeIdOrClassRef) {
-        return codeType as T;
-      }
+  /**
+   * @returns the global {@link CodeTypeCache} instance. If required, a new one is created (on first use).
+   */
+  getCodeTypeCache(): CodeTypeCache {
```
cguglielmo: I would just name it getCache.

eclipse-scout-core/src/code/CodeTypeCache.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  id: string;
+  /**
+   * Map of CodeType id to CodeType instance. Do not access directly. Instead, use {@link get}.
+   */
+  registry: Map<any /* CodeType id */, CodeType<any, any, any>>;
+
+  constructor() {
+    super();
+    this.events.registerSubTypePredicate('codeTypeAdd', (event: CodeTypeAddEvent, codeTypeId) => event?.codeType?.id === codeTypeId); // only works if the CodeTypeId can be converted to a string
```
cguglielmo: How can event and codetype be null? I don't see the reason why it is necessary to add the ? operator

eclipse-scout-core/src/util/Call.ts
```diff
@@ -64,15 +66,13 @@ export abstract class Call implements CallModel, ObjectWithType {
...
-    } else {
-      // Do not modify the passed value -> create a copy
-      this.retryIntervals = (this.retryIntervals ? this.retryIntervals.slice() : []);
-      // Remember initial number of retries (for logging)
+    this.retryIntervals = (this.retryIntervals ? this.retryIntervals.slice() : []); // Do not modify the passed value -> create a copy
+    if (objects.isNullOrUndefined(this.maxRetries)) {
       this.maxRetries = this.retryIntervals.length;
     }
+    if (objects.isNullOrUndefined(this.defaultRetryInterval)) {
+      this.defaultRetryInterval = 300;
```
cguglielmo: Just set it in constructor?

mvilliger: The idea behind this is, that if the model contains null for defaultRetryInterval, the default don't get overwritten with null. What do you mean?

PR #950, opened 2024-03-14
Expose SharedVariableMap to Scout JS

mvilliger:
Synchronize selected variables from the SharedVariableMap of the
ClientSession to the browser.

362057


eclipse-scout-core/src/session/SessionAdapter.ts
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {ModelAdapter} from './ModelAdapter';
```
cguglielmo: Import from index

eclipse-scout-core/src/table/Table.ts
```diff
@@ -4901,6 +4901,38 @@ export class Table extends Widget implements TableModel {
...
+      return;
+    }
+
+    const estimatedRowCount = this.estimatedRowCount;
+    const numRows = this.rows.length;
+    const decimalFormat = this.session.locale.decimalFormat;
+    const showingRowCountText = decimalFormat.format(numRows);
+    const estimatedRowCountText = decimalFormat.format(estimatedRowCount);
+    let message: string;
+    if (Device.get().supportsOnlyTouch()) {
```
cguglielmo: I think we should change this to Device.get().type === Device.Type.MOBILE, also in Java. I think this is only relevant for small screens

org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/table/MaxResultsHelper.java
```diff
@@ -0,0 +1,174 @@
...
+    /**
+     * Removes surplus items from the data {@link List} if the maximum number of results is exceeded and attaches a
+     * {@link LimitedResultInfoContributionDo} as contribution to the given {@link IDoEntity}. The contribution contains
+     * metadata about the limit operation:
+     * <ol>
+     * <li>If surplus items have been removed from the list
+     * ({@link LimitedResultInfoContributionDo#isLimitedResult()})</li>
+     * <li>The maximum number of entries that would have been allowed
+     * ({@link LimitedResultInfoContributionDo#getMaxRowCount()})</li>
+     * <li>({@link LimitedResultInfoContributionDo#getEstimatedRowCount()}) is currently not set and will always be <=
```
cguglielmo: Maybe we should add a member, so that this feature can be enabled? I tried to test it but I cannot enable it

cguglielmo (approved):

PR #951, opened 2024-03-14
Device: fix unwanted tablet mode for Firefox on Windows 11

bschwarzent:
Firefox on Windows 11 always uses "pretty" (floating) scrollbars that have a width of 0. Because Device#isWindowsTabletMode() assumes that a width of 0 indicates that the "tablet mode" is active, Firefox users on Windows 11 always get the tablet version of the application, even when they are using a normal desktop computer.

To fix this, the implementation of isWindowsTableMode() is changed. Instead of checking the scrollbar width, a media query is used to identify the type of the _primary_ input method.
- A normal desktop computer or a convertible in "desktop mode" uses a mouse cursor as primary input method, which is reported as 'fine'.
- A touch-only device or a convertible in "tablet mode" does not have a mouse. Therefore, the primary input method is reported as 'coarse'.

Note that Windows 11 no longer provides an option to manually activate or deactivate the "tablet mode". Instead, the mode is automatically activated or deactivated when the device is docked or undocked.

363875

### DIFF
--- a/eclipse-scout-core/src/util/Device.ts
+++ b/eclipse-scout-core/src/util/Device.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


mvilliger (approved):
Works correctly now in Firefox and Chromium based browsers on Windows convertibles (e.g. Surface Pro) in docked mode (with touchpad) and in tablet mode (with touch screen).

PR #954, opened 2024-03-15
Table: fix tooltip position when rows are updated

bschwarzent:
When a cell tooltip is open while the row is updated, the tooltip is updated because the new row might have a different tooltip text. Because the tooltip support reuses the same tooltip instance, the $anchor has to be updated as well. Otherwise, the position will be wrong, because the old cell div is no longer in the DOM.

Certain other properties also need to be updated on the existing tooltip instance. For example, if the text is set to an empty string, the tooltip should be closed.

377024

### DIFF
--- a/eclipse-scout-core/src/table/Table.ts
+++ b/eclipse-scout-core/src/table/Table.ts
@@ -14,7 +14,7 @@ import {
   PropertyChangeEvent, Range, scout, scrollbars, ScrollToOptions, Status, StatusOrModel, strings, styles, TableCompactHandler, TableControl, TableCopyKeyStroke, TableEventMap, TableFooter, TableHeader, TableLayout, TableModel,
   TableNavigationCollapseKeyStroke, TableNavigationDownKeyStroke, TableNavigationEndKeyStroke, TableNavigationExpandKeyStroke, TableNavigationHomeKeyStroke, TableNavigationPageDownKeyStroke, TableNavigationPageUpKeyStroke,
   TableNavigationUpKeyStroke, TableRefreshKeyStroke, TableRow, TableRowModel, TableSelectAllKeyStroke, TableSelectionHandler, TableStartCellEditKeyStroke, TableTextUserFilter, TableTileGridMediator, TableToggleRowKeyStroke, TableTooltip,
-  TableUpdateBuffer, TableUserFilter, TableUserFilterModel, Tile, TileTableHeaderBox, tooltips, UpdateFilteredElementsOptions, ValueField, Widget
+  TableUpdateBuffer, TableUserFilter, TableUserFilterModel, Tile, TileTableHeaderBox, tooltips, TooltipSupport, UpdateFilteredElementsOptions, ValueField, Widget
 } from '../index';
 import $ from 'jquery';
 
@@ -2007,6 +2007,10 @@ export class Table extends Widget implements TableModel {
       return;
     }
 
+    let tooltipSupport = this.$data.data('tooltipSupport') as TooltipSupport;
+    if ($row.isOrHas(tooltipSupport?.tooltip?.$anchor)) {
+      tooltipSupport.close();
+    }
     this._destroyTooltipsForRow(row);
     this._destroyCellEditorForRow(row);
 
@@ -3115,6 +3119,8 @@ export class Table extends Widget implements TableModel {
   }
 
   protected _renderUpdateRows(rows: TableRow[], oldRowsMap: Record<string, TableRow>) {
+    let tooltipSupport = this.$data.data('tooltipSupport') as TooltipSupport;
+
     // render row and replace div in DOM
     rows.forEach(row => {
       let oldRow = oldRowsMap[row.id];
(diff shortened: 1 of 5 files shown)


eclipse-scout-core/src/tooltip/TooltipSupport.ts
```diff
@@ -134,21 +134,29 @@ export class TooltipSupport {
...
+      this._destroyTooltip();
+      return;
     }
 
+    let $anchor = this._options.$anchor || $comp;
     let htmlEnabled = this._htmlEnabled($comp);
 
     if (this._tooltip && this._tooltip.rendered) {
       // update existing tooltip
+      let anchorChanged = this._tooltip.$anchor !== $anchor;
```
cguglielmo: This is probably always true. You need to compare the html elements.

eclipse-scout-core/src/tooltip/TooltipSupport.ts
```diff
@@ -134,21 +134,29 @@ export class TooltipSupport {
...
+      this._tooltip.$anchor = $anchor;
```
cguglielmo: I would add setters for the properties on the tooltip class and do the position update in set$Anchor to make it work without a tooltip support as well.

eclipse-scout-core/src/tooltip/TooltipSupport.ts
```diff
@@ -134,21 +138,25 @@ export class TooltipSupport {
...
+      this._tooltip.set$Anchor($anchor);
+      this._tooltip.setHtmlEnabled(htmlEnabled);
       this._tooltip.setText(text);
       this._tooltip.setSeverity(this._options.severity);
       this._tooltip.setMenus(this._options.menus);
     } else {
       // create new tooltip
       let options = $.extend({}, this._options, {
-        $anchor: this._options.$anchor || $comp,
+        $anchor: $anchor,
```
fschinkel, line 159:
```suggestion
      let options = $.extend({}, this._options, {$anchor, text, htmlEnabled});
```

eclipse-scout-core/src/table/Table.ts
```diff
@@ -2007,6 +2007,10 @@ export class Table extends Widget implements TableModel {
       return;
     }
 
+    let tooltipSupport = this.$data.data('tooltipSupport') as TooltipSupport;
+    if (tooltipSupport && tooltipSupport.tooltip && $row.isOrHas(tooltipSupport.tooltip.$anchor)) {
```
fschinkel: maybe use ?-notation

eclipse-scout-core/src/table/Table.ts
```diff
@@ -3125,11 +3131,27 @@ export class Table extends Widget implements TableModel {
       }
       let $updatedRow = $(this._buildRowDiv(row));
       $updatedRow.copyCssClasses(oldRow.$row, Table.SELECTION_CLASSES + ' first last');
+
+      // Check if the cell tooltip is currently pointing to a cell in this row. If yes, update its $anchor to the corresponding cell
+      // in the new row. Otherwise, the tooltip position will be wrong, because the old anchor is no longer part of the DOM.
+      let $updatedTooltipCell: JQuery = null;
+      if (tooltipSupport && tooltipSupport.tooltip) {
```
fschinkel: maybe use ?-notation

eclipse-scout-core/src/table/Table.ts
```diff
@@ -3125,11 +3131,27 @@ export class Table extends Widget implements TableModel {
...
+        }
+      }
+
       oldRow.$row.replaceWith($updatedRow);
       Table.linkRowToDiv(row, $updatedRow);
       this._destroyTooltipsForRow(row);
       this._destroyCellEditorForRow(row);
       this._installRow(row);
+
+      if ($updatedTooltipCell && $updatedTooltipCell.length) {
```
fschinkel: maybe use ?-notation

eclipse-scout-core/src/tooltip/Tooltip.ts
```diff
@@ -286,6 +294,14 @@ export class Tooltip extends Widget implements TooltipModel {
     }
   }
 
+  set$Anchor($anchor) {
+    if ($anchor?.[0] === this.$anchor?.[0]) { // compare DOM elements
+      return;
+    }
+    this.$anchor = $anchor;
```
fschinkel: Is it necessary to position the tooltip if it is not rendered?
Maybe use classic setProperty-render-behaviour.

fschinkel (changes requested):

eclipse-scout-core/src/tooltip/TooltipSupport.ts
```diff
@@ -89,6 +89,10 @@ export class TooltipSupport {
     this._destroyTooltip();
   }
 
+  get tooltip() {
```
fschinkel: return type

eclipse-scout-core/src/tooltip/Tooltip.ts
```diff
@@ -232,6 +228,18 @@ export class Tooltip extends Widget implements TooltipModel {
     }
   }
 
+  setHtmlEnabled(htmlEnabled: boolean) {
+    this.setProperty('htmlEnabled', htmlEnabled);
+  }
+
+  _renderHtmlEnabled() {
```
fschinkel: protected

fschinkel (changes requested):

fschinkel (approved):

PR #959, opened 2024-03-20
Clipboard: fix copyText() in popup windows

bschwarzent:
Modern browsers offer a clipboard API on the 'navigator' object. The writeText() method will throw an error if the window is not in focus. It is therefore important to use the correct window object. Like the document, the window has to be retrieved from the given 'parent' widget.

378306


mvilliger (approved):
Looks good and works fine thank you.
Instead of 'options.parent && options.parent.rendered' you could also write 'options?.parent?.rendered' but doesn't matter :)

bschwarzent:
I copied the `options.parent ....` check from the "doc" case below. But you are right, this can be sortened by using the optional chaning operator. Checking for "rendered" is not even necessary, since the window()/document() methods will return null when the widget is not rendered. I adjusted the code accordingly.

PR #960, opened 2024-03-21
Session: don't reload page on upload errors

bschwarzent:
Errors during the uploading of files allow the application to continue. The default yes button action reloads the page, which is not necessary in this case. To skip the reload, the button action is remove. (This was already the case in 22.0, but somehow got lost during the TS migration.)


mvilliger (approved):
Works. Thanks for fixing.

PR #961, opened 2024-03-28
Chart: several improvements

fschinkel:
Add add(List<BigDecimal>) to INTupleChartValueGroupBean. Fix tooltip and datalabels for scatter charts using a label map.

328315


org.eclipse.scout.rt.chart.shared/src/main/java/org/eclipse/scout/rt/chart/shared/data/basic/chart/NTupleChartValueGroupBean.java
```diff
@@ -9,15 +9,18 @@
  */
 package org.eclipse.scout.rt.chart.shared.data.basic.chart;
 
+import static org.eclipse.scout.rt.platform.util.CollectionUtility.size;
```
bschwarzent: I would not use a static import for this method.

PR #964, opened 2024-04-03
POC for #379569

paolobazzi:


paolobazzi:
This migration option is not any longer necessary since pull request https://github.com/eclipse-scout/scout.rt/pull/974/ allows to migrate id type names.

PR #970, opened 2024-04-07
Features/cgu/24.2/hybridmanager improvements

cguglielmo:


eclipse-scout-core/src/desktop/hybrid/HybridManager.ts
```diff
@@ -132,30 +139,65 @@ export class HybridManager extends Widget {
   }
 
   /**
+   * @deprecated use call instead.
+   */
+  triggerActionType(actionType: string, data?: object): string {
```
fschinkel: Why was this method renamed and not just deprecated?

eclipse-scout-core/src/desktop/hybrid/HybridManager.ts
```diff
@@ -132,30 +139,65 @@ export class HybridManager extends Widget {
...
   }
 
-  async triggerHybridActionAndWait(eventType: string, data?: object): Promise<object> {
-    const id = this.triggerHybridAction(eventType, data);
-    const r = await this.when(`hybridActionEnd:${id}`);
-    return r.data;
+  /**
+   * @deprecated use callAndWait instead.
+   */
+  triggerActionTypeAndWait(actionType: string, data?: object): JQuery.Promise<object> {
```
fschinkel: Why was this method renamed and not just deprecated?

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/form/fields/JsonAdapterPropertyConfigBuilder.java
```diff
@@ -26,11 +26,16 @@ public JsonAdapterPropertyConfigBuilder() {
   }
 
   public JsonAdapterPropertyConfigBuilder global() {
-    m_global = true;
+    m_global = (model) -> true;
```
fschinkel:
```suggestion
    m_global = model -> true;
```

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/form/fields/JsonAdapterProperty.java
```diff
@@ -44,8 +44,11 @@ protected IUiSession getUiSession() {
     return m_uiSession;
   }
 
-  protected boolean isGlobal() {
-    return m_global;
+  /**
+   * @return true whether a global adapter should be created for the given model.
+   */
+  protected boolean isGlobal(Object model) {
```
fschinkel, line 54: Do we need to keep the no-arg-method? Projects might implement own `JsonAdapterProperty`s or override the method instead of using `createConfig()`.

fschinkel (changes requested):

PR #973, opened 2024-04-10
Features/cgu/24.2/various logical grid and tile grid improvements

cguglielmo:


eclipse-scout-core/src/layout/logicalgrid/GridData.ts
```diff
@@ -168,6 +168,13 @@ export class GridData {
     $.extend(this, model);
   }
 
+  /**
+   * @returns a clone of this grid data enriched with the properties of the given model.
+   */
+  clone?(model?: GridData) {
```
bschwarzent: - Declare the return type (GridData)

eclipse-scout-core/src/tile/TileGridLayout.ts
```diff
@@ -123,6 +123,9 @@ export class TileGridLayout extends LogicalGridLayout {
 
   protected _storeBounds(tiles: Tile[]) {
     tiles.forEach((tile, i) => {
+      if (!tile.animateBoundsChange) {
```
bschwarzent, line 126: Seems okay, although it's not entirely clear how not setting some data properties disables the animation. I believe it relies on 6732927d0eb3f563e4a9e6080f1412770d37b4a1, correct?

eclipse-scout-core/src/tile/TileGrid.ts
```diff
@@ -383,6 +383,9 @@ export class TileGrid<TTile extends Tile = Tile> extends Widget implements TileG
   }
 
   protected _renderInsertTiles(tiles: (TTile | PlaceholderTile)[]) {
+    if (!this.animateTileInsertion) {
```
bschwarzent, line 386: Remark about the commit message of ee1a72a6779a66cff6159bdeb1d1bd55b126a1a3:
- Last line has an unnecessary space at the start.
- Do I understand the text correctly, that the example code _did not_ work, but now _does_ with this change?

eclipse-scout-core/src/form/fields/FormField.ts
```diff
@@ -847,10 +847,7 @@ export class FormField extends Widget implements FormFieldModel {
   }
 
   protected _setGridDataHints(gridData: GridData) {
-    if (!gridData) {
-      gridData = new GridData();
-    }
-    this._setProperty('gridDataHints', GridData.ensure(gridData));
+    this._setProperty('gridDataHints', GridData.ensure(gridData || new GridData()));
```
bschwarzent, line 850: Why is this necessary: `|| new GridData()`
The method name "ensure" suggests that it automatically converts null or undefined arguments to a valid object (like arrays#ensure). Is that not the case?

eclipse-scout-core/src/layout/logicalgrid/GridData.ts
```diff
@@ -168,6 +168,13 @@ export class GridData {
     $.extend(this, model);
   }
 
+  equals?(other: GridData) {
```
bschwarzent: - Declare the return type (boolean)
- What is the significance of the ? after the method name? If an implementation is provided with the base class, why should the method be declared as optional?

eclipse-scout-core/src/layout/logicalgrid/GridData.ts
```diff
@@ -168,6 +168,13 @@ export class GridData {
...
+    if (!other || !(other instanceof GridData)) {
```
bschwarzent, line 59: First case is not necessary (undefined and null are also not instance of GridData)

eclipse-scout-core/src/tile/TileGridLayout.ts
```diff
@@ -217,7 +217,7 @@ export class TileGridLayout extends LogicalGridLayout {
 
     let bounds = graphics.cssBounds(tile.$container);
     let fromBounds = tile.$container.data('oldBounds');
-    if (tile instanceof PlaceholderTile && !tile.$container.data('was-layouted')) {
+    if (!tile.$container.data('was-layouted')) {
       // Placeholders may not have fromBounds because they are added while layouting
```
bschwarzent, line 203: Maybe adjust this comment  (it now seems to apply to all tiles, not only placeholder tiles)

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutConfig.ts
```diff
@@ -38,8 +38,8 @@ export interface LogicalGridLayoutConfigModel {
 /**
  * Configures layouting hints for elements layouted by {@link LogicalGridLayout}.
  *
- * The configured hints only have an effect if theirs value is >=0.
- * Otherwise, the default values specified by CSS are applied (see {@link LogicalGridLayout._initDefaults}).
+ * The configured hints only have an effect if their value is >=0.
```
bschwarzent: Maybe use same condition in documentation than in code? (`>=0` vs. `> -1`)

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutConfig.ts
```diff
@@ -49,29 +49,47 @@ export class LogicalGridLayoutConfig implements LogicalGridLayoutConfigModel {
   columnWidth: number;
   rowHeight: number;
   minWidth: number;
+  protected _defaults: InitModelOf<LogicalGridLayoutConfig>;
+  protected _options: InitModelOf<LogicalGridLayoutConfig>;
 
-  constructor(options?: InitModelOf<LogicalGridLayoutConfig>) {
+  constructor(options?: InitModelOf<LogicalGridLayoutConfig>, defaults?: InitModelOf<LogicalGridLayoutConfig>) {
```
bschwarzent, line 33: There are many ways to contruct this object:

1. constructor - options + defaults
2. clone() - defaults only
3. ensure() - options only

Isn't this a bit confusing? When should which we use which method? There should be at least a little bit of documentation. I suspect that only the ensure() method should be called from outside, the other two are only internal, is that correct?

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutConfig.ts
```diff
@@ -102,4 +120,28 @@ export class LogicalGridLayoutConfig implements LogicalGridLayoutConfigModel {
...
+      setter(getter()?.clone()); // Clone will read the new defaults and apply the custom options
+    };
+    HtmlEnvironment.get().on('propertyChange', handler);
+    widget.one('destroy', () => {
+      HtmlEnvironment.get().off('propertyChange', handler);
+    });
+  }
+
+  static prepareSmallHgapConfig(layoutConfig?: ObjectOrModel<LogicalGridLayoutConfig>) {
+    if (!layoutConfig) {
```
bschwarzent: Doesn't ensure() handle this case?

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutConfig.ts
```diff
@@ -102,4 +120,28 @@ export class LogicalGridLayoutConfig implements LogicalGridLayoutConfigModel {
...
+    let handler = () => {
+      setter(getter()?.clone()); // Clone will read the new defaults and apply the custom options
+    };
+    HtmlEnvironment.get().on('propertyChange', handler);
+    widget.one('destroy', () => {
+      HtmlEnvironment.get().off('propertyChange', handler);
+    });
+  }
+
+  static prepareSmallHgapConfig(layoutConfig?: ObjectOrModel<LogicalGridLayoutConfig>) {
```
bschwarzent: Declare return type

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutConfig.ts
```diff
@@ -102,4 +120,28 @@ export class LogicalGridLayoutConfig implements LogicalGridLayoutConfigModel {
     }
     return new LogicalGridLayoutConfig(layoutConfig);
   }
+
+  static initHtmlEnvChangeHandler(widget: Widget, getter: () => LogicalGridLayoutConfig, setter: (config: LogicalGridLayoutConfig) => void) {
```
bschwarzent, line 123: Maybe add a little documentation? The fact that this method must be called exactly once during the initialization of the widget is not clear without reading the code.

eclipse-scout-core/src/tile/TileGridLayoutConfig.ts
```diff
@@ -22,20 +22,42 @@ export interface TileGridLayoutConfigModel extends LogicalGridLayoutConfigModel
...
-  constructor(options?: InitModelOf<TileGridLayoutConfig>) {
-    super(options);
-    options = options || {};
-    if (options.maxWidth > -2) {
-      this.maxWidth = options.maxWidth;
+  protected static _DEFAULT_CONFIG: InitModelOf<TileGridLayoutConfig> = undefined;
+
+  constructor(options?: InitModelOf<TileGridLayoutConfig>, defaults?: InitModelOf<TileGridLayoutConfig>) {
+    super(options, defaults);
+    this._init(); // Needs to be initialized again otherwise maxWidth would be set to undefined
```
bschwarzent: Is that comment correct? $.extend() usually ignores properties that are undefined.

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutConfig.ts
bschwarzent: Two widgets call ensure() to create the config object, and two other widgets call prepareSmallHgapConfig(). I think it would be better understandable if those two ways were more similar. Some ideas:
1. Add a defaults argument to ensure(). Usage:
`this._setProperty('layoutConfig', LogicalGridLayoutConfig.ensure(layoutConfig, {hgap: HtmlEnvironment.get().smallColumnGap}));`
2. Add a second ensureX() method. Usage:
`this._setProperty('layoutConfig', LogicalGridLayoutConfig.ensureWithSmallHgap(layoutConfig));`
4. Add a fluent method. Usage:
`this._setProperty('layoutConfig', LogicalGridLayoutConfig.ensure(layoutConfig).withSmallHgapGap());`

eclipse-scout-core/src/form/fields/groupbox/GroupBox.ts
```diff
@@ -283,6 +283,7 @@ export class GroupBox extends CompositeField implements GroupBoxModel {
       bodyLayoutConfig = new LogicalGridLayoutConfig();
     }
     this._setProperty('bodyLayoutConfig', LogicalGridLayoutConfig.ensure(bodyLayoutConfig));
+    LogicalGridLayoutConfig.initHtmlEnvChangeHandler(this, () => this.bodyLayoutConfig, layoutConfig => this.setBodyLayoutConfig(layoutConfig));
```
bschwarzent, line 283: Remark about the commit message of e7f880c32249eae5f56823b88ee13d84f0d07e88:

> Default values are now available directly on the layout config object.
> This makes it easier to access the actual values for custom calculations.

What exactly does this mean? Is there an example for such a "custom calculation"? And why are the default values needed to access the "actual values"? Are these the widget-specific values or the final, merged values?

eclipse-scout-core/src/tile/Tile.ts
```diff
@@ -36,6 +36,8 @@ export class Tile extends Widget implements TileModel {
     this.selected = false;
     this.selectable = false;
     this.plainText = null;
+    // Null to let TileGrid decide whether to enable animation
```
bschwarzent, line 41: Remark about the commit message of 6732927d0eb3f563e4a9e6080f1412770d37b4a1:
* There is an unnecessary space in front of the last line.

eclipse-scout-core/src/tile/TileGridLayout.ts
cguglielmo: you are right, the flag should be checked in _animateTiles as well

eclipse-scout-core/src/form/fields/FormField.ts
cguglielmo: It ensures the passed object has the GridData type, unless it is null or undefined. We use the same pattern for dates.ensure, numbers.ensure, LookupCall.ensure, DateFormat.ensure. I think this is correct because otherwise we would lose the null state which may be important. For arrays we explicitly don't want the null state. But I agree it is a little bit confusing...

eclipse-scout-core/src/layout/logicalgrid/GridData.ts
cguglielmo: Until now, the GridData object contained no methods and could be used as model object. To prevent compile errors I added the ? modifier.

I agree, it is confusing. We should probably create a GridDataModel instead

PR #977, opened 2024-04-11
Features/cgu/24.2/form field tile

cguglielmo:


eclipse-scout-core/src/tile/fields/FormFieldTile-dark.less
```diff
@@ -1,21 +1,25 @@
 /*
- * Copyright (c) 2023 BSI Business Systems Integration AG.
- * All rights reserved. This program and the accompanying materials
- * are made available under the terms of the Eclipse Public License v1.0
- * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
+ * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
```
fschinkel, line 2: This seems incorrect, still 2023. Please also regenerate for all other files, e.g. `FormFieldTile.ts` has 2023 as well.

fschinkel (approved):

eclipse-scout-core/src/tile/fields/FormFieldTile-dark.less
cguglielmo: The commits were authored in 2023, so it is correct ;-)

PR #979, opened 2024-04-12
Features/bschwarzent/24.2/286465 disable navigatedownbutton

bschwarzent:


eclipse-scout-core/src/util/dates.ts
```diff
@@ -465,6 +465,43 @@ export const dates = {
     return dateFormat.format(date);
   },
 
+  /**
+   * Returns a formatted date in the pattern `yyyy-MM-dd hh:mm:ss.SSS`.
+   * The locale is irrelevant, the time is always returned in 24h format.
+   *
+   * @param options A {@link DatesFormatTimestampOptions} object or a date.
+   */
+  formatTimestamp(options?: DatesFormatTimestampOptions | Date): string {
```
cguglielmo: What is the difference to scout.dates.format(new Date(), scout.Locale.DEFAULT, 'yyyy-MM-dd hh:mm:ss.SSS')?
To omit the time or millisecons, you can just use an adjusted pattern, so it does not need a custom implementation. 
The name is also a bit confusing, because it does not reflect the difference to format(). Maybe formatNormalized, or basicFormat, simpleFormat or something like it.

eclipse-scout-core/src/desktop/outline/navigation/NavigateDownButton.ts
```diff
@@ -61,30 +67,32 @@ export class NavigateDownButton extends NavigateButton {
...
-    // when it's not a leaf and not a detail - the button is only enabled when a single row is selected and that row is linked to a page
-    let table = this.node.detailTable;
-    if (table) {
-      return table.selectedRows.length === 1 && !!table.selectedRows[0].page;
+    // If there is a detail table that defines the child pages, check if a row is selected.
+    if (this._hasDetailTable(false)) {
+      let selectedRows = this.node.detailTable.selectedRows;
+      // Drill down is only possible when a single row is selected and that row is linked to a page
+      if (selectedRows.length === 1 && selectedRows[0].page) {
+        return this.node.pageForTableRow(selectedRows[0]);
```
cguglielmo: Why don't you just use selectedRows\[0].page? Why do we even need this method? :)

eclipse-scout-core/src/desktop/outline/Outline.ts
```diff
@@ -159,14 +161,36 @@ export class Outline extends Tree implements DisplayParent, OutlineModel {
...
 
-  /**
-   * This function returns the outline mediator instance. When we're in an online Scout application we must
-   * return a null instance here, because mediation is done server-side.
-   */
   protected _createMediator(): OutlineMediator {
     return scout.create(OutlineMediator);
   }
 
+  protected _installLocalTreeListener() {
```
cguglielmo, line 168: How can the node not be a Page? I would skip the filter and just add a cast.

eclipse-scout-core/src/desktop/outline/Outline.ts
```diff
@@ -159,14 +161,36 @@ export class Outline extends Tree implements DisplayParent, OutlineModel {
...
+    // Note: Although there is a Java outline mediator in Scout Classic, we always install the JS outline mediator.
+    // This enables hybrid outlines, i.e. Java outlines that contain both Java and JS pages.
+
+    // Mediate node selection
+    this.on('nodesSelected', (event: TreeNodesSelectedEvent) => {
+      this.mediator.onPageSelected(this.selectedNode());
+    });
+
+    // Mediate subtree changes
+    let handleSubTreeChanged = (nodes: TreeNode | TreeNode[]) => {
```
cguglielmo, line 178: Can you explain in the commit message, what the purpose of this change is? Is this relevant for the navigate down button?

eclipse-scout-core/src/desktop/outline/navigation/NavigateDownButton.ts
bschwarzent: You are correct, this method is not necessary. I marked it as deprecated and replaced all usages. The property "page" on `TableRow` now has a short js doc.

eclipse-scout-core/src/desktop/outline/Outline.ts
bschwarzent: Done.

eclipse-scout-core/src/desktop/outline/Outline.ts
bschwarzent: A more detailed description was added.

eclipse-scout-core/src/util/dates.ts
bschwarzent: Aggred, it is too confusing. I reverted this changed. Instead, the `dates.format()` function can be used as suggested.

Things to note:
- `Locale.DEFAULT` cannot be used as second argument, because that constant is only a _LocaleModel_, not a _Locale_. I used `Locale.ensure()` instead.
- The pattern should use uppercase `HH` for the hours to make sure DateFormat uses the 24-hour format (0-23) rather than the 12-hour format (1-12).

PR #980, opened 2024-04-12
Table: add TableOrganizer and ShowInvisibleColumnsForm for Scout JS

bschwarzent:


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/organizer/ShowInvisibleColumnsForm.java
```diff
@@ -100,7 +89,7 @@ protected int getConfiguredGridColumnCount() {
     @Override
     protected void execInitField() {
       super.execInitField();
-      setStatusVisible(false);
+      //setStatusVisible(false);
```
mvilliger: Is this no longer required? Then remove the line? Otherwise remove comment?

eclipse-scout-core/src/table/TableModel.ts
```diff
@@ -268,4 +268,5 @@ export interface TableModel extends WidgetModel {
    */
   textFilterEnabled?: boolean;
   defaultMenuTypes?: string[];
+  tableOrganizer?: ObjectOrModel<TableOrganizer>;
```
mvilliger: if a tableOrganizer model is passed as part of this TableModel, is it actually used when creating the instance?
Either it should have an effect or here only a TableOrganizier can be passed and no model?

eclipse-scout-core/src/table/TableAdapter.ts
```diff
@@ -721,6 +721,15 @@ export class TableAdapter extends ModelAdapter {
       }
       return this.createTilesOrig(rows);
     }, true);
+
+    // not used in classic mode since table is organized on the server
+    objects.replacePrototypeFunction(Table, '_createTableOrganizer', function(this: Table & { _createTableOrganizerOrig }) {
+      if (this.modelAdapter) {
+        // nop in classic mode
+        return;
```
mvilliger: maybe return null here instead of undefined? But actually makes no difference

eclipse-scout-core/src/table/ShowInvisibleColumnsFormModel.ts
```diff
@@ -0,0 +1,97 @@
...
+              id: 'ColumnsTable',
+              objectType: Table,
+              checkable: true,
+              checkableStyle: Table.CheckableStyle.CHECKBOX_TABLE_ROW,
+              headerVisible: false,
+              autoResizeColumns: true,
+              columns: [
+                {
+                  id: 'KeyColumn',
+                  objectType: Column,
```
mvilliger: Specify the generic here (Column)? WidgetMap can handle generics now.

eclipse-scout-core/src/table/ShowInvisibleColumnsForm.ts
```diff
@@ -0,0 +1,80 @@
...
+  table: Table;
+  insertAfterColumn: Column;
+
+  constructor() {
+    super();
+    this.table = null;
+    this.insertAfterColumn = null;
+  }
+
+  protected override _jsonModel(): WidgetModel {
```
mvilliger: return type could be a FormModel

eclipse-scout-core/src/table/ShowInvisibleColumnsForm.ts
```diff
@@ -0,0 +1,80 @@
...
+      // If a "insertAfterColumn" is provided, move the (still hidden) columns to a position right after that column.
+      // We do this _before_ making them visible to prevent the animation. If no valid "insertAfterColumn" is provided,
+      // the selected columns are only made visible, but not moved.
+      // TODO BSH [js-table] Is it ok to directly manipulate the columns array on the table like this? Table#udpateColumnStructure is only useful for new columns
+      arrays.removeAll(this.table.columns, columnsToInsert);
+      arrays.insertAll(this.table.columns, columnsToInsert, this.table.columns.indexOf(this.insertAfterColumn) + 1);
+    }
+
+    // Make the columns visible (at their new location)
+    columnsToInsert.forEach(column => column.setVisible(true));
```
mvilliger: each setVisible triggers a 'columnStructureChanged' event. Maybe only one event should be fired after all columns have changed? Or is this intended, that this is executed multiple times?
At least Table#onColumnVisibilityChanged is executed several times which also executes _updateRowWidth() and redraw several times then? Maybe better to only execute this once after all columns have been updated?

eclipse-scout-core/src/table/ShowInvisibleColumnsForm.ts
```diff
@@ -0,0 +1,80 @@
...
+    });
+    this.widget('ColumnsTable').insertRows(rows);
+  }
+
+  protected _getAddableColumns(): Column[] {
+    // FIXME bsh [js-table] Find a better solution, maybe list of columns in form data?
+    if (this.table.tableOrganizer) {
+      return this.table.tableOrganizer.getAddableColumns(this.insertAfterColumn);
+    }
+    return this.table.displayableColumns()
```
mvilliger: Can this case actually happen? This Form is only called from the Organizer, right? Then it must be present? Or not?

eclipse-scout-core/src/table/ShowInvisibleColumnsForm.ts
```diff
@@ -0,0 +1,80 @@
...
+        cells: [
+          scout.create(Cell, {value: column}),
+          this._getColumnTitle(column)
+        ]
+      } as TableRow;
+    });
+    this.widget('ColumnsTable').insertRows(rows);
+  }
+
+  protected _getAddableColumns(): Column[] {
```
mvilliger: The list of available columns is computed by the tableOrganizer. Maybe instead of get them here, the organizer could pass it in the model for this form when creating it?

eclipse-scout-core/src/table/ShowInvisibleColumnsForm.ts
```diff
@@ -0,0 +1,80 @@
...
+  protected override _jsonModel(): WidgetModel {
+    return model();
+  }
+
+  override importData() {
+    let rows = this._getAddableColumns().map(column => {
+      return {
+        cells: [
+          scout.create(Cell, {value: column}),
+          this._getColumnTitle(column)
```
mvilliger: In Scout Classic if the tooltip is used as column title, it is displayed using italic font. Do we want this here as well? Or not?

eclipse-scout-core/src/table/TableOrganizer.ts
```diff
@@ -0,0 +1,175 @@
...
+      return false; // already removed
+    }
+    return this.table.visibleColumns().length > 1;
+  }
+
+  /**
+   * Returns true if the given column can be modified.
+   */
+  isColumnModifiable(column: Column): boolean {
+    if (this.table.isCustomizable()) {
```
mvilliger: directly return this.table.isCustomizable()

eclipse-scout-core/src/table/TableOrganizer.ts
```diff
@@ -0,0 +1,175 @@
...
+      table: this.table,
+      insertAfterColumn: insertAfterColumn
+    });
+    form.open();
+  }
+
+  protected _removeColumn(column: Column) {
+    column.setVisible(false);
+
+    // TODO bsh [js-table] Should we keep the sorting? In Java, it stays intact (only visible in the organize menu).
```
mvilliger: maybe keep it the same as in java? Or is this expected by the user if a column is removed that the rows change in order?

mvilliger:
Table grouping seems not to work correctly. E.g. when grouping by column A and removing column B, the whole grouping seems to be removed. This seems not to be correct?
Furthermore if grouping for multiple columns is configured and one of these is removed, the whole grouping is removed as well?

mvilliger:
Another issue:

1. Add an aggregation for a number column (e.g. avg).
2. Remove this number column -> the aggregation row is removed correctly
3. Add the aggregation column again -> aggregation row is not shown, but if the column header is clicked it still shows as active

eclipse-scout-core/src/table/ShowInvisibleColumnsForm.ts
bschwarzent: The table organizer is null in classic mode (method _createTableOrganizer is replaced in TableAdapter)

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/organizer/ShowInvisibleColumnsForm.java
bschwarzent: Yes, this is no longer required (table status is already invisible on the table field). Method was removed.

eclipse-scout-core/src/table/TableAdapter.ts
bschwarzent: The TableAdapter now sets the organizer to a RemoteTableOrganizer in the _initProperties method.

eclipse-scout-core/src/table/ShowInvisibleColumnsFormModel.ts
bschwarzent: Updated.

eclipse-scout-core/src/table/ShowInvisibleColumnsForm.ts
bschwarzent: Done.

eclipse-scout-core/src/table/ShowInvisibleColumnsForm.ts
bschwarzent: Code was changed to pass "false" as a second argument to the setVisible() method, and manually calls onColumnVisibilityChanged() once afterwards.

eclipse-scout-core/src/table/ShowInvisibleColumnsForm.ts
bschwarzent: No longer relevant, logic was moved to TableOrganizer.

eclipse-scout-core/src/table/ShowInvisibleColumnsForm.ts
bschwarzent: Not sure how useful this is, but I agree that it's better to have to two implementations to be as similar as possible. Italic font was added.

eclipse-scout-core/src/table/TableOrganizer.ts
bschwarzent: Agreed. Code was adjusted.

bschwarzent:
Thanks for the review. The aggregate rows are now updated correctly.

After additional feedback from @cguglielmo, the `tableOrganizer` property on the table was renamed to `organizer`. The static flags "columnAddable", "removable" and "modifiable" are now also considered in Scout JS. Their default value was change from false to true, but in Scout Classic, they are still defined by the model on the UI server. A special implementation `RemoteTableOrganizer` is used in classic mode.

eclipse-scout-core/src/table/TableModel.ts
bschwarzent: Type was changed to `TableOrganizer` only, because the organizer does not even have a model.

eclipse-scout-core/src/table/ShowInvisibleColumnsForm.ts
bschwarzent: Agreed. The form now receives the available columns and returns the selected columns via data property. The logic was moved to the table organizer.

eclipse-scout-core/src/table/TableOrganizer.ts
bschwarzent: I left this as it is, so all isColumnXy() methods are structured similarly.

PR #982, opened 2024-04-19
Page: remove destroyed detail form and table from page

bschwarzent:
A page may reference a detailForm or a detailTable. If the corresponding widget is destroyed, the member should also be set to null. Otherwise, it may happen that the destroyed widget is still accessed, creating the potential for unexpected errors ('Widget is destroyed').

For example, if a node page with both a detail form and a detail table is destroyed while still being selected, the DesktopBench will be triggered by a pageChanged event and tries to update the outline content with an already destroyed widget.

### DIFF
--- a/eclipse-scout-core/src/desktop/outline/pages/Page.ts
+++ b/eclipse-scout-core/src/desktop/outline/pages/Page.ts
@@ -137,9 +137,11 @@ export class Page extends TreeNode implements PageModel {
     super._destroy();
     if (this.detailTable) {
       this.detailTable.destroy();
+      this.detailTable = null;
     }
     if (this.detailForm) {
       this.detailForm.destroy();
+      this.detailForm = null;
     }
     this.trigger('destroy');
   }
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/src/desktop/outline/pages/Page.ts
```diff
@@ -137,9 +137,11 @@ export class Page extends TreeNode implements PageModel {
     super._destroy();
     if (this.detailTable) {
       this.detailTable.destroy();
+      this.detailTable = null;
```
cguglielmo, line 140: Ist this necessary? Shouldn't this be done by the destroy handler?

eclipse-scout-core/src/desktop/outline/pages/Page.ts
```diff
@@ -433,6 +435,11 @@ export class Page extends TreeNode implements PageModel {
     }
     this.detailForm = form;
     if (form) {
+      form.one('destroy', () => {
```
cguglielmo, line 438: Maybe remove the listener when the form changes, see FormMenu or FormTableControl

bschwarzent:
Change submitted on 25.1: https://github.com/eclipse-scout/scout.rt/commit/08be3bb1f4fdafce4f5c71d78e08f90a31cb7044

PR #984, opened 2024-04-25
Features/cgu/24.2/isd

cguglielmo:


cguglielmo:
jenkins run tests

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayout.ts
```diff
@@ -32,6 +34,8 @@ export class LogicalGridLayout extends AbstractLayout {
   columnWidth: number;
   rowHeight: number;
   minWidth: number;
+  spy: LogicalGridLayoutSpy;
```
fschinkel: protected?

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayout.ts
```diff
@@ -218,4 +224,14 @@ export class LogicalGridLayout extends AbstractLayout {
     dim.height += insets.vertical();
     return dim;
   }
+
+  setSpyEnabled(spyEnabled: boolean) {
+    this.spyEnabled = spyEnabled;
```
fschinkel, line 235: When this method is called multiple times with `true` there will be multiple spies that are not disposed.

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayout.ts
```diff
@@ -218,4 +224,14 @@ export class LogicalGridLayout extends AbstractLayout {
...
+    if (!this.spyEnabled) {
+      this.spy?.dispose();
+      this.spy = null;
+    } else {
+      this.spy = new LogicalGridLayoutSpy();
```
fschinkel: trigger layout in order to draw the spies cell bounds?

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
```diff
@@ -228,77 +230,67 @@ export class LogicalGridLayoutInfo implements LogicalGridLayoutInfoModel {
           prefw = this.logicalWidthInPixel(cons);
         }
         prefw = Math.floor(prefw);
-        for (j = cons.gridx; j < cons.gridx + cons.gridw && j < this.cols; j++) {
+        for (let j = cons.gridx; j < cons.gridx + cons.gridw && j < this.cols; j++) {
```
fschinkel: Please simplify this loop. As `cons.gridw` is `1` this loop is pointless and can be replaced by a simple if-statement.
```suggestion
        if (cons.gridx < this.cols) {
```

eclipse-scout-core/src/tile/TileGrid.ts
```diff
@@ -1266,7 +1266,7 @@ export class TileGrid<TTile extends Tile = Tile> extends Widget implements TileG
   findTileIndexAt(x: number, y: number, startIndex?: number, reverse?: boolean): number {
```
bschwarzent, line 1266: Test?

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
```diff
@@ -228,77 +230,67 @@ export class LogicalGridLayoutInfo implements LogicalGridLayoutInfoModel {
...
           if (!fixedWidths[j]) {
-            spanWidth += prefWidths[j];
+            spanWidth[lc.PREF] += prefWidths[j];
+            spanWidth[lc.MAX] += maxWidths[j];
           }
         }
+        let hGaps = (hSpan - 1) * this.hgap;
         if (cons.widthHint > 0) {
-          distWidth = cons.widthHint - spanWidth - (hSpan - 1) * this.hgap;
+          distWidth[lc.PREF] = cons.widthHint - spanWidth[lc.PREF] - hGaps;
```
fschinkel: Please simplify:
`distWidth[lc.PREF]` will always be `value - spanWidth[lc.PREF] - hGaps` and `distWidth[lc.MAX]` is `cons.maxWidth - spanWidth[lc.MAX] - hGaps`.
Then `distWidth[lc.PREF]` and `spanWidth[lc.PREF]` (or `distWidth[lc.MAX]` and `spanWidth[lc.MAX]`) are passed to `_distributeWidth` where the values are summed up again.
So it will be the same to just calculate `distWidth = value - hGaps` and pass only this value to the `_distributeWidth` method.

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
```diff
@@ -228,77 +230,67 @@ export class LogicalGridLayoutInfo implements LogicalGridLayoutInfoModel {
...
-              prefWidths[j] = Math.max(equalWidth, prefWidths[j]);
-            }
-            if (cons.weightx === 0) {
-              fixedWidths[j] = true;
-            }
-          }
-          if (last > -1) {
-            prefWidths[last] += remainder;
-          }
+        if (distWidth[lc.PREF] > 0) {
```
fschinkel: After simplifying the above, this needs to be: 
```suggestion
        if (distWidth[lc.PREF] > spanWidth[lc.PREF]) {
```

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
```diff
@@ -228,77 +230,67 @@ export class LogicalGridLayoutInfo implements LogicalGridLayoutInfoModel {
...
+          this._distributeWidth(cons, distWidth[lc.PREF], spanWidth[lc.PREF], prefWidths, fixedWidths, Math.max.bind(Math));
         }
+        distWidth[lc.MAX] = cons.maxWidth - spanWidth[lc.MAX] - hGaps;
```
fschinkel: I think this is incorrect. Every column has a maxWidth of 10240. So `distWidth[lc.MAX]` needs to be `Math.min(cons.gridw * 10240, cons.maxWidth) - spanWidth[lc.MAX]` (or just `Math.min(cons.gridw * 10240, cons.maxWidth)` after the above simplification).

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
```diff
@@ -308,26 +300,80 @@ export class LogicalGridLayoutInfo implements LogicalGridLayoutInfoModel {
...
-      for (i = 0; i < this.cols; i++) {
+      for (let i = 0; i < this.cols; i++) {
         this.weightX[i] = this.weightX[i] * f;
       }
     }
   }
 
+  /**
+   * @param cons the current grid data
+   * @param distWidth the widths to distribute to the columns. The widgets are distributed to the columns equally.
```
fschinkel:
```suggestion
   * @param distWidth the width to distribute to the columns. The width is distributed to the columns equally.
```

eclipse-scout-core/src/tile/TileGridSelectionHandler.ts
```diff
@@ -253,58 +282,52 @@ export class TileGridSelectionHandler {
...
 
-  protected _computeFocusedTile(diff: number): TileGridSelectionInstruction {
+  /**
+   * If no tiles are selected, an object is returned containing the tile that should be focused and selected.
+   * This is the first tile if diff is > 0 and the last tile otherwise.
+   *
+   * If there are tiles selected but there is no focused tile, an object is returned containing the tile to be focused.
+   * This is the last tile if diff is > 0 and the first tile otherwise.
+   */
+  protected _computeFocusedTileOrSelection(diff: number): TileGridSelectionInstruction {
```
bschwarzent, line 321: I find it a bit confusing that the first part of this method looks exactly like the implementation of computeFocusedTile(). Maybe this could be combined by passing the array of tiles as an argument to computeFocusedTile()?

Why does the method use the word "or", but the implementation returns both the focused tile "and" the selection? 

If all other computeXYZ() methods are public, shouldn't this method be public as well?

eclipse-scout-core/src/tile/TileGridSelectionHandler.ts
```diff
@@ -205,41 +228,47 @@ export class TileGridSelectionHandler {
   }
 
   computeSelectionX(xDiff: number, extend: boolean): TileGridSelectionInstruction {
-    let tiles = this.getVisibleTiles();
-    let focusedTile = null;
-    let focusedTileIndex = -1;
-    let result = this._computeFocusedTile(xDiff);
+    let result = this._computeFocusedTileOrSelection(xDiff);
     if (result.selectedTiles !== null) {
```
bschwarzent:
```suggestion
    if (result.selectedTiles) {
```

Or should we even need to check the length of the array? Maybe a JsDoc on the interface would help clarify the API.

Same with computeSelectionY.

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
```diff
@@ -308,26 +300,80 @@ export class LogicalGridLayoutInfo implements LogicalGridLayoutInfoModel {
...
+   * @param spanWidth the cumulated width of every cell of every row visited so far in the span range [grid.x, grid.x + grid.w]
+   * @param widths the column widths that have been distributed so far when the previous rows were visited
+   * @param fixedWidths the columns with a fixed width (weightx = 0)
+   * @param calc the function that decides whether to use the newly calculated column width or the existing one
```
fschinkel: jsdoc indicates that you can only choose between two values which is not true.

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
```diff
@@ -308,26 +300,80 @@ export class LogicalGridLayoutInfo implements LogicalGridLayoutInfoModel {
...
+   */
+  protected _distributeWidth<T>(cons: LogicalGridData, distWidth: number, spanWidth: number, widths: number[], fixedWidths: boolean[], calc: (equalWidth: number, width: number) => number) {
```
fschinkel: The method will do ok for `weightx !== 0`. But it will definitely do something strange for `weightx === 0` if there are some columns that are already fixed.
Consider the following 2 rows of fields:
 - 3 fields, all with `gridw = 1`, 2 of them have `weightx = 0` and are `250` wide, the third one has `weightx = -1`
 - 1 field with `gridw = 3`, `weightx = 0` and `widthHint = 1500`

Two columns are fixed because of the 1x1-fields, the 1x3-field will now define the width of the third column and fixate it to 480 ((1500 - 2 * 30) / 3). This makes the field 1040 (480 + 2 * 30 + 2 * 250) wide.
So the field has the possibility to calculate a width that fulfills its own needs but decides to not use this possibility but it calculates a "wrong" width and forces all following fields to stick to this width.

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
```diff
@@ -308,26 +300,80 @@ export class LogicalGridLayoutInfo implements LogicalGridLayoutInfoModel {
...
+      }
+    }
+    if (last > -1) {
+      widths[last] += remainder;
+    }
+  }
+
+  /**
+   * @param cons the current grid data
+   * @param distHeight the heights to distribute to the rows. The widgets are distributed to the rows equally.
```
fschinkel:
```suggestion
   * @param distHeight the height to distribute to the rows. The height is distributed to the rows equally.
```

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
```diff
@@ -308,26 +300,80 @@ export class LogicalGridLayoutInfo implements LogicalGridLayoutInfoModel {
...
+      for (let i = 0; i < this.cols; i++) {
         this.weightX[i] = this.weightX[i] * f;
       }
     }
   }
 
+  /**
+   * @param cons the current grid data
+   * @param distWidth the widths to distribute to the columns. The widgets are distributed to the columns equally.
+   * @param spanWidth the cumulated width of every cell of every row visited so far in the span range [grid.x, grid.x + grid.w]
```
fschinkel: remove parameter

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
fschinkel: remove generic

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
```diff
@@ -308,26 +300,80 @@ export class LogicalGridLayoutInfo implements LogicalGridLayoutInfoModel {
...
+    }
+  }
+
+  /**
+   * @param cons the current grid data
+   * @param distHeight the heights to distribute to the rows. The widgets are distributed to the rows equally.
+   * @param spanHeight the cumulated height of every cell of every column visited so far in the span range [grid.y, grid.y + grid.h]
+   * @param widths the row heights that have been distributed so far when the previous column were visited
+   * @param fixedHeights the rows with a fixed height (weighty = 0)
+   */
```
fschinkel, line 345: jsdoc for calc is missing

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
```diff
@@ -308,26 +300,80 @@ export class LogicalGridLayoutInfo implements LogicalGridLayoutInfoModel {
...
+  protected _distributeHeight<T>(cons: LogicalGridData, distHeight: number, spanHeight: number, heights: number[], fixedHeights: boolean[], calc: (equalWidth: number, width: number) => number) {
```
fschinkel: remove generic

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
```diff
@@ -336,76 +382,66 @@ export class LogicalGridLayoutInfo implements LogicalGridLayoutInfoModel {
           prefh = this.logicalHeightInPixel(cons);
         }
         prefh = Math.floor(prefh);
-        for (j = cons.gridy; j < cons.gridy + cons.gridh && j < this.rows; j++) {
+        for (let j = cons.gridy; j < cons.gridy + cons.gridh && j < this.rows; j++) {
```
fschinkel: replace with if (see `_initializeColumns`)

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
```diff
@@ -308,26 +300,80 @@ export class LogicalGridLayoutInfo implements LogicalGridLayoutInfoModel {
...
+      if (cons.weighty === 0) {
+        fixedHeights[j] = true;
+      }
+    }
+    if (last > -1) {
+      heights[last] += remainder;
+    }
+  }
+
   protected _initializeRows() {
```
fschinkel, line 365: This method does exactly the same as `_initializeColumns`. Can we write this in a more generic way, to use the same code instead of duplicating it.

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
```diff
@@ -336,76 +382,66 @@ export class LogicalGridLayoutInfo implements LogicalGridLayoutInfoModel {
...
           if (!fixedHeights[j]) {
-            spanHeight += prefHeights[j];
+            spanHeight[lc.PREF] += prefHeights[j];
+            spanHeight[lc.MAX] += maxHeights[j];
           }
         }
+        let vGaps = (vSpan - 1) * this.vgap;
         if (cons.heightHint > 0) {
-          distHeight = cons.heightHint - spanHeight - (vSpan - 1) * this.vgap;
+          distHeight[lc.PREF] = cons.heightHint - spanHeight[lc.PREF] - vGaps;
```
fschinkel: simplify (see `_initializeColumns`)

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
```diff
@@ -336,76 +382,66 @@ export class LogicalGridLayoutInfo implements LogicalGridLayoutInfoModel {
...
-              prefHeights[j] = Math.max(equalHeight, prefHeights[j]);
-            }
-            if (cons.weighty === 0) {
-              fixedHeights[j] = true;
-            }
-          }
-          if (last > -1) {
-            prefHeights[last] += remainder;
-          }
+        if (distHeight[lc.PREF] > 0) {
```
fschinkel: simplify (see `_initializeColumns`)

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
```diff
@@ -336,76 +382,66 @@ export class LogicalGridLayoutInfo implements LogicalGridLayoutInfoModel {
...
+          this._distributeHeight(cons, distHeight[lc.PREF], spanHeight[lc.PREF], prefHeights, fixedHeights, Math.max.bind(this));
         }
+        distHeight[lc.MAX] = cons.maxHeight - spanHeight[lc.MAX] - vGaps;
```
fschinkel: incorrect (see `_initializeColumns`)

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
fschinkel: same problems as described for `_distributeWidth`

eclipse-scout-core/test/layout/RectangleSpec.ts
```diff
@@ -29,6 +29,17 @@ describe('Rectangle', () => {
...
   });
 
+  describe('union', () => {
+    let r1 = new Rectangle(0, 0, 675, 558);
+    let r2 = new Rectangle(687, 0, 674, 558);
+
+    it('produces same results as java.awt.Rectangle', () => {
+      let r = r1.union(r2);
+      let expected = new Rectangle(0, 0, 1361, 558);
+      expect(expected.equals(r)).toBe(true);
```
fschinkel:
```suggestion
      expect(r).toEqual(new Rectangle(0, 0, 1361, 558));
```

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/tile/AbstractTile.java
```diff
@@ -57,6 +59,7 @@ public abstract class AbstractTile extends AbstractWidget implements ITile {
   private IDataChangeListener m_internalDataChangeListener;
   private boolean m_filterAccepted = true;
   private volatile boolean m_loaded = false;
+  private IFuture<Void> m_loadJob;
```
bschwarzent: All other private variables have at least a protected getter/setter to allow subclasses to use them. For consistency reasons, this should be done for this future as well.

Also, check the naming. Most places call these "future", not "job". (Technically, the job is the runnable that is passed to the job manager, the result is a promise-like object representing the jobs state.)

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/tile/AbstractTile.java
```diff
@@ -386,6 +389,32 @@ public void loadData() {
     }
   }
 
+  @Override
+  public void cancelLoading() {
+    if (m_loadJob != null) {
+      m_loadJob.cancel(true);
```
bschwarzent: Maybe set the member to null? Or add .whenDone() when scheduling the job and set it to null, so it will also be reset upon normal completion.

eclipse-scout-core/test/layout/logicalgrid/LogicalGridLayoutInfoSpec.ts
```diff
@@ -1,102 +1,322 @@
...
+  function expectWidths(rows: Rectangle[][], rowNum: number, widths: number[], height: number, hgap: number, vgap: number) {
+    let row = rows[rowNum];
+    expect(row.length).toBe(widths.length);
 
-    it('produces same results as java.awt.Rectangle', () => {
-      let r = r1.union(r2);
-      let expected = new Rectangle(0, 0, 1361, 558);
-      expect(expected.equals(r)).toBe(true);
-    });
+    let x = 0;
```
fschinkel, line 36: add `y` here

eclipse-scout-core/test/layout/logicalgrid/LogicalGridLayoutInfoSpec.ts
```diff
@@ -1,102 +1,322 @@
...
-        },
-        attr: attrKey => attrKey === 'id' ? compName : undefined
-      };
-    };
-
-    function mockHtmlComp(jquery) {
-      return {
-        prefSize: () => new Dimension(1, 1)
-      };
+    let y = 0;
```
fschinkel, line 50: add `x` here

eclipse-scout-core/test/layout/logicalgrid/LogicalGridLayoutInfoSpec.ts
```diff
@@ -1,102 +1,322 @@
...
  * which is available at https://www.eclipse.org/legal/epl-2.0/
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-import {Dimension, Insets, LogicalGridData, LogicalGridLayoutInfo, Rectangle} from '../../../src/index';
+import {Dimension, InitModelOf, Insets, LogicalGridData, LogicalGridLayoutInfo, Rectangle} from '../../../src/index';
 
-/* This test also exists as Java code, to make sure Java and JS code produces the same results */
 describe('LogicalGridLayoutInfo', () => {
+  function $dummyComps(count: number) {
```
fschinkel: return type missing

eclipse-scout-core/test/layout/logicalgrid/LogicalGridLayoutInfoSpec.ts
```diff
@@ -1,102 +1,322 @@
...
-    let cons = [gd1, gd2];
-    let lgli = new LogicalGridLayoutInfo({
-      $components: components,
-      cons: cons,
-      hgap: 5,
-      vgap: 5,
-      rowHeight: 30
-    });
-    let parentSize = new Dimension(500, 23);
+  function toCols(rows: Rectangle[][]) {
```
fschinkel: return type missing

eclipse-scout-core/src/style/sizes.less
```diff
@@ -390,4 +390,5 @@
 /* -------------------------------- */
 @iphone5-height: 568px;
 @iphone5-width: 320px;
-@iphone6-width: 375px;
+@iphone6-width: 375px; // includes iphone 6, 6s, 6s plus, 7, 8
```
mvilliger, line 394: Looks good but the comment of this commit has a typo: sizes.lesS

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/tile/AbstractTile.java
```diff
@@ -386,6 +389,32 @@ public void loadData() {
...
+  public void reloadData() {
+    if (!isLoading()) {
+      loadData();
+      return;
+    }
+    cancelLoading();
+    addPropertyChangeListener(new PropertyChangeListener() {
+      @Override
+      public void propertyChange(PropertyChangeEvent evt) {
+        if (ITile.PROP_LOADING.equals(evt.getPropertyName()) && !isLoading()) {
```
bschwarzent: 1. Filter for property name can be passed as first argument to addPropertyChangeListener().
2. Is the additional check for isLoading() necessary? We know the property was true when we added the listener, so it should only be triggered by a change from true to false. Or did you explicitly want to cover the case where someone fired a property change event without actual changing the property?

eclipse-scout-core/test/layout/logicalgrid/LogicalGridLayoutInfoSpec.ts
fschinkel: please add jsdoc or comment saying that this simply transposes the given matrix

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/tile/ITile.java
```diff
@@ -56,8 +57,21 @@ public interface ITile extends IWidget, IOrdered, IStyleable, IExtensibleObject,
 
   void ensureDataLoaded();
 
+  /**
+   * Loads the data using {@link ITileDataLoader}. Does nothing, if the data is already being loaded.
```
bschwarzent:
```suggestion
   * Loads the data using {@link ITileDataLoader}. Does nothing if the data is already being loaded.
```

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/tile/ITileGrid.java
```diff
@@ -46,7 +46,7 @@ public interface ITileGrid<T extends ITile> extends IWidget, IExtensibleObject,
   String PROP_ASYNC_LOAD_JOBNAME_PREFIX = "tileAsyncDataLoadJob";
```
bschwarzent: It's very confusing that these constants start with PROP_ like ordinary properties. But they seem only to be referenced in execution hints. Maybe rename them to EXECUTION_HINT_ prefix? (And yes, I know that it was me who originally added these constants... 🙈)

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/ui/UserAgentUtility.java
```diff
@@ -56,4 +56,25 @@ public static UserAgent getCurrentUserAgent() {
     }
   }
 
+  /**
```
mvilliger: Is this method really necessary? It would also be possible to use the ClientRunContext instead?

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
```diff
@@ -308,26 +300,80 @@ export class LogicalGridLayoutInfo implements LogicalGridLayoutInfoModel {
...
+   * @param cons the current grid data
+   * @param distWidth the widths to distribute to the columns. The widgets are distributed to the columns equally.
+   * @param spanWidth the cumulated width of every cell of every row visited so far in the span range [grid.x, grid.x + grid.w]
+   * @param widths the column widths that have been distributed so far when the previous rows were visited
+   * @param fixedWidths the columns with a fixed width (weightx = 0)
+   * @param calc the function that decides whether to use the newly calculated column width or the existing one
+   */
+  protected _distributeWidth<T>(cons: LogicalGridData, distWidth: number, spanWidth: number, widths: number[], fixedWidths: boolean[], calc: (equalWidth: number, width: number) => number) {
+    let hSpan = cons.gridw;
+    let equalWidth = Math.floor((distWidth + spanWidth) / hSpan);
```
fschinkel: This is not a proper way to distribute equally.
Consider the 99 distributed to 10 columns.
This will lead to 9 columns being 9 wide and the last one being 18 wide.

See pie charts or #222794 for a proper algorithm (ticket has a proof attached).

eclipse-scout-core/test/layout/logicalgrid/LogicalGridLayoutInfoSpec.ts
```diff
@@ -1,102 +1,322 @@
...
-  });
+    it('considers widthHint', () => {
+      let gd1 = new LogicalGridData({gridx: 0, gridy: 0, gridw: 1, gridh: 1, weightx: 0, widthHint: 70});
+      let gd2 = new LogicalGridData({gridx: 1, gridy: 0, gridw: 1, gridh: 1, weightx: 1});
+      let rows = newLogicalGridLayoutInfo([gd1, gd2]).layoutCellBounds(parentSize, parentInsets);
+      expect(rows[0][0]).toEqual(new Rectangle(0, 0, 70, 30));
+      expect(rows[0][1]).toEqual(new Rectangle(75, 0, parentSize.width - 75, 30));
+    });
+
+    describe('maxWidth', () => {
```
fschinkel, line 104: add test cases for the problems pointed out in `LogicalGridLayoutInfo`

eclipse-scout-core/src/tile/fields/FormFieldTile.less
```diff
@@ -13,10 +13,15 @@
   color: @tile-default-color;
   --tile-background-color: @tile-default-background-color;
   background-color: var(--tile-background-color);
-  overflow: hidden;
-
   #scout.double-drop-shadow(@spread1: 1px, @color1: @dashboard-tile-border-color, @y2: 0px, @blur2: 0px);
 
+  // Always use inverted colors which are darker than regular color to make the resizable visible even if the dashboard contains mixed tiles (some inverted, some not)
```
bschwarzent, line 19: Maybe mention that this variable is used by the "Resizable" helper class. Initially I was confused by the name, because I don't think colors are resizable.

org.eclipse.scout.rt.chart.shared/src/main/java/org/eclipse/scout/rt/chart/shared/data/basic/chart/IChartConfig.java
```diff
@@ -58,6 +58,7 @@ public interface IChartConfig extends Serializable {
 
   String BUBBLE_SIZE_OF_LARGEST_BUBBLE = "options.bubble.sizeOfLargestBubble";
   String BUBBLE_MIN_BUBBLE_SIZE = "options.bubble.minBubbleSize";
+  String LEGEND_LABELS_BOX_WIDTH = "options.plugins.legend.labels.boxWidth";
```
fschinkel: add before `SPEEDO_CHART_GREEN_AREA_POSITION` and separate with an empty line, so we have standard properties (chartjs) first and then the custom ones

eclipse-scout-core/src/tile/fields/FormFieldTile.less
```diff
@@ -13,10 +13,15 @@
...
+  --resizable-color: @dashboard-tile-default-inverted-selected-border-color;
```
bschwarzent, line 20: This commit could be merged with the previous one ("FormFieldTile: use the tile selection color for the resizable overlay")

eclipse-scout-core/src/resizable/Resizable.ts
```diff
@@ -223,17 +215,20 @@ export class Resizable implements ResizableModel, ObjectWithType {
...
-      maxHeight = $resizable.cssMaxHeight(),
-      initialBounds = graphics.bounds($resizable, {exact: true});
+    let $resizable = this.$container;
+    let $myWindow = $resizable.window();
+    let $handle = $(event.target);
+    let minWidth = $resizable.cssMinWidth();
+    let minHeight = $resizable.cssMinHeight();
+    let maxWidth = $resizable.cssMaxWidth();
+    let maxHeight = $resizable.cssMaxHeight();
+    let initialBounds = graphics.bounds($resizable, {exact: true});
```
bschwarzent: Why not move "let $offsetParent = ..." up one line and merge the two "initialBounds" lines?

eclipse-scout-core/src/focus/FocusManager.ts
```diff
@@ -413,38 +413,46 @@ export class FocusManager implements FocusManagerOptions {
...
 
-    // 6. Allow focus gain on elements with a focusable parent, e.g. when clicking on a row in a table.
+    // Allow focus gain on elements with a focusable parent, e.g. when clicking on a row in a table.
     if (focusUtils.containsParentFocusableByMouse($element, $element.entryPoint())) {
       return true;
     }
 
+    // Don't prevent default action for draggable elements which is dragstart event
+    if (focusUtils.isDraggable($element)) {
+      // Unfortunately, preventDefault will not only prevent dragstart but also focus gain
```
bschwarzent: I don't really understand this comment. We return true here, so preventDefault() will not be called. Or do you mean that by _not_ calling preventDefault(), we allow the focus to be changed (which would normally be suppressed by preventDefault)?

But shouldn't the draggable element always be focusable here? There are two previous checks (isFocusableByMouse and is(':focusable')) that should have exited the method already.

eclipse-scout-core/src/focus/focusUtils.ts
```diff
@@ -94,5 +102,22 @@ export const focusUtils = {
       activeElement = ownerDocument.activeElement;
     }
     return activeElement === element;
+  },
+
+  /**
+   * Stores the currently focused element and focuses this element again in the next animation frame if the focus changed to the entry point element.
+   * This is useful if the current task would focus the entry point element which cannot be prevented.
+   */
+  restoreFocusLater($entryPoint: JQuery) {
```
bschwarzent, line 111: This seems to be very hacky. Is there no better solution?

eclipse-scout-core/src/testing/JasmineScoutUtil.ts
```diff
@@ -101,11 +101,32 @@ export const JasmineScoutUtil = {
...
       fail('Ajax call not mocked for url: ' + request.url + ', method: ' + request.method);
     });
+  },
+
+  /**
+   * Calls the given mock as soon as a hybrid action with the given actionType is called.
+   * The mock is called asynchronously using setTimeout to let the runtime code add any required event listeners first.
+   *
+   * The mock may return an object with [id, widget] if the action is supposed to create widgets.
+   * The id contains the action id and the id for the widget used in the hybrid action (`${actionId}${widgetId}`).
```
fschinkel: What do you want to say with the last sentence? "needs to contain"? I think this is not generally true.

eclipse-scout-chart/src/chart/Chart.less
```diff
@@ -391,11 +391,15 @@
 @chart-min-width: 75px;
 
 /* use padding area for label, legend */
```
bschwarzent: I think this comments belongs to the .field.chart, not to the .form-field -> move inside

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/hybrid/AbstractFormHybridAction.java
```diff
@@ -32,34 +43,34 @@ protected boolean isShowFormOnStart() {
...
       if (FormEvent.TYPE_STORE_AFTER == e.getType()) {
-        DO_ENTITY result = createEmptyResult();
-        exportResult(form, result);
+        DO_ENTITY result = exportResult(form);
+        if (result == null) {
+          result = createEmptyResult();
+        }
         fireHybridWidgetEvent("save", result);
       }
       else if (FormEvent.TYPE_RESET_COMPLETE == e.getType()) {
```
fschinkel, line 64: Result needs to be exported and sent to the UI in the reset case as well.
See `HybridManager._onHybridFormEvent`. The sent `data` will be set onto the Form and then a reset-Event is triggered. Therefore, the current data (after reset) needs to be sent otherwise it will be empty in the UI after a reset.

eclipse-scout-core/src/util/MoveSupport.less
```diff
@@ -0,0 +1,88 @@
...
+        #scout.drop-shadow-large();
+        opacity: 0;
+        pointer-events: none;
+        border-radius: inherit;
+      }
+
+      & > .resizable-handle {
+        // Disable interaction with handles on the clone while dragging
+        // Cannot be deactivated on the clone itself because it would enable the interaction with elements behind the clone
+        pointer-events: none;
```
bschwarzent, line 76: Doesn't this potentially affect _all_ elements in the clone container (not only resizable handles)? Maybe we can add a general rule (using the * selector), so we would not have to add additional rules on all draggable widgets (e.g. FormFieldTile.less).

fschinkel (changes requested):

eclipse-scout-core/src/util/MoveSupport.less
```diff
@@ -0,0 +1,88 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+#scout {
+
+  .move-support() {
+
+    &.dragged {
+      z-index: -1; // draw behind all other elements (during animation)
+      border: 0 !important; // remove border (if there is one) while dragging so the overlay covers the whole element
```
bschwarzent, line 18: Is this only needed for the tiles? If so, maybe add it to the Tile.less file? There are other properties which also might overlap the overlay, and they are not handled here (e.g. box-shadow).

eclipse-scout-core/src/util/MoveSupport.less
```diff
@@ -0,0 +1,88 @@
...
+        right: 0;
+        bottom: 0;
+        border-radius: inherit;
+        background-color: @move-support-drag-placeholder-background-color;
+        border: 2px dashed @move-support-drag-placeholder-border-color;
+        z-index: 2; // draw over elements inside the tile having a z-index
+      }
+
+      & .scroll-shadow {
+        // Scroll shadow would be drawn over after element -> hide it
```
bschwarzent: Really? ::after element has z-index 2, .scroll-shadow has z-index 1, so it  should drawn above the shadow.

eclipse-scout-core/src/util/MoveSupport.less
```diff
@@ -0,0 +1,88 @@
...
+        --scroll-shadow-size: transparent !important;
+        transition: none;
+      }
+    }
+
+    // done class is added when dragging is finished and original state restored right before dragged and dragged-end classes are removed
+    // the class will be removed about 50ms later and is used to disable the opacity transition. This ensures the dragged element does not flicker when shown again.
+    &.drag-done {
```
bschwarzent: Is this only needed if the draggable element has a transition? If yes, this should be mentioned in the comment.

eclipse-scout-core/src/util/MoveSupport.less
```diff
@@ -0,0 +1,88 @@
...
+      transition: none !important;
+    }
+
+    &.clone.dragging {
```
bschwarzent: Is the ".clone" class needed here, or only added for clarification? I believe the original element gets the "dragged" class, while the clone gets the "dragging" class. If that is correct, add a short comment to the respective css rules.

In general, we should probably have same big comment at the top of this file explaining the different CSS classes. After reviewing the other files, I find it hard to tell which class is for what without looking at the implementation.

eclipse-scout-core/src/util/MoveSupport.less
```diff
@@ -0,0 +1,88 @@
...
+
+    &.dragged {
+      z-index: -1; // draw behind all other elements (during animation)
+      border: 0 !important; // remove border (if there is one) while dragging so the overlay covers the whole element
+
+      &.releasing {
+        opacity: 0;
+        transition: opacity 500ms;
+
+        &.moved {
```
bschwarzent: Who manages the .moved class? I could not find it in MoveSupport.ts. Does it work similarly to the .drag-done class? Why not use the same class?

eclipse-scout-core/src/util/MoveSupport.less
```diff
@@ -0,0 +1,88 @@
...
+
+      & .scroll-shadow {
+        // Scroll shadow would be drawn over after element -> hide it
+        --scroll-shadow-size: transparent !important;
+        transition: none;
+      }
+    }
+
+    // done class is added when dragging is finished and original state restored right before dragged and dragged-end classes are removed
+    // the class will be removed about 50ms later and is used to disable the opacity transition. This ensures the dragged element does not flicker when shown again.
```
bschwarzent: This comment mentions 50ms, but in the code it's 100ms.

eclipse-scout-core/src/util/MoveSupport.ts
```diff
@@ -0,0 +1,468 @@
...
+    // Remove clone
+    this._moveData.$clone && this._moveData.$clone.remove();
+
+    // A done class makes it possible to disable transitions that must not be active while the clone will be swapped with the dragged element
+    let $draggedElement = this._moveData.$draggedElement;
+    $draggedElement.addClass('drag-done');
+    setTimeout(() => {
+      $draggedElement.removeClass('drag-done');
+    }, 100);
+    this._moveData.$draggedElement.removeClass('dragged releasing');
```
bschwarzent, line 175: Could this be merged with the line "$draggedElement.addClass('drag-done')"? At least use the same local $draggedElement variable.

eclipse-scout-core/src/util/arrays.ts
```diff
@@ -609,5 +609,12 @@ export const arrays = {
       return array.splice(0, array.length);
     }
     return [];
+  },
+
+  swap<T>(array: T[], element1: T, element2: T) {
```
bschwarzent, line 677: Is this really a useful function? If yes, add documentation and tests. We should probably also properly handle null cases etc. Currently, the method is only called in TileGridMoveSupport, so maybe we could just inline it there.

eclipse-scout-core/src/tile/Tile.ts
```diff
@@ -25,6 +25,8 @@ export class Tile extends Widget implements TileModel {
   rowId: string;
   selected: boolean;
   selectable: boolean;
+  movable: boolean;
+  movableProducer: () => TileMoveHandler;
```
bschwarzent, line 29: Is there maybe a better name for this? I was initially a bit confused of the name "movableProducer", which does not seem to have anything to do with the "movable" property. Maybe "moveHandlerProducer"?

eclipse-scout-core/src/tile/TileMoveHandler.ts
```diff
@@ -0,0 +1,61 @@
...
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {InitModelOf, ObjectModel, ObjectWithType, scout, SomeRequired, Tile, TileGrid, TileGridMoveSupport} from '..';
+import $ from 'jquery';
+
+export class TileMoveHandler implements TileMovableModel, ObjectWithType {
```
bschwarzent: Why is the model not called TileMoveHandlerModel?

eclipse-scout-core/src/tile/TileMoveHandler.ts
```diff
@@ -0,0 +1,61 @@
...
+  declare model: TileMovableModel;
+  declare initModel: SomeRequired<this['model'], '$container'>;
+  objectType: string;
+  tileGrid: TileGrid;
+  moveSupport: TileGridMoveSupport;
```
bschwarzent: Move this member down below the model properties.
It also seems a bit useless, because it only used in a spec. Maybe at least call cancel() in the destroy() method?

eclipse-scout-core/src/tile/TileGridMoveSupport.ts
```diff
@@ -0,0 +1,105 @@
...
+    let draggedTileBounds = objects.extractProperties(draggedTile.gridDataHints, {}, ['x', 'y', 'w', 'h']);
+    let targetTile: Tile = tileBelowCursor;
+    let targetTileBounds = objects.extractProperties(targetTile.gridDataHints, {}, ['x', 'y', 'w', 'h']);
+    arrays.swap(newElements, targetTile, draggedTile);
+    targetTile.setGridDataHints(targetTile.gridDataHints.clone(draggedTileBounds));
+    draggedTile.setGridDataHints(draggedTile.gridDataHints.clone(targetTileBounds));
+    this.widget.setTiles(newElements);
+
+    // Update element infos right after layout is done but BEFORE animation starts to get the final position of the tiles
+    let def = $.Deferred();
```
bschwarzent: The majority of the Deferred variables in our code base are called "deferred", not "def".

eclipse-scout-core/src/tile/TileGrid.ts
```diff
@@ -61,6 +61,7 @@ export class TileGrid<TTile extends Tile = Tile> extends Widget implements TileG
   updateTextFilterText: string;
   defaultMenuTypes: string[];
   wrappable: boolean;
+  movableProducer: (tile: Tile) => TileMoveHandler;
```
bschwarzent, line 65: I find it a little bit confusing that both TileGrid and Tile have a member called "movableProducer" but with different types. This here seems to be a "movableProducerProducer"! :-)

eclipse-scout-core/src/tile/tileUtil.ts
```diff
@@ -0,0 +1,168 @@
...
+      x: tiles.length > 0 ? Number.MAX_SAFE_INTEGER : 0,
+      y: tiles.length > 0 ? Number.MAX_SAFE_INTEGER : 0,
+      width: minWidth,
+      height: minHeight
+    });
+    for (let tile of tiles) {
+      // If tiles use explicit x/y values work with hints because they are always set. GridData will only be set after layouting.
+      let gridData = tile.gridDataHints;
+      if (gridData.x < 0 || gridData.y < 0) {
+        // // Use gridData object instead if available
```
bschwarzent: Unnecessary slashes

eclipse-scout-core/src/tile/TileResizeHandler.ts
```diff
@@ -0,0 +1,210 @@
...
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {arrays, graphics, GridData, InitModelOf, LogicalGridLayout, Point, Predicate, Rectangle, Resizable, ResizableModel, scout, Tile, TileGrid, tileUtil} from '..';
+
+export class TileResizeHandler extends Resizable implements TileResizableModel {
```
bschwarzent: Again, I'm confused by the naming. TileResizeModel -> TileResizeHandlerModel?

eclipse-scout-core/src/tile/TileResizeHandler.ts
```diff
@@ -0,0 +1,210 @@
...
+      cell.width = initialCell.width;
+    }
+    return initialCell.union(cell);
+  }
+
+  /**
+   * @returns the logical bounds for the new range
+   */
+  protected _computeLogicalBounds(newBounds: Rectangle): Rectangle {
+    let unionGridData;
```
bschwarzent: Should "unionGridData" and "gridData" be named "unionBounds" and "bounds"? Return type suggests return value is not a grid data object (unlike _computeGridData).

eclipse-scout-core/src/tile/TileResizeHandler.ts
```diff
@@ -0,0 +1,210 @@
...
+        }
+      }
+    }
+    return unionGridData;
+  }
+
+  /**
+   * @returns the cell bounds for the cell that is on the other side of the dragged edge.
+   */
+  protected _findInitialCell() {
```
bschwarzent: Declare return type

eclipse-scout-core/src/tile/TileResizeHandler.ts
```diff
@@ -0,0 +1,210 @@
...
+        }
+      }
+    }
+    return tileBefore;
+  }
+}
+
+export interface TileResizableModel extends ResizableModel {
+  tileGrid: TileGrid;
+  ignorer?: (tile: Tile) => boolean;
```
bschwarzent, line 212: Document this

eclipse-scout-core/src/tile/TileGrid.ts
```diff
@@ -61,6 +62,8 @@ export class TileGrid<TTile extends Tile = Tile> extends Widget implements TileG
...
+  resizableProducer: (tile: Tile) => Resizable;
```
bschwarzent, line 66: Same issue as with movableProducer.

eclipse-scout-core/src/resizable/Resizable.ts
```diff
@@ -332,3 +356,15 @@ export class Resizable implements ResizableModel, ObjectWithType {
...
   }
 }
+
+export interface ResizableContext {
+  initialBounds: Rectangle;
+  currentBounds: Rectangle;
+  minBounds: Rectangle;
+  maxBounds: Rectangle;
+  distance: number[];
+  edge: string;
```
bschwarzent: Consider using a string union type: edge: 'e' | 'n' | ...;

eclipse-scout-core/src/resizable/Resizable.ts
```diff
@@ -261,49 +260,74 @@ export class Resizable implements ResizableModel, ObjectWithType {
...
       newBounds: newBounds,
       initialBounds: this._context.initialBounds
     });
   }
 
+  protected _resizeEnd() {
+    if (this._context.currentBounds.equals(this._context.initialBounds)) {
+      return;
+    }
+    this.$container.trigger('resizeend', {
```
bschwarzent: Is the spelling here intentional? It does not seem to be a native event, and I could not find any other reference to "resizeend" in the code.

eclipse-scout-core/src/resizable/Resizable.ts
```diff
@@ -261,49 +260,74 @@ export class Resizable implements ResizableModel, ObjectWithType {
...
     this._context = null;
   }
 
   protected _onMousemove(event: MouseMoveEvent) {
-    let ctx = this._context,
-      newBounds = ctx.initialBounds.clone(),
-      distance = this._calcDistance(ctx.mousedownEvent, event);
+    let newBounds = this._computeBounds(event);
+    if (newBounds) {
+      $.throttle(this._resizeHandler, Resizable.FPS)(newBounds);
```
bschwarzent, line 282: Does that work? I think $.throttle returns a new function that wraps the original function. If $.throttle is called during each mouse movement, new wrappers are created, but the subsequent calls are not throttled.

eclipse-scout-core/src/tile/TileGrid.ts
```diff
@@ -294,6 +324,16 @@ export class TileGrid<TTile extends Tile = Tile> extends Widget implements TileG
     this.setTiles([]);
   }
 
+  moveTileBefore(tileToMove: ObjectOrModel<TTile>, sibling: TTile) {
+    let tiles = arrays.moveBefore(this._tiles, tileToMove as Tile, sibling);
```
bschwarzent: Why does the type of the first argument allow models? this._tiles does not contain any models (according to its type), therefore a model could never be moved.

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayout.ts
cguglielmo: this is not possible here since we don't have the $container. The idea is to use scout.setLogicalGridSpyEnabled

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/hybrid/AbstractFormHybridAction.java
cguglielmo: How did it work before my change? Nothing was done with the result. Does it need to be passed to fireHybridWidgetEvent?

eclipse-scout-core/src/tile/TileGrid.ts
cguglielmo: There are keystroke tests using this function, I think this is enough

eclipse-scout-core/src/util/MoveSupport.less
cguglielmo: This may be needed for other elements as well. Analytics uses another approach (left: -1 etc), but this one works for wider boarders as well.

eclipse-scout-core/src/util/MoveSupport.less
cguglielmo: you are right, z-index was probably introduced later

eclipse-scout-core/src/util/MoveSupport.less
cguglielmo: I am quite sure that I got hover effects on the clone (resize-handles, charts) without pointer-events:none, but I cannot reproduce it anymore... But yes a more general rule would be great, maybe > *. But I need to be able to reproduce it first...

eclipse-scout-core/src/util/MoveSupport.less
cguglielmo: I found the name dragging confusing and too generic, that is why I added the clone class. I replaced it now with dragged-clone since we do not need both classes.

eclipse-scout-core/src/util/MoveSupport.less
cguglielmo: Maybe we should remove the transition all along because it is only used for analytics right now

eclipse-scout-core/src/tile/TileGridSelectionHandler.ts
cguglielmo: which part do you mean? The logic is different: if diff is > 0 and selection is empty, the first tile is used. If there is a selection, the last tile is used

The method either selects and focuses or just focuses. I am not sure if And is really better...

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
cguglielmo: This is existing behavior, please file a ticket if you would like to have it fixed.

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
cguglielmo: This is existing behavior, please file a ticket if you would like to have it fixed.

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
cguglielmo: I know, but I think it is easier to understand this way

eclipse-scout-core/src/resizable/Resizable.ts
cguglielmo: This is an existing issue, I'll create a separate pr for this

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/tile/ITileGrid.java
cguglielmo: I have removed the PROP_ prefix

eclipse-scout-core/src/layout/logicalgrid/LogicalGridLayoutInfo.ts
cguglielmo: This works already, see new spec

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/hybrid/AbstractFormHybridAction.java
```diff
@@ -32,34 +43,40 @@ protected boolean isShowFormOnStart() {
...
-    addWidget(form);
-
-    startForm(form);
   }
 
-  protected abstract FORM createForm(DO_ENTITY data);
-
-  protected void prepareForm(FORM form) {
-    form.setShowOnStart(isShowFormOnStart());
+  protected DO_ENTITY exportResultInternal(FORM form) {
```
fschinkel: Please move to other "exportResult"-methods

eclipse-scout-core/src/testing/JasmineScoutUtil.ts
```diff
@@ -101,11 +101,32 @@ export const JasmineScoutUtil = {
...
       fail('Ajax call not mocked for url: ' + request.url + ', method: ' + request.method);
     });
+  },
+
+  /**
+   * Calls the given mock as soon as a hybrid action with the given actionType is called.
+   * The mock is called asynchronously using setTimeout to let the runtime code add any required event listeners first.
+   *
+   * The mock may return an object with [id, widget] if the action is supposed to create widgets.
+   * The id needs to contain the action id and the id for the widget used in the hybrid action (`${actionId}${widgetId}`).
```
fschinkel: This id-format is not always correct, is it? Or it needs to be pointed out that the widget id may be empty. If one wants to mock e.g. an openForm action, the id needs to be `event.data.id`.

The id needs to be the same as created by the `IHybridAction` (see `AbstractHybridAction.addWidget(IWidget)`, `AbstractHybridAction.addWidgets(Map<String, IWidget>)`)

fschinkel (changes requested):

PR #985, opened 2024-04-25
Support CAST and INTERVAL in classical SqlFormatter

imotsch:
Support CAST and INTERVAL in classical SqlFormatter

Added support for CAST and INTERVAL clauses to the
SqlFormatter class.

Examples:
CAST ( t.COLUMN_1 AS NUMBER)
INTERVAL '1 day'

This change does not affect the SQLService at all
and is thus a low impact change improving
effectiveness and productivity of logging done with
the SqlFormatter.


stephan-merkli (commented):
Looks good to me. @matthiaso: could you have a look at it it too? Thanks

matthiaso (approved):

PR #986, opened 2024-04-26
Primary sort order for IDoValueMigrationHandler

LukasHuser:
Untyped data object value migration handlers need to be applied before regular value migrations because they may convert typed values to the expected target type. A typical use case is conversion from UnknownId to a typed target ID in case of renaming a type ID name.

A primary sort order is introduced to allow (primary) sorting independent of the type version (which is used for secondary sorting). The primary sort order can be used to group different types of value migrations.

Regular value migrations (i.e. sub-classes of
AbstractDoValueMigrationHandler use
IDoValueMigrationHandler.DEFAULT_PRIMARY_SORT_ORDER and are applied after untyped value migrations (i.e. sub-classes of AbstractDoValueUntypedMigrationHandler) which use
IDoValueMigrationHandler.UNTYPED_PRIMARY_SORT_ORDER

379569

### DIFF
--- a/org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/migration/IDoValueMigrationHandler.java
+++ b/org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/migration/IDoValueMigrationHandler.java
@@ -36,8 +36,9 @@ import org.eclipse.scout.rt.platform.namespace.NamespaceVersion;
  * constants from the current code base, i.e. they are never refactored, if the code base changes).
  * <p>
  * Value migration handlers are always applied <i>after</i> all {@link IDoStructureMigrationHandler} have been applied.
- * Value migrations are applied in a specific order, defined by the given {@link #typeVersion()}. Migration handlers
- * with the same type version are applied in the order provided by the bean manager.
+ * Value migrations are applied in a specific order, defined by the given {@link #primarySortOrder()} and
+ * {@link #typeVersion()}. Migration handlers with the same primary sort order and type version are applied in the order
+ * provided by the bean manager.
  * <p>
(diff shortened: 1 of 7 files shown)


org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/migration/IDoValueMigrationHandler.java
```diff
@@ -138,6 +139,9 @@
 @ApplicationScoped
 public interface IDoValueMigrationHandler<T> {
 
+  double UNTYPED_PRIMARY_SORT_ORDER = 1000.0;
```
paolobazzi, line 147: add javadoc indicating default usage of these constants? similar like javadoc comment for primarySortOrder() method

paolobazzi (approved):

org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/migration/IDoValueMigrationHandler.java
```diff
@@ -153,7 +168,17 @@ public interface IDoValueMigrationHandler<T> {
   Class<T> valueClass();
 
   /**
-   * Type version used for sorting.
+   * Primary sort order used for sorting. The primary sort order can be used to group different types of value migration
+   * handlers. Secondary sort order is by {@link #typeVersion()}.
+   * <p>
+   * Example: Untyped DO value migrations (see {@link AbstractDoValueUntypedMigrationHandler}) use
+   * {@link #UNTYPED_PRIMARY_SORT_ORDER} and are always applied before regular migration handlers.
```
stephan-merkli: ... before regular migration handlers ({@link AbstractDoValueMigrationHandler}) using {@link #DEFAULT_PRIMARY_SORT_ORDER}).

PR #1000, opened 2024-05-08
CellEditorPopup: ignore key stroke when cell editor opens a form

bschwarzent:
When a keystroke is triggered on a descendant of the cell editor field, the cell editor popup should not be closed. This should also work if the cell editor opens a form, which is not rendered within the cell editor popup's $container.

Consider a cell editor that opens a form that also has an editable column. If the inner cell editor is closed with Enter, the outer cell editor popup should _not_ be completed automatically.

Instead of checking if the key stroke's scope target is part of the $container, we also check the widget hierarchy of the scope target. If it belongs to the cell editor, we don't trigger completeEdit().

378432


bschwarzent:
This change had too many side-effects. Therefore we fixed the problem differently: #1350

PR #1001, opened 2024-05-08
UiSession: fix theme initialization

bschwarzent:
- URL param (?theme=xyz) always takes precedence over the current theme
- Otherwise, if the model already has a theme set, use that one.
- Otherwise, use the "current theme" from the request (session attribute, cookie or default value).

A warning is only logged when the requested theme or the model theme is set but is not valid according to UiThemeHelper. When the model does not specify a theme, a default theme is assigned automatically without log message.

When the theme is determined, it is written back to the model, so the logo URL and potentially other things can be adjusted.

In Scout JS, we have no UiSession or ClientSession. For a similar functionality, the URL param is manually compared with the current theme (called "active theme").

270127

### DIFF
--- a/eclipse-scout-core/src/desktop/Desktop.ts
+++ b/eclipse-scout-core/src/desktop/Desktop.ts
@@ -1682,7 +1682,7 @@ export class Desktop extends Widget implements DesktopModel, DisplayParent {
     let theme = this.theme;
     if (this.url.hasParameter('theme')) {
       theme = strings.nullIfEmpty(this.url.getParameter('theme') as string) || Desktop.DEFAULT_THEME;
-    } else if (theme === null) {
+    } else if (!theme) {
       theme = this._activeTheme();
     }
     this.setTheme(theme);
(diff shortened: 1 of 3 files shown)


eclipse-scout-core/src/desktop/Desktop.ts
```diff
@@ -1680,7 +1680,7 @@ export class Desktop extends Widget implements DesktopModel, DisplayParent {
 
   protected _initTheme() {
     let theme = this.theme;
-    if (this.url.hasParameter('theme')) {
+    if (!this.modelAdapter && this.url.hasParameter('theme')) { // only needed for Scout JS (in Scout classic, the theme URL parameter is handled in UiSession#initUiTheme)
```
fschinkel: move to DesktopAdapter

fschinkel (changes requested):

fschinkel (approved):

eclipse-scout-core/src/desktop/Desktop.ts
bschwarzent: Done

PR #1009, opened 2024-05-14
RemoteFileService: enhance locale handling

cyrill-wyss:
The remote file service uses the given locale to find language dependant templates. This logic is too complex and can be simplified. Also, the implemented security pattern is too restricting (i.e. not allowing # in locales even tho # is valid locale character). Last but not least, there is little comments on the complicated code parts and no unit test at all.

The security pattern is changed into a simple allow list of expected locale characters.
The part which checks if there is already a matching locale at the end of the filename is simplified.
Test caseses for the new function are added.

382967

### DIFF
--- a/org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/services/common/file/RemoteFileService.java
+++ b/org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/services/common/file/RemoteFileService.java
@@ -21,7 +21,9 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
+import java.util.function.Predicate;
 import java.util.regex.Pattern;
 
 import org.eclipse.scout.rt.platform.config.CONFIG;
@@ -34,6 +36,12 @@ import org.eclipse.scout.rt.shared.services.common.file.RemoteFile;
 import org.eclipse.scout.rt.shared.servicetunnel.RemoteServiceAccessDenied;
 
 public class RemoteFileService implements IRemoteFileService {
+
+  private static final String LOCALE_DELIMITER = "_";
+  private static final Pattern PATTERN_LOCALE_DELIMITER = Pattern.compile(LOCALE_DELIMITER);
+  private static final int LOCALE_TEXT_MAX_LENGTH = 64;
+  private static final Pattern LOCALE_SECURITY_PATTERN = Pattern.compile("[-a-z0-9_#]+", Pattern.CASE_INSENSITIVE);
+
   private String m_rootPath;
 
   public RemoteFileService() {
@@ -224,9 +232,7 @@ public class RemoteFileService implements IRemoteFileService {
     }
   }
 
-  private static final Pattern LOCALE_SECURITY_PATTERN = Pattern.compile("[a-z]+([_][a-z0-9]+)?([_][a-z0-9_]+)?", Pattern.CASE_INSENSITIVE);
-
-  private File getFileInternal(RemoteFile spec) {
+  protected File getFileInternal(RemoteFile spec) {
     File root = new File(getRootPath());
     File folder = null;
     if (spec.getDirectory() == null || spec.getDirectory().isEmpty()) {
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/services/common/file/RemoteFileService.java
```diff
@@ -256,43 +257,57 @@ private File getFileInternal(RemoteFile spec) {
...
-      }
-      if (!prefix.endsWith("_")) {
-        prefix = prefix + "_";
+    String filename = tryAddLocaleToFileName(canonicalSimpleName, spec.getLocale(), canonicalFolder, file -> file.exists());
+    File file = new File(canonicalFolder, filename);
+    return file;
+  }
+
+  private static final int LOCALE_TEXT_MAX_LENGTH = 64;
+  private static final Pattern LOCALE_SECURITY_PATTERN = Pattern.compile("[-a-z0-9_.#]+", Pattern.CASE_INSENSITIVE);
```
mvilliger: what is the first "-" used for? The one before "a"? Is it required?

And for security reasons we should block "..". This was not accepted before but would be ok now, right?

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/services/common/file/RemoteFileService.java
```diff
@@ -256,43 +257,57 @@ private File getFileInternal(RemoteFile spec) {
...
+    String localeText = locale.toString().replaceAll("__", "_");
+    if (localeText.length() > LOCALE_TEXT_MAX_LENGTH) {
+      return filename;
+    }
+    // make sure only valid locale characters are accepted.
+    if (!LOCALE_SECURITY_PATTERN.matcher(localeText).matches()) {
+      throw new SecurityException("invalid or inaccessible path");
+    }
+
+    String[] checkedLocaleParts = Arrays.stream(localeText.split("_", 3))
```
mvilliger: You could use Pattern.compile(..).splitAsStream() here? Maybe you can also save the pattern in a constant? The limit can be implemented using limit()

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/services/common/file/RemoteFileService.java
```diff
@@ -256,43 +257,57 @@ private File getFileInternal(RemoteFile spec) {
...
+        .map(s -> "_" + s)
+        .map(s -> s.toLowerCase(Locale.ROOT))
+        .toArray(n -> new String[n]);
```
mvilliger: Could be written as .toArray(String[]::new)

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/services/common/file/RemoteFileService.java
```diff
@@ -256,43 +257,57 @@ private File getFileInternal(RemoteFile spec) {
...
+      return filename;
+    }
+    // make sure only valid locale characters are accepted.
+    if (!LOCALE_SECURITY_PATTERN.matcher(localeText).matches()) {
+      throw new SecurityException("invalid or inaccessible path");
+    }
+
+    String[] checkedLocaleParts = Arrays.stream(localeText.split("_", 3))
+        .map(s -> "_" + s)
+        .map(s -> s.toLowerCase(Locale.ROOT))
```
mvilliger, line 286: Perform lowercase first, as the underscore must not be processed?

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/services/common/file/RemoteFileService.java
cyrill-wyss: The - is necessary as locales may contain it (e.g. DE-de). However, The dot character is not necessary => Removed it from the pattern.

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/services/common/file/RemoteFileService.java
```diff
@@ -256,43 +262,55 @@ private File getFileInternal(RemoteFile spec) {
...
+    }
+    // make sure only valid locale characters are accepted.
+    if (!LOCALE_SECURITY_PATTERN.matcher(localeText).matches()) {
+      throw new SecurityException("invalid or inaccessible path");
+    }
+
+    String[] checkedLocaleParts = PATTERN_LOCALE_DELIMITER.splitAsStream(localeText)
+        .limit(3)
+        .map(s -> s.toLowerCase(Locale.ROOT))
+        .map(s -> "_" + s)
```
mvilliger: here the delimiter constant should be used as well?

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/services/common/file/RemoteFileService.java
```diff
@@ -256,43 +262,55 @@ private File getFileInternal(RemoteFile spec) {
...
-        if (localeText.indexOf('_') == -1) {
-          filename = canonicalSimpleName;
-          break;
-        }
-        localeText = localeText.substring(0, localeText.lastIndexOf('_'));
-        filename = prefix + localeText + suffix;
-        test = new File(canonicalFolder, filename);
+    }
+    if (!prefix.endsWith("_")) {
+      prefix = prefix + "_";
```
mvilliger: here the delimiter constant should be used as well?

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/services/common/file/RemoteFileService.java
```diff
@@ -256,43 +262,55 @@ private File getFileInternal(RemoteFile spec) {
...
+    // Check if a file containing the given locale name or parts of it exists.
+    // Otherwise, return the original filename
+    String extendedFilename = prefix + localeText + suffix;
+    File test = new File(canonicalFolder, extendedFilename);
+    while (!fileExistsPredicate.test(test)) {
+      if (localeText.indexOf('_') == -1) {
+        extendedFilename = filename;
+        break;
       }
+      localeText = localeText.substring(0, localeText.lastIndexOf('_'));
```
mvilliger: here the delimiter constant should be used as well?

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/services/common/file/RemoteFileService.java
```diff
@@ -256,43 +262,55 @@ private File getFileInternal(RemoteFile spec) {
...
+    }
+    if (!prefix.endsWith("_")) {
+      prefix = prefix + "_";
+    }
+    // Check if a file containing the given locale name or parts of it exists.
+    // Otherwise, return the original filename
+    String extendedFilename = prefix + localeText + suffix;
+    File test = new File(canonicalFolder, extendedFilename);
+    while (!fileExistsPredicate.test(test)) {
+      if (localeText.indexOf('_') == -1) {
```
mvilliger: here the delimiter constant should be used as well?

mvilliger (approved):

PR #1014, opened 2024-05-28
Item Summary Contributions

fschinkel:


eclipse-scout-core/src/tile/fields/FormFieldTile.less
```diff
@@ -45,6 +45,7 @@
 
   & > .form-field {
     padding: 16px @tile-field-padding @tile-field-padding;
+    border-radius: inherit;
```
bschwarzent, line 48: Is this correct? FormFields usually don't have a border or a border radius - but the `.field` inside does!

eclipse-scout-core/src/App.ts
```diff
@@ -298,6 +298,7 @@ export class App extends EventEmitter {
     this._installSyntheticActiveStateHandler();
     this._ajaxSetup();
     this._installExtensions();
```
bschwarzent, line 300: I find this extension mechanism quite confusing. Why is there an _installExtensions() method (clearly intended to be overwritten) _and_ an event that can be used to register extensions? What even _are_ extensions? How are they different from objectFactories.ts? The extension instances seem to be useless, they are only created to call the init method.

Edit: okay, I sort of get it now. But can we please improve the documentation? Also it seems to support multiple extensions for the same method, but no mechanism to control their order.

eclipse-scout-core/src/Extension.ts
```diff
@@ -52,10 +52,10 @@ export class Extension<E> {
   /**
    * Calls {@link scout.create} for each extension class in the given extensions array.
    *
-   * @param extensions an array of strings containing extension class names
+   * @param extensions a single objectType or an array of objectTypes
    */
-  static install(extensions: string[]) {
-    extensions.forEach(ext => {
+  static install(extensions: string | ObjectType | (string | ObjectType)[]) {
```
bschwarzent, line 57: Better readable?

```suggestion
  static install(extensions: string | string[] | ObjectType | ObjectType[]) {
```

org.eclipse.scout.rt.api/src/main/java/org/eclipse/scout/rt/api/code/CodeResource.java
```diff
@@ -94,12 +98,13 @@ protected void loadOtherTextsFor(Map<String, CodeTypeDo> codeTypeMap) {
...
     }
 
-    for (Locale otherLanguage : otherAppLanguages) {
-      RunContexts.copyCurrent()
-          .withLocale(otherLanguage)
-          .call(this::getCodeTypesById).values()
-          .forEach(otherCodeTypeDo -> mergeCodeTypeTexts(otherCodeTypeDo, codeTypeMap));
-    }
+    otherAppLanguages.stream()
+        // order by language tag, locales without location come first ([de, en_US, de_DE, en, en_GB] -> [de, de_DE, en, en_GB, en_US])
```
bschwarzent: What do you mean by "location"? Country/region? If so, why is "en" in your example not sorted after "de"? Is the ROOT locale also relevant here?

org.eclipse.scout.rt.api/src/main/java/org/eclipse/scout/rt/api/code/CodeResource.java
```diff
@@ -127,11 +132,44 @@ protected void mergeCodeTexts(CodeDo from, Collection<CodeDo> targetList) {
...
       return;
     }
-    from.getTexts().forEach(target::withText);
+    mergeTexts(from.getTexts(), target.getTexts(), target::withText);
     Collection<CodeDo> targetChildren = getChildCodes(target);
     from.getChildren().forEach(code -> mergeCodeTexts(code, targetChildren));
   }
 
+  /**
+   * Merges the to text maps (languageTag to text). If there is already an entry in the target map with the same text
```
bschwarzent:
```suggestion
   * Merges two text maps (languageTag to text). If there is already an entry in the target map with the same text
```

What exactly does "merge" mean? Is a new map created or is one of the maps altered? It seems that the targetTexts map is directly updated. If so, the description should probably be more like "Merges the entries from the fromTexts into the targetTexts map". But what is the "withText" consumer used for? Why are the maps not just updated? Or a new merged map returned?

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/services/common/code/CodeService.java
```diff
@@ -93,17 +95,20 @@ public void invalidate(ICacheEntryFilter<CodeTypeCacheKey, ICodeType<?, ?>> filt
...
-      catch (ReflectiveOperationException e) {
-        throw BEANS.get(PlatformExceptionTranslator.class)
-            .translate(e)
-            .withContextInfo("key", key)
-            .withContextInfo("codeTypeClass", key.getCodeTypeClass());
-      }
-    };
+    return key -> RunContexts.copyCurrent(true)
+        .withLocale(key.getLocale())
+        .withTransactionScope(TransactionScope.REQUIRED)
```
bschwarzent, line 102: Document why this is needed.

eclipse-scout-core/src/tile/fields/FormFieldTile.less
fschinkel: The example was a GroupBox with a background color. The GroupBox does not have a .field but a .form-field. Therefore, I added it to .form-field in order to handle FormFields and GroupBoxes with the same rule.

eclipse-scout-core/src/Extension.ts
fschinkel: This is missing a mixed array with `string` and `ObjectType`.

bschwarzent (approved):

PR #1021, opened 2024-05-31
Planner: Fix nullPointer when layouting hierarchical menus

Knicksaas:
When the menu property is updated with the same value after the initialization, a null pointer exception occurs when layouting hierarchical menus. This is because the menu gets removed, but not rendered again. The equals check in Widget#setPropertyInternal and MenuBar#setMenuItems are not the same. This change prevents the removal of the menus.

384547


cguglielmo:
jenkins run tests

cguglielmo:
jenkins run tests

PR #1022, opened 2024-05-31
PropertiesHelper: Add 'importenv' key for importing variable files

jmiserez:
This allows importing variable files using the special key 'importenv'.

Contrary to the existing 'import' key, this does not add the entries as config properties and the values are not checked by any IConfigurationValidator.

This is useful to allow importing another variable file from within a variable files, such as the one passed in via -Dscout.env. E.g.:
importenv[0]=anotherfile.properties

or with a relative path:
importenv[1]=${CURRENT_DIR}/somefile.properties

383208


jmiserez:
Will be solved differently, by extending the scout.env variable instead.

PR #1023, opened 2024-06-03
DoEntityDeserializer: Improve log for read attribute value failures

paolobazzi:
Retain message of the causing exception for the logger message.

385075


matthiaso (approved):

paolobazzi:
Message is already available as suppressed message in stacktrace -> close.

PR #1024, opened 2024-06-03
Charts: fix numeric tick labels

fschinkel:
chart.js creates a CategoryScale if labels are set directly on the data. The CategoryScale itself registers a ticks callback which is overridden if the labels need to be reformatted. The original callback calls CategoryScale.getLabelForValue. If the formatters are passed as bound functions there is no access to the scale and therefore, no access to CategoryScale.getLabelForValue. Therefore, pass the formatter functions wrapped in an unbound function that extracts the renderer and the scale. This function then calls the formatter as it was bound to the renderer and passes the scale as an argument. The formatters then use the scale to call CategoryScale.getLabelForValue before custom formatting. Fix empty label for numeric columns in ChartTableControl.

379927

### DIFF
--- a/eclipse-scout-chart/src/chart/ChartJsRenderer.js
+++ b/eclipse-scout-chart/src/chart/ChartJsRenderer.js
@@ -10,7 +10,7 @@
  */
 import {AbstractChartRenderer, Chart, chartJsDateAdapter} from '../index';
 import ChartJs from 'chart.js/auto';
-import {_adapters as chartJsAdapters} from 'chart.js';
+import {_adapters as chartJsAdapters, CategoryScale} from 'chart.js';
 import {arrays, colorSchemes, Event, graphics, numbers, objects, Point, scout, strings, styles, tooltips} from '@eclipse-scout/core';
 import ChartDataLabels from 'chartjs-plugin-datalabels';
 import $ from 'jquery';
@@ -155,9 +155,9 @@ export default class ChartJsRenderer extends AbstractChartRenderer {
     this.colorSchemeCssClass = '';
     this.minRadialChartDatalabelSpace = 25;
 
-    this._labelFormatter = this._formatLabel.bind(this);
-    this._xLabelFormatter = this._formatXLabel.bind(this);
-    this._yLabelFormatter = this._formatYLabel.bind(this);
+    this._labelFormatter = this._createLabelFormatter(this._formatLabel);
+    this._xLabelFormatter = this._createLabelFormatter(this._formatXLabel);
+    this._yLabelFormatter = this._createLabelFormatter(this._formatYLabel);
 
     this._xAxisFitter = this._fitXAxis.bind(this);
     this._yAxisFitter = this._fitYAxis.bind(this);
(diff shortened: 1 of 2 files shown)


eclipse-scout-chart/src/table/controls/ChartTableControl.js
```diff
@@ -749,7 +749,7 @@ export default class ChartTableControl extends TableControl {
         let label,
           keyX = xAxis[x];
         if (xAxis.column instanceof NumberColumn) {
-          label = keyX;
+          label = objects.isNullOrUndefined(keyX) ? xAxis.format(keyX) : keyX;
```
bschwarzent, line 754: What use is it to pass `null`/`undefined` to xAxis.format()?

eclipse-scout-chart/src/chart/ChartJsRenderer.js
```diff
@@ -1485,7 +1486,8 @@ export default class ChartJsRenderer extends AbstractChartRenderer {
     }, plugins.datalabels);
   }
 
-  _formatLabel(label) {
+  _formatLabel(label, scale) {
```
bschwarzent: Move next to the other format functions below (_formatXLabel)

eclipse-scout-chart/src/chart/ChartJsRenderer.js
```diff
@@ -1495,14 +1497,37 @@ export default class ChartJsRenderer extends AbstractChartRenderer {
     }
```
bschwarzent: Use the `?.` operator or return `null` in the else case

eclipse-scout-chart/src/chart/ChartJsRenderer.js
```diff
@@ -1495,14 +1497,37 @@ export default class ChartJsRenderer extends AbstractChartRenderer {
...
   }
 
-  _formatXLabel(label) {
+  /**
+   * Creates an unbound function that calls the given formatter on this {@link ChartJsRenderer}. The context of this
+   * unbound function is passed to the formatter as argument.
+   */
+  _bindLabelFormatter(formatter) {
```
bschwarzent: The method is called "bind (something)", but it returns something that is *not* bound. Isn't that a bit confusing? Unfortunately, I have no better ideas... Maybe `_prepareLabelFormatter` or `_createLabelFormatter`? But since it is internal, it probably doesn't matter.

eclipse-scout-chart/src/table/controls/ChartTableControl.js
fschinkel: The axis formats numbers as two digit decimals and `null`/`undefined` as the text `-empty-` or something similar. We want to leave the number format to the chart but need the `-empty-` string. Therefore, we pass `null`/`undefined` to the axis but nothing else.

bschwarzent: Thanks for the explanation, that makes sense! Maybe add this as a comment in the code so this special behavior is documented.

bschwarzent (approved):

PR #1032, opened 2024-06-07
JsonHybridManager: prevent unnecessary 'hybridWidgetEvent'

bschwarzent:
Add missing break statement in switch block. Otherwise, all hybrid events will generate an additional action event of type 'hybridWidgetEvent' in addition to the normal 'hybridEvent' action event.


fschinkel (approved):
Do we need to backport this to 22.0, 23.1, 23.2, 24.1?

bschwarzent:
Yes, I will cherry-pick the fix to the lower branches.

bschwarzent:
Backported to 23.2 and 24.1 (HybridManager does not yet exist on 22.0)

PR #1035, opened 2024-06-10
Embedded forms: never show form on open

cguglielmo:
If form.open() is called for embedded forms
(wrapped form field, form menu etc.), which happens for example for JsForms, it must not be shown by the desktop because the container is responsible to render the embedded form. If the desktop tries to show the form, it would change the displayParent and therefore also the parent which must be avoided.

383085

### DIFF
--- a/eclipse-scout-core/src/form/fields/wrappedform/WrappedFormField.ts
+++ b/eclipse-scout-core/src/form/fields/wrappedform/WrappedFormField.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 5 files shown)


eclipse-scout-core/src/form/fields/wrappedform/WrappedFormField.ts
```diff
@@ -65,21 +74,16 @@ export class WrappedFormField extends FormField implements WrappedFormFieldModel
...
 
-    this.innerForm.setDisplayHint(Form.DisplayHint.VIEW); // by definition, an inner form is a view.
-    this.innerForm.setModal(false); // by definition, an inner form is not modal.
-    this.innerForm.setClosable(false); // Disable close key stroke
-    this.innerForm.renderInitialFocusEnabled = this.initialFocusEnabled; // do not render initial focus of form if disabled.
-    if (this.innerForm.rootGroupBox) {
-      this.innerForm.rootGroupBox.setMenuBarPosition(GroupBox.MenuBarPosition.BOTTOM);
-    }
-
+    this.innerForm.renderInitialFocusEnabled = this.initialFocusEnabled;
```
fschinkel: This modifies the model of the innerForm. I think we should move this call to `_setInnerForm` (or `_adaptForm`) and `_setInitialFocusEnabled`

eclipse-scout-core/src/form/fields/wrappedform/WrappedFormField.ts
```diff
@@ -65,21 +74,16 @@ export class WrappedFormField extends FormField implements WrappedFormFieldModel
...
     this.innerForm.render();
 
     this.addField(this.innerForm.$container);
     this.innerForm.invalidateLayoutTree();
 
     // required because active element is lost when 'addField' is called.
-    this._renderInitialFocusEnabled();
+    if (this.initialFocusEnabled) {
```
fschinkel: Why do we not check the property of the innerForm, like it is done in `Form._renderProperties`? Can these two properties become inconsistent (e.g. by a custom form that modifies this property during its render-phase)?

fschinkel (changes requested):

fschinkel (approved):

PR #1056, opened 2024-07-11
UiNotifications: add publishOverCluster-property to options

fschinkel:
The publishOverCluster property ca be used to explicitly publish a notification over all cluster nodes or not. The

### DIFF
--- a/org.eclipse.scout.rt.uinotification/src/main/java/org/eclipse/scout/rt/api/uinotification/UiNotificationPutOptions.java
+++ b/org.eclipse.scout.rt.uinotification/src/main/java/org/eclipse/scout/rt/api/uinotification/UiNotificationPutOptions.java
@@ -12,6 +12,14 @@ package org.eclipse.scout.rt.api.uinotification;
 public class UiNotificationPutOptions {
   private Boolean m_transactional;
   private Long m_timeout;
+  private Boolean m_publishOverCluster;
+
+  public UiNotificationPutOptions copy() {
+    return new UiNotificationPutOptions()
+        .withTransactional(getTransactional())
+        .withTimeout(getTimeout())
+        .withPublishOverCluster(getPublishOverCluster());
+  }
 
   /**
    * @param timeout
@@ -27,6 +35,11 @@ public class UiNotificationPutOptions {
     return m_timeout;
   }
 
+  /**
+   * @param transactional
+   *          Whether the notification should be put into the registry only if the current transaction is committed
+   *          successfully or not. Default is {@code true}.
+   */
   public UiNotificationPutOptions withTransactional(Boolean transactional) {
     m_transactional = transactional;
     return this;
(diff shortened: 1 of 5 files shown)


org.eclipse.scout.rt.uinotification/src/main/java/org/eclipse/scout/rt/api/uinotification/UiNotificationPutOptions.java
```diff
@@ -36,7 +48,41 @@ public Boolean getTransactional() {
...
+
+  public Boolean getPublishOverCluster() {
+    return m_publishOverCluster;
+  }
+
+  /*
+   * factory methods
+   */
+
+  /**
```
cguglielmo, line 69: What is the benefit of this create method compared to just using the constructor? I don't think this is necessary.

org.eclipse.scout.rt.uinotification/src/main/java/org/eclipse/scout/rt/api/uinotification/UiNotificationRegistry.java
```diff
@@ -339,7 +342,9 @@ protected void putInternal(UiNotificationMessageDo message, boolean publishOverC
     finally {
       m_lock.writeLock().unlock();
     }
-    if (publishOverCluster) {
+    if (Optional.ofNullable(options)
```
cguglielmo: Please use same pattern as in other methods

    if (options == null) {
      options = new UiNotificationPutOptions();
    }

org.eclipse.scout.rt.uinotification/src/main/java/org/eclipse/scout/rt/api/uinotification/UiNotificationPutOptions.java
```diff
@@ -27,6 +35,10 @@ public Long getTimeout() {
     return m_timeout;
   }
 
+  /**
+   * @param transactional
+   *     Whether a notification is processed as a transaction member or not. Default is {@code true}.
```
cguglielmo: Maybe write more clearly:

Whether the notification should be put into the registry only if the current transaction is committed successfully.

org.eclipse.scout.rt.uinotification/src/main/java/org/eclipse/scout/rt/api/uinotification/UiNotificationPutOptions.java
```diff
@@ -36,7 +48,41 @@ public Boolean getTransactional() {
     return m_transactional;
   }
 
+  /**
+   * @param publishOverCluster
+   *          Whether a notification is published to all other cluster nodes or not. Default is {@code true}.
```
cguglielmo: Whether 'the' notification

cguglielmo (commented):
Remove "The" at the end of the commit comment

PR #1058, opened 2024-07-12
373737 konfigurierbare Kachel-Tabellen-Anker

mgollwitzer-bsi:


eclipse-scout-core/src/table/Table.less
```diff
@@ -655,3 +655,19 @@
 .organize-columns-behind-scrollbar-column.table-cell.last {
   padding: 0;
 }
+
+.anchored-table {
```
egloff: This css-class is only used in our internal crm-repo and therefore should be defined there...

eclipse-scout-core/src/table/TableTileGridMediator.ts
```diff
@@ -294,7 +294,7 @@ export class TableTileGridMediator extends Widget implements TableTileGridMediat
   protected _createTileAccordion(): TileAccordion {
     return scout.create(TileAccordion, {
       parent: this.table,
-      virtual: true,
+      virtual: false,
```
egloff: I don't think it is good to change the default behavior of all TileAccordions created by the TableTileGridMediator to "not-virtual". This should be set via a property only where it is needed (analogue gridColumnCount).

egloff (changes requested):

PR #1059, opened 2024-07-15
Synchronize inactive codes to the UI, return active only by default

bschwarzent:
Codes of a code type can be active or inactive. When a code type is marked with @ApiExposed, the corresponding codes are automatically transferred to the UI.

Previously, only active codes were synchronized. To allow the UI to resolve IDs of inactive codes, they have to be synchronized as well. Because existing code should not behave differently, CodeType#codes still only returns active codes by default. To also return inactive codes, the option {activeOnly: true} should be set.

Likewise, the method CodeType#visitChildren will only visit active codes, unless the argument activeOnly=true is passed explicitly.

### DIFF
--- a/eclipse-scout-core/src/code/Code.ts
+++ b/eclipse-scout-core/src/code/Code.ts
@@ -117,16 +117,19 @@ export class Code<TCodeId> implements ObjectWithType {
 
   /**
    * Visits all children of this Code recursively without visiting this Code itself.
+   *
+   * By default, only active child codes are visited. To visit inactive child codes as well, set the `activeOnly` argument to false.
    */
-  visitChildren(visitor: TreeVisitor<Code<TCodeId>>): boolean | TreeVisitResult {
-    for (let i = 0; i < this.children.length; i++) {
-      let child = this.children[i];
+  visitChildren(visitor: TreeVisitor<Code<TCodeId>>, activeOnly = true): boolean | TreeVisitResult {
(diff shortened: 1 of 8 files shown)


eclipse-scout-core/src/code/Code.ts
```diff
@@ -117,16 +117,19 @@ export class Code<TCodeId> implements ObjectWithType {
...
   /**
    * Visits all children of this Code recursively without visiting this Code itself.
+   *
+   * By default, only active child codes are visited. To visit inactive child codes as well, set the `activeOnly` argument to false.
    */
-  visitChildren(visitor: TreeVisitor<Code<TCodeId>>): boolean | TreeVisitResult {
-    for (let i = 0; i < this.children.length; i++) {
-      let child = this.children[i];
+  visitChildren(visitor: TreeVisitor<Code<TCodeId>>, activeOnly = true): boolean | TreeVisitResult {
+    let children = this.children.filter(child => child.active || !activeOnly);
```
mvilliger: Here, if activeOnly is false (=all codes should be visited), the filter could be skipped instead of filtering with all codes accepting the filter. But maybe this optimization is not necessary.

mvilliger (approved):

mvilliger:
The commit message seems to be wrong (2 places where activeOnly should be false?)

bschwarzent:
Commit message was adjusted.

mvilliger (approved):

PR #1062, opened 2024-07-18
HttpProxy: Fix cookie support

matthiaso:
Restore previous behavior: Cookies must use session cookie store if proxy method is called from within a session; default cookie store must only be used if no session available.

### DIFF
--- a/org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
+++ b/org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
@@ -20,6 +20,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Date;
 import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.List;
@@ -27,10 +28,11 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.concurrent.Callable;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
-import java.util.function.Supplier;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -42,6 +44,9 @@ import jakarta.servlet.ServletInputStream;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 
+import org.apache.hc.client5.http.ContextBuilder;
+import org.apache.hc.client5.http.cookie.Cookie;
+import org.apache.hc.client5.http.cookie.CookieStore;
 import org.apache.hc.client5.http.impl.async.CloseableHttpAsyncClient;
 import org.apache.hc.client5.http.protocol.HttpClientContext;
 import org.apache.hc.core5.concurrent.FutureCallback;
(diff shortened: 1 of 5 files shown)


org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
```diff
@@ -231,8 +256,22 @@ public void proxy(HttpServletRequest req, HttpServletResponse resp, HttpProxyReq
...
   protected HttpContext createHttpContext() {
-    Supplier<HttpClientContext> httpClientContextSupplier = getHttpClientContextSupplier();
-    return httpClientContextSupplier != null ? httpClientContextSupplier.get() : HttpClientContext.create();
+    ContextBuilder contextBuilder = ContextBuilder.create();
+
+    // cookie store
+    ISession currentSession = ISession.CURRENT.get();
+    if (currentSession != null && m_defaultCookieStore != null) {
+      contextBuilder.useCookieStore(createSpecificSessionCookieStore(currentSession));
+    }
```
bschwarzent, line 277: Consider: `else if (m_defaultCookieStore != null) { contextBuilder.useCookieStore(m_defaultCookieStore); }`

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
```diff
@@ -670,4 +709,46 @@ public void cancelled() {
       }
     };
   }
+
+  /**
+   * Cookie store which uses {@link #m_defaultCookieStore} with all operations run for the specified {@link ISession}.
+   * May be used if async threads access the cookie store where {@link ISession#CURRENT} is not set correctly.
+   */
+  protected class SpecificSessionCookieStore implements CookieStore {
```
bschwarzent: private?

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
```diff
@@ -88,13 +97,15 @@ public class HttpProxy {
   private String m_remoteBaseUrl;
   private final List<IHttpHeaderFilter> m_requestHeaderFilters;
   private final List<IHttpHeaderFilter> m_responseHeaderFilters;
+  private final CookieStore m_defaultCookieStore;
```
bschwarzent, line 99: protected getter?

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
```diff
@@ -670,4 +722,46 @@ public void cancelled() {
       }
     };
   }
+
+  /**
+   * Cookie store which uses {@link #m_defaultCookieStore} with all operations run for the specified {@link ISession}.
+   * May be used if async threads access the cookie store where {@link ISession#CURRENT} is not set correctly.
+   */
+  public class SpecificSessionCookieStore implements CookieStore {
```
bschwarzent: Ist the public modifier useful? The class is not static, thus it cannot be created to create a new instance without a surrounding instance. It might be useful to check "instanceof", but is this really a valid use case? The class also seems not to be very extensible, because some members are private, so we might as well make the whole class private.

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
```diff
@@ -670,4 +722,46 @@ public void cancelled() {
...
+
+    private ISession m_session;
+
+    public SpecificSessionCookieStore(ISession session) {
+      m_session = session;
+    }
+
+    @Override
+    public void addCookie(Cookie cookie) {
+      runWithSession(() -> getDefaultCookieStore().addCookie(cookie));
```
bschwarzent, line 741: Did you change this from `m_defaultCookieStore` to `getDefaultCookieStore()` intentionally? The method createHttpContext() still directly accesses the member. Maybe this should both use the same pattern?

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
matthiaso: Thanks, changed to private

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
matthiaso: Using `getDefaultCookieStore()` everywhere (adds the possibility to change behavior in sub-classes).

bschwarzent (approved):

PR #1065, opened 2024-07-23
TileUtil: improve matrix

cguglielmo:
- if only tiles with x or y < 0 are used, the matrix.x/y must be set to 0. Currently, they are set to MAX_SAFE_INTEGER
- minWidth/minHeight are not used anymore -> drop these parameters


cguglielmo:
jenkins run tests

bschwarzent (approved):

PR #1069, opened 2024-07-24
Apache Http Client: Enable idle/expired connection eviction

matthiaso:
389832

### DIFF
--- a/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/HttpConfigurationProperties.java
+++ b/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/HttpConfigurationProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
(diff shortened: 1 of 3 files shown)


org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/HttpConfigurationProperties.java
```diff
@@ -153,4 +155,54 @@ public String getKey() {
       return "scout.http.retryOnSocketExceptionByConnectionReset";
     }
   }
+
+  /**
+   * Enable automatic eviction of idle connections with the specified number in seconds (use zero to disable).
```
paolobazzi, line 160: Javadoc in constructor tells "zero to disable", description "negative number to disable"  ->adjust

paolobazzi (approved):

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/HttpConfigurationProperties.java
matthiaso: Thanks, description will be changed to `zero to disable` as it is a AbstractPositiveIntegerConfigProperty - negative numbers should be disallowed.

PR #1070, opened 2024-07-25
HttpProxy: Fix cookie support

matthiaso:
Restore previous behavior: Cookies must use session cookie store if proxy method is called from within a session; default cookie store must only be used if no session available.

383346


matthiaso:
Cherry-picking https://github.com/eclipse-scout/scout.rt/commit/1a485685c11432c7c298399df69067d042ff7dd5

PR #1082, opened 2024-08-02
Improve resolution of textKeys and iconIds

bschwarzent:
- Match entire string instead of substrings (that was most likely not intentional)
- The default namespace for iconIds is 'scout'. This now also works if the icons object is replaced.
- Cleanup code
- Cleanup JsDoc
- Add specs

### DIFF
--- a/eclipse-scout-core/src/util/icons.ts
+++ b/eclipse-scout-core/src/util/icons.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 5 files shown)


eclipse-scout-core/src/util/icons.ts
```diff
@@ -134,26 +134,25 @@ export const icons = {
...
+      let iconId = match[1];
+      let parts = iconId.split('.');
+      if (parts.length === 1) {
+        // look for icon in global object scout.icons.[0]
+        value = window['scout']['icons'][parts[0]];
+      } else if (parts.length === 2) {
         // look for icon in global object [0].icons.[1]
-        // @ts-expect-error
-        value = window[tmp[0]].icons[tmp[1]];
+        value = window[parts[0]]['icons'][parts[1]];
```
fschinkel, line 155: potential null pointer if namespace does not exist

fschinkel (changes requested):

bschwarzent:
Thanks. The change was already submitted to 25.2 with PR #1528. I somehow missed your review on this PR. The code has now been improved accordingly with an additional commit: https://github.com/eclipse-scout/scout.rt/commit/b55118d8db18eff026d806f4d09a3fe5575cbe74

PR #1083, opened 2024-08-02
Features/bsh/24.2/390261 fix date format

bschwarzent:


cguglielmo (commented):
Commit message has a small typo: "This lead to a" -> "This leads to a". Besides that, it looks good to me

PR #1088, opened 2024-08-05
TreeProposalChooser: Fix hierarchy when duplicated lookupRows provided

nsteger:
For hierarchical smart fields where a child node is provided multiple times in the lookup result, the parent may wasn't resolved correctly such that the child node was displayed at top level. In Java, duplicated lookup rows are already filtered for hierarchical SmartFields, see
HierarchicalLookupResultBuilder.getRowsWithParents(List>, VALUE). Analogous logic is now added in TypeScript.

Revert 251c05954edee6e1a6fc992320b56a3f964a2af8 since now solved generally, independent of the LookupCall.


cguglielmo (commented):
Looks good.
Could you please fix the incorrect sentence in the commit message?
"parent **may wasn't** resolved correctly"

nsteger:
> Looks good. Could you please fix the incorrect sentence in the commit message? "parent **may wasn't** resolved correctly"

done

PR #1089, opened 2024-08-05
ChartFieldTile: fix colorScheme if tileWidget changes

fschinkel:
The colorScheme of the ChartFieldTile is always transferred to the
chart. The tile listens to all changes of the charts config and adjusts
the colorScheme accordingly. In addition, the tile needs to update the
colorScheme if the tileWidget changes.

390174


bschwarzent (approved):
♥

PR #1090, opened 2024-08-06
Features/cgu/24.2/resize move 389402

cguglielmo:


eclipse-scout-core/src/form/fields/smartfield/TreeProposalChooser.ts
```diff
@@ -215,6 +215,18 @@ export class TreeProposalChooser<TValue> extends ProposalChooser<TValue, Tree, P
     return leafs;
   }
 
+  /**
+   * This function creates a list of flat tree nodes from a list of lookup rows.
```
bschwarzent:
```suggestion
   * This function creates a flat list of tree nodes from a list of lookup rows.
```

eclipse-scout-core/src/form/fields/smartfield/TreeProposalChooser.ts
```diff
@@ -83,7 +83,7 @@ export class TreeProposalChooser<TValue> extends ProposalChooser<TValue, Tree, P
       appendResult = scout.nvl(result.appendResult, false);
 
     if (appendResult) {
-      treeNodesFlat = lookupRows.map(this._createTreeNode.bind(this));
```
bschwarzent: Typos in commit message:
grafik

eclipse-scout-core/src/resizable/Resizable.ts
```diff
@@ -229,15 +231,42 @@ export class Resizable implements ResizableModel, ObjectWithType {
...
   }
 
   protected _onMouseUp(event: MouseUpEvent) {
+    this.finish();
+  }
+
+  /**
+   * Finishes the resizing by cleaning up all temporary states and triggering the `resizeEnd` event with the initial and new bounds.
+   */
+  public finish() {
```
bschwarzent: We usually don't write the "public" modifier in TS.

eclipse-scout-core/src/main.less
```diff
@@ -326,18 +326,16 @@ button {
...
-.col-resize,
-.col-resize * { /* NOSONAR (universal selector is okay here) */
-  cursor: col-resize !important; /* NOSONAR (!important is okay here) */
-}
-
-.row-resize,
-.row-resize * { /* NOSONAR (universal selector is okay here) */
-  cursor: row-resize !important; /* NOSONAR (!important is okay here) */
-}
+#scout.resize(col);
```
bschwarzent: Maybe name the mixin "resize-cursor" instead of just "resize"? Right now, it is not entirely obvious what this block does (without looking at the documentation of the mixin).

eclipse-scout-core/src/style/mixins.less
```diff
@@ -633,4 +633,14 @@
...
     }
   }
+
+  /**
+   * Adds a rule for a css class with the name @{edge}-resize that defines the cursor with the same name, e.g. resize-w.
+   */
+  .resize(@edge) {
+    .@{edge}-resize,
+    .@{edge}-resize * { /* NOSONAR (universal selector is okay here) */
+      cursor: ~'@{edge}-resize !important'; /* NOSONAR !important is necessary to override any defined cursor on underlying components */
```
bschwarzent, line 643: You probably tested this, but why is the value a string? Isn't this usually done to prevent LESS from converting anything inside? But @{edge} should be replaced with the actual value.

eclipse-scout-core/src/form/fields/smartfield/TreeProposalChooser.ts
cguglielmo: This commit should not be in this PR.... :/

eclipse-scout-core/src/form/fields/smartfield/TreeProposalChooser.ts
cguglielmo: This commit should not be in this PR.... :/

eclipse-scout-core/src/resizable/Resizable.ts
cguglielmo: true

eclipse-scout-core/src/main.less
cguglielmo: good idea

eclipse-scout-core/src/style/mixins.less
cguglielmo: This is the only way I got it to compile, the output looks correct, the single quotes will be gone and the variable replaced.

PR #1091, opened 2024-08-06
eclipse-scout/chart: add exports in package.json

fschinkel:
The test code contains a "require" which is interpreted as commonjs
feature. When a module depends on eclipse-scout/chart it interprets
the files in this dependency as ES module if explicitly stated in the
package.json or if there are no commonjs features in the source code. As
we can not eliminate the "require" from the test code the exports are
added to the package.json to ensure only the correct run time files are
checked.
Exclude test d.ts files from packaging.

### DIFF
--- a/eclipse-scout-chart/package.json
+++ b/eclipse-scout-chart/package.json
@@ -20,14 +20,26 @@
     "chart"
   ],
   "publishConfig": {
+    "exports": {
+      ".": {
+        "types": "./dist/d.ts/src/index.d.ts",
+        "import": "./dist/eclipse-scout-chart.esm.js"
+      },
+      "./src/*": "./src/*"
+    },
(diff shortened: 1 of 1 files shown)


eclipse-scout-chart/package.json
```diff
@@ -20,14 +20,24 @@
     "chart"
   ],
   "publishConfig": {
+    "exports": {
+      ".": {
```
mvilliger, line 24: is the dot really required inside exports?

fschinkel: it needs to override the entry on l. 35 in the published module

mvilliger: yes, sure, but can it not be removed there as well? and also the src entry might not be required. But anyway. It should work like this as well and is aligned with the core module. So both is fine with me.

PR #1093, opened 2024-08-07
Application: Add log for platform and application startup duration

paolobazzi:
391404

### DIFF
--- a/org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/Application.java
+++ b/org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/Application.java
@@ -88,12 +88,16 @@ public class Application {
   protected final AtomicReference<Server> m_server = new AtomicReference<>();
 
   public static void main(String[] args) {
+    long t0 = System.nanoTime();
     LOG.info("Starting platform");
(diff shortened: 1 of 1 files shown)


org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/Application.java
```diff
@@ -84,13 +84,15 @@ public class Application {
...
   protected final AtomicReference<Server> m_server = new AtomicReference<>();
 
   public static void main(String[] args) {
     LOG.info("Starting platform");
     IPlatform platform = Platform.get();
     platform.awaitPlatformStarted();
+    LOG.info("Platform start took {} ms", StringUtility.formatNanos(System.nanoTime() - START_TIME));
 
     LOG.info("Starting application");
     INSTANCE.get().start();
```
brunokoeferli, line 99: I would log the duration of the application start here (instead of in the `logServerReady` method). This reduce changes in start methods and do not require to pass variables between methods.

```suggestion
    INSTANCE.get().start();
    LOG.info("Application start took {} ms, total startup (platform/application) took {} ms" ....
```

brunokoeferli:
lgtm

PR #1096, opened 2024-08-08
FormFieldTile: fix style

fschinkel:
The border radius of the .field was set to inherit to ensure it never overflows its parent, the .form-field. As the .field has a padding this was not really necessary and led to content being cut of. Therefore, do not inherit border radius to the .field on a FormFieldTile but set the overflow of .form-field to hidden.

390226


cguglielmo (commented):
Please explain in the commit message why this change was necessary.

PR #1101, opened 2024-08-13
ConfigurationUtility.isMethodOverwrite inefficient reflection

paolobazzi:
Add lazy initialized cache for declared methods for each queried class. Additionaly fetching all declared methods of a class avoids catching NoSuchMethodExceptions when trying to find a method.

359087

### DIFF
--- a/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/reflect/ConfigurationUtility.java
+++ b/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/reflect/ConfigurationUtility.java
@@ -9,10 +9,13 @@
  */
 package org.eclipse.scout.rt.platform.reflect;
 
+import static org.eclipse.scout.rt.platform.util.Assertions.assertNotNull;
+
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/reflect/ConfigurationUtility.java
```diff
@@ -210,32 +217,44 @@ public static <T> T newInnerInstance(Object instance, Class<T> innerClass) {
...
-    try {
-      declaredMethod = declaringType.getDeclaredMethod(methodName, parameterTypes);
-    }
-    catch (NoSuchMethodException | SecurityException e) {
-      LOG.error("cannot find declared method {}.{}", declaringType.getName(), methodName, e);
+    assertNotNull(declaringType, "declaringType must not be null");
+    assertNotNull(methodName, "methodName must not be null");
+    Method declaredMethod = getDeclaredMethod(declaringType, methodName, parameterTypes);
+    if (declaredMethod == null) {
+      LOG.error("cannot find declared method {}.{}", declaringType.getName(), methodName);
```
mvilliger, line 224: Is log level error really correct? Maybe warn should be sufficient?

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/reflect/ConfigurationUtility.java
```diff
@@ -210,32 +217,44 @@ public static <T> T newInnerInstance(Object instance, Class<T> innerClass) {
...
+   */
+  public static Method getDeclaredMethod(Class<?> clazz, String methodName, Class[] parameterTypes) {
+    for (Method method : getAllDeclaredMethods(clazz)) {
+      if (method.getName().equals(methodName) && Arrays.equals(method.getParameterTypes(), parameterTypes)) {
+        return method;
+      }
+    }
+    return null;
+  }
+
```
mvilliger, line 250: We could also introduce a cache for the method signature (declaring class, name, param types). Then the search loop would no longer be necessary?

PR #1102, opened 2024-08-13
Ensure creation time is unique per node

mvilliger:
If multiple notifications are added within the same millisecond the creation time should still be unique for all notifications.


mvilliger:
Jenkins run tests

PR #1107, opened 2024-08-15
Improve error text for cyclic dependencies

mfi123:
The dependencies displayed in the exception are only a subset of all dependencies. Error text improved to avoid confusion.

383291


egloff (approved):
Looks good to me

PR #1111, opened 2024-08-19
IdCodec: use IdCodecException

fschinkel:
Introduce new IdCodecException and use it in IdCodec whenever an exception is thrown during serialization/deserialization of an IId. Also add an IdCodecExceptionMapper, that returns a "400 - Bad Request" when an IdCodecException is thrown and only logs the stacktrace if debug is enabled.

391961

### DIFF
--- a/org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/id/IdCodec.java
+++ b/org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/id/IdCodec.java
@@ -10,7 +10,7 @@
 package org.eclipse.scout.rt.dataobject.id;
 
 import static org.eclipse.scout.rt.platform.security.SecurityUtility.createMac;
-import static org.eclipse.scout.rt.platform.util.Assertions.*;
+import static org.eclipse.scout.rt.platform.util.Assertions.assertNotNull;
 import static org.eclipse.scout.rt.platform.util.Base64Utility.encodeUrlSafe;
 import static org.eclipse.scout.rt.platform.util.CollectionUtility.hashSet;
 import static org.eclipse.scout.rt.platform.util.ObjectUtility.isOneOf;
@@ -23,6 +23,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
 import java.util.function.Function;
@@ -69,9 +70,9 @@ public class IdCodec {
      */
     LENIENT,
     /**
-     * This will create a signature using the unqualified serialized
-     * {@link IId} and add it as a suffix. If this suffix is incorrect an error is thrown during deserialization.
-     * {@link IId}s can be excluded from using signature creation using {@link IdSignature}.
+     * This will create a signature using the unqualified serialized {@link IId} and add it as a suffix. If this suffix
+     * is incorrect an error is thrown during deserialization. {@link IId}s can be excluded from using signature
+     * creation using {@link IdSignature}.
      */
     SIGNATURE
   }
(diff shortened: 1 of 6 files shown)


org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/id/IdCodec.java
```diff
@@ -382,11 +386,17 @@ public String removeSignature(Class<? extends IId> idClass, String unqualifiedId
...
     }
-    assertEqual(unqualifiedIdSignatureParts.length, 2, "Unqualified id must be signed.");
-    assertEqual(unqualifiedIdSignatureParts[1], createSignature(unqualifiedIdSignatureParts[0]), "Signature of unqualified id does not match.");
+    if (unqualifiedIdSignatureParts.length != 2) {
+      throw new IdCodecException("Unqualified id must be signed.");
+    }
+    if (!Objects.equals(unqualifiedIdSignatureParts[1], createSignature(unqualifiedIdSignatureParts[0]))) {
+      throw new IdCodecException("Signature of unqualified id does not match.");
+    }
   }
```
paolobazzi, line 403: maybe write a bit shorter:
```
    if (!isOneOf(IdCodecFlag.SIGNATURE, flags) || !idInventory().isIdSignature(idClass)) {
      checkEquals(unqualifiedIdSignatureParts.length, 1, "Unqualified id must not be signed.");
      return;
    }
    checkEquals(unqualifiedIdSignatureParts.length, 2, "Unqualified id must be signed.");
    checkEquals(unqualifiedIdSignatureParts[1], createSignature(unqualifiedIdSignatureParts[0]), "Signature of unqualified id does not match.");
  }

  protected void checkEquals(Object o1, Object o2, String msg) {
    if (!Objects.equals(o1, o2)) {
      throw new IdCodecException(msg);
    }
  }
````

File: org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/id/IdCodecException.java [hunk 2]
```diff
@@ -0,0 +1,53 @@
...
+import org.slf4j.Logger;
+import org.slf4j.helpers.FormattingTuple;
+import org.slf4j.helpers.MessageFormatter;
+
+/**
+ * Represents errors that occur during serialization/deserialization in the {@link IdCodec}.
+ *
+ * @since 24.2
+ */
+public class IdCodecException extends RuntimeException {
```
paolobazzi: inherit from PlatformException

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/exception/IdCodecExceptionMapper.java
```diff
@@ -0,0 +1,37 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class IdCodecExceptionMapper extends AbstractExceptionMapper<IdCodecException> {
+
+  private static final Logger LOG = LoggerFactory.getLogger(IdCodecExceptionMapper.class);
+
+  @Override
+  protected Response toResponseImpl(IdCodecException exception) {
+    if (LOG.isDebugEnabled()) {
```
paolobazzi: Shorten to
```
//noinspection PlaceholderCountMatchesArgumentCount
LOG.info("{}: {}", exception.getClass().getSimpleName(), exception.getMessage(), LOG.isDebugEnabled() ? exception : null);
```
And I would suggest to use level INFO as in IllegalArgumentExceptionMapper and JsonParseExceptionMapper.
An invalid request from outside should not lead to log warns,

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/exception/IdCodecExceptionMapper.java
```diff
@@ -0,0 +1,37 @@
...
+
+import jakarta.ws.rs.core.Response;
+
+import org.eclipse.scout.rt.dataobject.id.IdCodecException;
+import org.eclipse.scout.rt.platform.BEANS;
+import org.eclipse.scout.rt.rest.error.ErrorResponseBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class IdCodecExceptionMapper extends AbstractExceptionMapper<IdCodecException> {
```
paolobazzi, line 20: Add testcase in org.eclipse.scout.rt.rest.exception.DefaultExceptionMapperTest

PR #1113, opened 2024-08-22
Testing: improve check of async jobs started in test cases

andibur:
Add execution hint for background jobs executing deferred operations that are not directly attached to a particular test case.

391676

### DIFF
--- a/org.eclipse.scout.rt.platform.test/src/main/java/org/eclipse/scout/rt/testing/platform/runner/statement/AssertNoRunningJobsStatement.java
+++ b/org.eclipse.scout.rt.platform.test/src/main/java/org/eclipse/scout/rt/testing/platform/runner/statement/AssertNoRunningJobsStatement.java
@@ -10,15 +10,16 @@
 package org.eclipse.scout.rt.testing.platform.runner.statement;
 
 import static org.junit.Assert.fail;
-import java.util.ArrayList;
+
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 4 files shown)


org.eclipse.scout.rt.platform.test/src/main/java/org/eclipse/scout/rt/testing/platform/runner/statement/AssertNoRunningJobsStatement.java
```diff
@@ -65,37 +69,56 @@ public void evaluate() throws Throwable {
...
   /**
-   * Finds running job names which comply with the given filter.
+   * Waits until the given futures are done and removes them.
    */
-  private List<String> findJobNames(final Predicate<IFuture<?>> filter) {
-    final List<String> jobs = new ArrayList<>();
-    for (final IFuture<?> future : Jobs.getJobManager().getFutures(filter)) {
-      jobs.add(future.getJobInput().getName());
+  private void removeAndAwaitDone(Set<IFuture<?>> futures) {
+    // implementation note: we do not use IJobManager.awaitDone because we want to report jobs running longer than 500ms.
```
bschwarzent:
```suggestion
    // implementation note: we do not use IJobManager.awaitDone because we want to report jobs running longer than REPORT_THRESHOLD_NANOS
```

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/job/JobInput.java
```diff
@@ -34,6 +34,12 @@
 @Bean
 public class JobInput {
 
+  /**
+   * Execution hint used to advise Scout testing not to wait for marked jobs to complete before continuing with the test
+   * execution. See AssertNoRunningJobsStatement.
```
bschwarzent: Advice users to only add this hint when really necessary.

PR #1114, opened 2024-08-23
Table/Tree: fix scrollToSelection() while opening form

bschwarzent:
- Remove filter for listener (event history is expected to receive all events)
- When revealing the selection, not only check if the widget is rendered but also check if the layout is valid. Otherwise, the measured scroll position will be wrong.

393011


bschwarzent:
Replaced by https://github.com/eclipse-scout/scout.rt/pull/1665

PR #1120, opened 2024-08-28
ComboMenu: fix prefSize when child actions have a custom layout

bschwarzent:
A ComboMenu may contain a FormFieldMenu which in turn contains a FormField, e.g. a Button. When the MenuBar is layouted initially, it reads the preferred size of each menu item before they are actually layouted. The NullLayout just ready the current CSS size from the DOM. This is not necessarily correct if the inner layout (e.g. a Button) relies on its layout to report the correct preferred size. To fix this, we explicitly set a ColumnLayout to the ComboMenu that has a different implementation of the preferredLayoutSize() method.

389822


cguglielmo:
Jenkins run tests

PR #1125, opened 2024-09-02
Menubar: don't wrap right menu box

cguglielmo:
It can happen that the right menu box wraps if it is slightly too large and does not fully fit into the menubar.

The bug was introduced with 24.1 with the change of the DOM order of the menu boxes due to accessibility reasons.
It looks like wrapping cannot be controlled easily when using float: right, so it now uses display: flex and margin-left: auto which seems to work reliably.

388293


cguglielmo:
Jenkins run tests

PR #1129, opened 2024-09-03
Show tooltip when unchecking last calendar

Knicksaas:


eclipse-scout-core/src/calendar/ResourcePanel.ts
```diff
@@ -41,29 +44,60 @@ export class ResourcePanel extends Widget {
...
     if (arrays.hasElements(this.tree.checkedNodes)) {
       super._onTreeNodesChecked(event);
     } else if (!this._populating) {
       // Reapply the value to the tree
       this._syncValueToTree(this.value);
+      this._createImpossibleToUncheckTooltip(event.nodes[0]);
+    }
+  }
+
+  protected _createImpossibleToUncheckTooltip(node: TreeNode) {
```
nsteger: I wouldn't use the concept of TooltipSupport here, instead create Tooltip directly, see for example:
StatusIndicator._showTooltip

I wouldn't even use a tooltip here but a popup instead? I tried this in the GUI and the tooltip feels unusual to me. When a calendar name is too long, I can get two overlapping tooltips.

eclipse-scout-core/src/calendar/StaticTooltip.ts
```diff
@@ -0,0 +1,31 @@
...
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {InitModelOf, TooltipSupport} from '../index';
+
+export const STATIC_TOOLTIP_VISIBLE = 3000;
+
+export class StaticTooltip extends TooltipSupport {
```
nsteger: I would delete this class

nsteger (commented):
commits need to be squashed before merge

eclipse-scout-core/src/tooltip/Tooltip.ts
```diff
@@ -78,6 +78,11 @@ export class Tooltip extends Widget implements TooltipModel {
     this._openLater = false;
   }
 
+  protected override _init(model: InitModelOf<this>) {
+    super._init(model);
+    this.resolveTextKeys(['text']);
```
nsteger: did you intentionally not revert this change? If so, the change should be in a separate commit with corresponding commit comment

Knicksaas: Yes I did, because it's a bug and why not fixing it right away. Fixing this in a separate commit seems legit. Can you also review this PR: #1198 and merge when everything is OK?

nsteger (approved):
backport to 24.2 as discussed, but code ok

PR #1134, opened 2024-09-04
HTTP client connection manager pools metrics

paolobazzi:
Add OpenTelemetry metrics for HTTP client connection pool:
- active/idle connections
- max connections

393561

### DIFF
--- a/org.eclipse.scout.rt.shared.test/src/test/java/org/eclipse/scout/rt/shared/servicetunnel/http/HttpServiceTunnelTest.java
+++ b/org.eclipse.scout.rt.shared.test/src/test/java/org/eclipse/scout/rt/shared/servicetunnel/http/HttpServiceTunnelTest.java
@@ -20,7 +20,6 @@ import java.io.OutputStream;
 import org.eclipse.scout.rt.platform.serialization.SerializationUtility;
 import org.eclipse.scout.rt.shared.SharedConfigProperties.ServiceTunnelTargetUrlProperty;
 import org.eclipse.scout.rt.shared.http.AbstractHttpTransportManager;
-import org.eclipse.scout.rt.shared.http.IHttpTransportBuilder;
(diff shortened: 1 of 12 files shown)


org.eclipse.scout.rt.shared.test/src/test/java/org/eclipse/scout/rt/shared/servicetunnel/http/HttpServiceTunnelTest.java
```diff
@@ -86,6 +85,11 @@ protected IHttpTransportManager getHttpTransportManager() {
               .setLowLevelHttpResponse(expectedResponse)
               .build();
 
+          @Override
+          public String getName() {
+            return "scout.test-service-tunnel";
```
brunokoeferli: Maybe we should add a more specific domain.

e.g.
```suggestion
            return "scout.transport.test-service-tunnel";
```

or
```suggestion
            return "scout.http.test-service-tunnel";
```


More examples: `scout.transport.basic-auth`/`scout.http.basic-auth` and `scout.transport.default`/`scout.http.default`. I would favor `transport` over `http` as the corresponding attribute name (`http.client.name`) already contains `http` in its name.

brunokoeferli (approved):

org.eclipse.scout.rt.shared.test/src/test/java/org/eclipse/scout/rt/shared/servicetunnel/http/HttpServiceTunnelTest.java
paolobazzi: Change to:

- "scout.transport.basic-auth"
- "scout.transport.default"
- "scout.transport.service-tunnel"
- "scout.transport.test-service-tunnel"

PR #1140, opened 2024-09-05
Connection properties for REST HTTP client

paolobazzi:
Added config properties and REST client properties for:
- max connections in total
- max connections per route
- connection keep alive time
- connection validate after inactivity time

393566


paolobazzi:
Jenkins run tests

PR #1145, opened 2024-09-06
Fix incorrect parsing of HEX NCRs

Knicksaas:


Knicksaas (commented):
The commit message contains the wrong custom issue number 
Current: 394064
Correct: 370628

Change before merge

PR #1146, opened 2024-09-06
HttpProxy: Avoid causing another exception during exception handling

matthiaso:
Apparently throwing another exception while handling a previous exception (within failed) method may corrupt some connection pools [IOReactorShutdownException for following requests].

394532

### DIFF
--- a/org.eclipse.scout.rt.server.commons.test/src/test/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxyTest.java
+++ b/org.eclipse.scout.rt.server.commons.test/src/test/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxyTest.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import java.util.Random;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
@@ -55,6 +56,7 @@ import org.eclipse.jetty.server.handler.HandlerCollection;
 import org.eclipse.scout.rt.platform.BEANS;
 import org.eclipse.scout.rt.platform.IgnoreBean;
 import org.eclipse.scout.rt.platform.context.RunContexts;
+import org.eclipse.scout.rt.platform.internal.BeanInstanceUtil;
 import org.eclipse.scout.rt.platform.util.EnumerationUtility;
 import org.eclipse.scout.rt.platform.util.ImmutablePair;
 import org.eclipse.scout.rt.platform.util.ObjectUtility;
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
```diff
@@ -739,7 +739,14 @@ public void failed(Exception ex) {
...
           LOG.trace("Response is invalidated", e);
         }
-        asyncContext.complete();
+        try {
+          // this method is already called to handle an exception from within a http-client catch block
+          // it must not throw another exception again, otherwise caller may not handle this well again, it may even corrupt the whole connection pool
+          asyncContext.complete();
+        }
+        catch (Exception e) {
+          LOG.error("Unable to complete async context, assuming already completed", e);
```
paolobazzi: Maybe just LOG.warn? Should be nothing which the implementation could handle to avoid this errors?

org.eclipse.scout.rt.server.commons.test/src/test/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxyTest.java
```diff
@@ -576,18 +633,37 @@ public HttpServletResponse testProxyRequestInternal(Handler handler, long timeou
...
+    private final Supplier<AbstractAsyncHttpClientManager<?>> m_clientManagerSupplier;
     private final Function<Server, ServerConnector> m_serverConnectorFunction;
 
-    public HttpProxyTestParameter(AbstractAsyncHttpClientManager<?> clientManager, Function<Server, ServerConnector> serverConnectorFunction) {
-      m_clientManager = clientManager;
+    public HttpProxyTestParameter(Class<? extends AbstractAsyncHttpClientManager<?>> clientManagerClazz, Function<Server, ServerConnector> serverConnectorFunction) {
+      m_clientManagerClazz = clientManagerClazz;
+      m_clientManagerSupplier = () -> {
+        try {
+          return m_clientManagerClazz.getConstructor(new Class[]{}).newInstance();
```
paolobazzi: maybe use `BeanInstanceUtil.beanInstanceCreator(clientManagerClazz);` here?

paolobazzi (approved):

org.eclipse.scout.rt.server.commons.test/src/test/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxyTest.java
matthiaso: Good idea, thanks for the feedback, will be using this method.

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/servlet/HttpProxy.java
matthiaso: Changed to LOG.warn

PR #1157, opened 2024-09-09
Reduce number of CodeType to DataObject conversions

mvilliger:
Only convert CodeTypes to DataObjects if required:
- when requesting CodeTypes by Id using the CodeResource
- when creating UI notifications for CodeTypes invalidated in the cache

394632


org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/services/common/code/IApiExposedCodeTypeContributor.java
```diff
@@ -18,12 +18,12 @@
  * Contributor for CodeTypes that should be exposed by the Scout REST api.
```
samueleiben, line 18: Detail: Maybe we could move the part about the REST api to `IApiExposedCodeTypeDoProvider` as the resource now calls the provider instead of the contributor?

mvilliger: Basically yes, but still the ApiExposedCodeTypeContributor only returns CodeTypes which should be available on the UI via REST. So the comment is still valid for the contributor as well. Therefore i would keep it like that. Ok?

samueleiben: Fine by me 👍

PR #1162, opened 2024-09-09
Permission cache: invalidate on logout only if triggered by HTTP request

andibur:
The HTTP session timeout of the backend server is small compared to
the lifetime of the HTTP session on the UI server (default 5 minutes).
Whenever a backend session is invalidated due to inactivity, the
permissions of the affected user are also removed from the
AccessControlCache. This leads to a client notification and now also to
a UI notification. The UI server removes the user from its cache and the
web browser reloads the permissions. Both although the user is still
logged in and his permissions did not change.

It is correct that the AccessControlCache is invalidated in the event of
a dedicated logout by the user.

It is also desirable that a session timeout on the UI server leads to a
complete logout on the backend server, including invalidation of the
AccessControlCache.

394781


paolobazzi (approved):
Looks good!

PR #1163, opened 2024-09-09
HealthCheck: Log detailed warning in case health check is not successful

rogersteblerbsi:
361571

### DIFF
--- a/org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/healthcheck/AbstractHealthCheckServlet.java
+++ b/org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/healthcheck/AbstractHealthCheckServlet.java
@@ -22,6 +22,7 @@ import javax.servlet.http.HttpServletResponse;
 import org.eclipse.scout.rt.platform.BEANS;
 import org.eclipse.scout.rt.platform.Platform;
 import org.eclipse.scout.rt.platform.context.RunContext;
+import org.eclipse.scout.rt.platform.util.LazyValue;
 import org.eclipse.scout.rt.server.commons.servlet.AbstractHttpServlet;
 import org.eclipse.scout.rt.server.commons.servlet.HttpServletControl;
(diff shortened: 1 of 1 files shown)


org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/healthcheck/AbstractHealthCheckServlet.java
```diff
@@ -114,6 +114,10 @@ protected void doChecks(RunContext context, HttpServletRequest req, HttpServletR
       resp.getWriter().print(Platform.get().inDevelopmentMode() ? detailedOutput : output);
     }
     else {
+      if (statusCode != HttpServletResponse.SC_OK) {
+        String detailedOutput = generateOutput(statusCode, checks, failed, true);
+        LOG.warn(detailedOutput);
+      }
```
matthiaso: Definitely better than before, just one comment:

Do we really want to print a warning only if level is higher than _DEBUG_? If _DEBUG_ level is enabled, it would just be printed out as debug log (which might still be ignored, for example some people may only read _WARN_ and _ERROR_ and ignore anything else if there is a huge log volume). I would suggest always printing a warning if status does not equal 200.

For example (for the whole block):

```
    LazyValue<String> detailedOutput = new LazyValue<>(() -> generateOutput(statusCode, checks, failed, true));
    boolean isDevelopmentMode = Platform.get().inDevelopmentMode();
    if (statusCode != HttpServletResponse.SC_OK) {
      LOG.warn(detailedOutput.get());
    }
    else if (LOG.isDebugEnabled() || isDevelopmentMode) {
      LOG.debug(detailedOutput.get());
    }
    resp.getWriter().print(isDevelopmentMode ? detailedOutput.get() : output);
```

What do you think?

rogersteblerbsi: Good idea, I use your code now.

matthiaso:
Jenkins run tests

matthiaso:
Run tests

matthiaso (approved):

Issue #1176, opened 2024-09-11
RunContext.CURRENT.get().getTransaction() and ITransaction.CURRENT.get() mismatch


imotsch:
When nesting multiple RunContext chains the return value of calling
-   RunContext.CURRENT.get().getTransaction()
-   ITransaction.CURRENT.get()
is not consistent.

Also the copy algorithm regarding the RunContextm_transaction member is - by design and on purpose - different in
- RunContext.CURRENT.get().copy uses m_transaction=other.m_transcation
- RunContexts.coppyCurrent() uses m_transaction=ITransaction.CURRENT.get()

### Discrepancy
1. Whenever RunContext chains are nested one should always use RunContexts.coppyCurrent() in order to pass the current ITransaction to the child run context.

​This major difference in functionality is not reflected in the implementation code and not obvious in the javadoc.

2. In particular when using the transaction scope REQUIRES_NEW this difference can have major impacts.

```
**RunContexts.empty**().withScope(REQUIRES_NEW).run{
  // RunContext.CURRENT.get().getTransaction() is null due to RunContexts.empty()
  // ITransaction.CURRENT.get() is tx1 since a new ITranscation was created.

  // nesting with copyCurrent
  **RunContexts.copyCurrent**().withScope(REQUIRED).run{
    // all fine, OK
    // RunContext.CURRENT.get().getTransaction() is tx1 as set by copyCurrent()
    // ITransaction.CURRENT.get() is tx1 since no new ITranscation was created.
  }

  // nesting with RunContext.copy
  **RunContext.CURRENT.get().copy**().withScope(REQUIRED).run{
    // NOK
    // RunContext.CURRENT.get().getTransaction() is null as set by copy() that uses the parent RunContext.m_transaction that is never set.
    // ITransaction.CURRENT.get() is a new transcation tx2 since a new ITranscation had to be created. This is not intuitive and not the expectation of the programmer.
  }
}
```
### Improvements
1. The method RunContext.getTransaction() is only used internally by run context chains and should never be used by user code. --> Add javadoc and @Derecation marker to that method

2. On the static method RunContexts.copyCurrent and method RunContext.copy() --> Add javadoc to clarifiy the difference regarding transaction and what to use for nesting run context chains.

3. When using the scope REQUIRES_NEW at build time of a RunContext once the chain with thread locals is applied in TransactionProcessor#runTxRequiresNew add a statement to set the current RunContext m_transaction to the correct value by
[...]

PR #1177, opened 2024-09-11
RunContext.getTransaction vs. ITransaction

imotsch:
https://github.com/eclipse-scout/scout.rt/issues/1176 394954

### DIFF
--- a/org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/context/RunContextTest.java
+++ b/org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/context/RunContextTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -14,13 +14,16 @@ import static org.mockito.Mockito.*;
 
 import java.security.AccessController;
 import java.security.PrivilegedExceptionAction;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Locale;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 
 import javax.security.auth.Subject;
@@ -33,6 +36,7 @@ import org.eclipse.scout.rt.platform.job.Jobs;
 import org.eclipse.scout.rt.platform.logger.DiagnosticContextValueProcessor.IDiagnosticContextValueProvider;
 import org.eclipse.scout.rt.platform.nls.NlsLocale;
 import org.eclipse.scout.rt.platform.security.SimplePrincipal;
+import org.eclipse.scout.rt.platform.transaction.BasicTransaction;
 import org.eclipse.scout.rt.platform.transaction.ITransaction;
 import org.eclipse.scout.rt.platform.transaction.ITransactionMember;
 import org.eclipse.scout.rt.platform.transaction.TransactionScope;
@@ -46,11 +50,13 @@ import org.junit.runner.RunWith;
 import org.mockito.InOrder;
 import org.mockito.Mockito;
 import org.mockito.stubbing.Answer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 
 @RunWith(PlatformTestRunner.class)
 public class RunContextTest {
-
+  private static final Logger LOG = LoggerFactory.getLogger(RunContextTest.class);
   private static final ThreadLocal<String> THREAD_LOCAL = new ThreadLocal<>();
 
   @Test
@@ -60,6 +66,7 @@ public class RunContextTest {
     assertNull(runContext.getLocale());
     assertTrue(toSet(runContext.getPropertyMap().iterator()).isEmpty());
     assertNull(runContext.getCorrelationId());
+    //noinspection deprecation
     assertNull(runContext.getTransaction());
     assertEquals(TransactionScope.REQUIRED, runContext.getTransactionScope());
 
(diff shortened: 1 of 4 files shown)


org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/context/RunContextTest.java
```diff
@@ -803,6 +816,76 @@ public void testCopyCurrent_RunMonitorTwoConcurrentJobs() {
...
+          actual.add("3b " + ITransaction.CURRENT.get());
+        });
+
+        r3 = RunContext.CURRENT.get().copy()// using copy() instead of RunContexts.copyCurrent()
+            .withNewTransactionSupplier(newTx)
+            .withTransactionScope(TransactionScope.REQUIRED);
+        r3.run(() -> {
+          LOG.info("3c {} {}", RunContext.CURRENT.get().getTransaction(), ITransaction.CURRENT.get());
+          expected.add("3c tx2");
+          actual.add("3c " + ITransaction.CURRENT.get());// not that this ony is using tx1(m_transaction member) and not tx2 (ITransaction.CURRENT.get())
```
bschwarzent: Typos? (not, ony)

I don't quite understand the comment. Isn't it _supposed to be_ tx2 instead of tx1? (see line above, tx2 is expected)

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/context/RunContexts.java
```diff
@@ -77,6 +77,9 @@ public static RunContext empty() {
    * cancellation of the current (parent) monitor. Cancellation works top-down, so cancellation of the context's monitor
    * has no effect to the current (parent) monitor.
    * <p>
+   * {@link ITransaction}<br>
+   * Uses the current {@link ITransaction} as of <code>ITransaction.CURRENT.get()</code>.
+   * <p>
    * {@link TransactionScope}<br>
    * Uses the transaction scope {@link TransactionScope#REQUIRED} which starts a new transaction only if not running in
```
bschwarzent, line 85: > Uses the transaction scope {@link TransactionScope#REQUIRED}

This is not entirely correct. The value of the "transactionScope" property of the new instance depends on the RunContextFactory bean. This should be reformulated accordingly, or the JavaDoc should be overwritten as well on subclasses.

imotsch:
Build Success: https://jenkins-pe-ci.openshift.bsiag.local/view/pe-tests/job/pe-customtest/2433/

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/transaction/TransactionProcessor.java
```diff
@@ -105,6 +106,9 @@ protected RESULT runTxRequiresNew(final Chain<RESULT> chain) throws Exception {
       newTransaction.registerMember(transactionMember);
     }
 
+    // Update this RunContext's m_transaction field for consistency
+    RunContext.CURRENT.get().withTransaction(newTransaction);
```
bschwarzent: Input from @andibur: Consider restoring the previous value in the finally-block below.

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/context/RunContext.java
```diff
@@ -571,6 +576,10 @@ protected void fillEmpty() {
 
   /**
    * Creates a copy of <code>this</code> context.
+   * <p>
+   * <b>NOTE</b> use {@link RunContexts#copyCurrent()} when creating a nested RunContext inside a running chain. This
```
bschwarzent: Maybe we should explain the use case of this method a bit more in detail and clarify that it should not be used under normal circumstances.

Something like this:

`<b>Important:</b> Be careful when using this method! It should only be used to create a shallow copy of a run context that you have under control, e.g. a template context. Avoid using it to create a copy of the current run context, as simply copying may not make sense for all values (e.g. transaction scope). In most cases, it is recommended to use {@link RunContext#copyCurrent()} instead to create a new run context. Unlike {@link copy()}, it initializes the values according to the current thread state. For example, the transaction scope is reset to the default value of the active {@link RunContextFactory} and the transaction is taken from {@code ITransaction.CURRENT.get()}.`

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/context/RunContext.java
```diff
@@ -330,8 +331,12 @@ public RunContext withTransactionScope(final TransactionScope transactionScope)
   }
 
   /**
+   * @deprecated Use <code>ITransaction.CURRENT.get()</code> to access the current transaction. This method is only used
```
bschwarzent: > This method is only used at build time of a {@link RunContext} when a specific transaction is to be used by a RunContext.

This statement seems to apply to withTransaction(), not getTransaction().

Maybe reformulate:
`This member represents the transaction that was set during the build time of a run context, either explicitly or via {@link copy()}. It may differ from the actual transaction used during the execution of the runnable, e.g. when the transaction scope {@link TransactionScope.REQUIRES_NEW} is used.`

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/context/RunContexts.java
imotsch: done

org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/context/RunContextTest.java
imotsch: done

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/transaction/TransactionProcessor.java
imotsch: done

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/context/RunContext.java
imotsch: thank you, very good point, done.

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/context/RunContext.java
imotsch: done

PR #1179, opened 2024-09-11
Fix flip/flop HttpRetryTest

matthiaso:
testPostWithUnsupportedRetryAndFailureWhileRetrievingResponse

Testing retry scenario, the first POST request gets interrupted (e.g. network cable loose connection during response), the second request gets
 through. The test itself waited for the response of the second request
 and checked if both requests were received by the remote server (also
 locally set-up in this test). It seems like often the remote-server
 also works through the requests in order (and adds something to
 m_serverPostLog); however there are also multiple threads handling the
 incoming requests, the remote server may as well handle the requests
 out of order (e.g. first handle the second request, afterwards handle
 the first one). But as we only wait for the second request/response the
  servlet may not even have worked through the first request yet.
  Solution: Wait until both requests have been received at the servlet
  (timeout with 60 seconds in total if something went terribly wrong).

### DIFF
--- a/org.eclipse.scout.rt.server.test/src/test/java/org/eclipse/scout/rt/server/commons/http/HttpRetryTest.java
+++ b/org.eclipse.scout.rt.server.test/src/test/java/org/eclipse/scout/rt/server/commons/http/HttpRetryTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
  *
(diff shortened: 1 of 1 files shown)


matthiaso:
# Internal reminder: Do not forget to cherry-pick backward after merge!

bschwarzent:
Change looks good in principle. Some small remarks:

* I don't understand why there are 2 received POST requests. The request explicitly overwrites `retrySupported()` with false. But it seemed to work previously, so it probably is ok.
* Why is the `m_` prefix used for m_client and m_server, but not for the other members of the test class?
* Please align the first character of each line in the commit message:
grafik

PR #1187, opened 2024-09-17
AbstractHttpTransportManager ensure initialization is run only once

matthiaso:
395516

### DIFF
--- a/org.eclipse.scout.rt.shared.test/src/test/java/org/eclipse/scout/rt/shared/http/AbstractHttpTransportManagerTest.java
+++ b/org.eclipse.scout.rt.shared.test/src/test/java/org/eclipse/scout/rt/shared/http/AbstractHttpTransportManagerTest.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2010-2024 BSI Business Systems Integration AG.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     BSI Business Systems Integration AG - initial API and implementation
+ */
+package org.eclipse.scout.rt.shared.http;
+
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.shared.test/src/test/java/org/eclipse/scout/rt/shared/http/AbstractHttpTransportManagerTest.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+import org.junit.Test;
+
+import com.google.api.client.http.HttpRequestInitializer;
+
```
paolobazzi, line 28: Use Plattformtestrunner?

org.eclipse.scout.rt.shared.test/src/test/java/org/eclipse/scout/rt/shared/http/AbstractHttpTransportManagerTest.java
```diff
@@ -0,0 +1,64 @@
...
+public class AbstractHttpTransportManagerTest {
```
paolobazzi, line 30: How long does this test normally run? Would @SlowTest annotation make sense here?

paolobazzi (approved):

org.eclipse.scout.rt.shared.test/src/test/java/org/eclipse/scout/rt/shared/http/AbstractHttpTransportManagerTest.java
matthiaso: Don't think it is necessary for this specific test - however it does not cause any harm adding it here.

Added _PlatformTestRunner_, especially because it is not defined if initialization of the transport manager may at some point in time be using the platform (accessing bean manager), we might as well just add the runner now here.

PR #1188, opened 2024-09-18
AbstractHttpTransportManager ensure initialization is run only once

matthiaso:
395516


matthiaso:
See https://github.com/eclipse-scout/scout.rt/pull/1187

PR #1191, opened 2024-09-18
Form-State-XML: Return success flag for XML-import on forms

dwl-bsi:
The import of the state of a form via XML document is lenient and ignores unknown fields or fields with errors. Such errors are only logged but some systems need to know if a form state could be imported successfully or not. Add a boolean result type which indicates whether the XML-import was successful.

373731


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/form/IForm.java
```diff
@@ -595,17 +595,25 @@ public interface IForm extends IWidget, ITypeWithSettableClassId, IStyleable, ID
...
 
   void storeToXml(Element root);
 
-  void loadFromXml(Element root);
+  /**
+   * Imports the form state from an XML document.
+   *
+   * @param root
+   *          The XML document that contains the form state.
+   * @return True, if the XML document was loaded successfully. All fields and properties in the XML document which
```
jwinkens: `<code>true<true>`, am ende noch sowas wie,  `otherwise <code>false<code>`

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/form/AbstractForm.java
```diff
@@ -2348,7 +2354,17 @@ public void loadFromXml(Element root) {
...
-          f.loadFromXml(xField);
+          try {
+            success &= f.loadFromXml(xField);
+          }
+          catch (Exception e) {
+            LOG.warn("Could not load field {}", String.join(".", xmlFieldIds), e);
+            success = false; // error while loading value
+          }
+        }
+        else if (isXmlFieldWithStoredValue(xField)) {
```
jwinkens: Does it really matter if the field has a value or not? Can we reliably detect that the xml field has a value?

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/form/AbstractForm.java
```diff
@@ -2365,6 +2381,31 @@ public void loadFromXml(Element root) {
         }
       }
     }, ITabBox.class);
+    return success;
+  }
+
+  protected boolean isXmlFieldWithStoredValue(Element xField) {
+    // heuristic for inner forms: do not check recursively for field values. simply assume that an inner form has a value
+    return isXmlWrappedFormField(xField) ||
```
jwinkens: I am not sure if this is actually complete. Especially considering one can write their own special fields.

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/util/ObjectUtility.java
```diff
@@ -205,4 +207,30 @@ public static boolean isOneOf(Object o, Collection<?> elements) {
     }
     return false;
   }
+
+  /**
+   * @return {@code true} if the given object represents an actual value as opposed to an empty value such as null,
+   *         empty optional, empty collection, empty string etc.
+   */
+  public static boolean hasValue(Object o) {
```
jwinkens: I don't know if that is too great of an idea either. People are going to rely on it and probably don't expect a true here "just because we don't know". I'd much rather prefer we do not try to find out if an xml entry has a value or not.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/form/AbstractForm.java
jwinkens: This also seems to be a behaviour change. Previously fields that could not be resolved were just ignored (regardless of having a value or not), but now we consider it to be an error. We should probably check with someone if that is ok. We may consider refactoring this a bit and overwrite the relevant parts in our own Form implementation if we have stricter requirements to what is a successful load and what is not.

jwinkens (commented):
We are probably well-advised to get a second opinion on this. We may consider moving some of the heuristics/stricter requirements to our code base.

jwinkens:
Change looks better now. Javadoc still needs some fixes. There are multiple places where the `<code></code>` is missing.

PR #1193, opened 2024-09-20
Features/cgu/24.2/isd copy paste

cguglielmo:


eclipse-scout-core/src/desktop/DisplayChildController.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  acceptView(view: DisplayChild): boolean {
+    // Only render display child if displayParent is already rendered.
+    // If not, the child will be rendered once the displayParent is rendered.
+    return this.displayParent.rendered;
+  }
+
+  protected _register(child: DisplayChild) {
+    // use _registerChild in subclass
```
mvilliger, line 99: Maybe make the class and these methods abstract?

eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
```diff
@@ -288,8 +288,10 @@ export class PageWithTable extends Page implements PageWithTableModel {
   }
 
   protected _getLimitedResultInfoDo(tableData: any): LimitedResultInfoContributionDo {
-    return arrays.ensure(tableData?._contributions)
-      .find((contribution: DoEntity) => contribution._type === 'scout.LimitedResultInfoContribution') as LimitedResultInfoContributionDo;
+    if (!tableData) {
+      return null;
+    }
+    return dataObjects.getContribution('scout.LimitedResultInfoContribution', tableData) as LimitedResultInfoContributionDo;
```
mvilliger, line 291: Maybe it would be easier, if dataObjects.getContribution would itself check if tableData is null?

eclipse-scout-core/src/util/objects.ts
```diff
@@ -368,14 +368,32 @@ export const objects = {
   },
 
   /**
-   * @returns true if the given object is an object: no primitive type (number, string, boolean, bigint, symbol), no array, not null and not undefined.
+   * @deprecated The method was renamed to {@link isObject}. Use the new name or consider using {@link isPOJO} instead.
```
mvilliger: If we deprecate this method, we should also migrate all code to the new function. Otherwise lots of deprecated warnings will popup. Maybe only consider this deprecation for 25.1 and keep the existing 24.2 API stable?

eclipse-scout-core/src/util/objects.ts
```diff
@@ -368,14 +368,32 @@ export const objects = {
...
+  isObject<T>(obj: T): obj is Exclude<typeof obj, Primitive | undefined | null | T[]> {
     return typeof obj === 'object' &&
       !objects.isNullOrUndefined(obj) &&
       !Array.isArray(obj);
   },
 
+  /**
+   * @returns true if the given object is a plain old JavaScript object, which is an object created by the object literal notation ({}) or using `new Object()`.
+   */
+  isPOJO<T>(obj: T): obj is Exclude<typeof obj, Primitive | undefined | null | T[]> {
```
mvilliger: Specs for this new utility method are missing

eclipse-scout-core/src/dataobject/dataObjects.ts
```diff
@@ -0,0 +1,59 @@
...
+  removeContribution<TContributionDo extends DoEntity>(contributionClassOrType: DoContributionClassOrType<TContributionDo>, doEntity: DoEntityWithContributions) {
+    arrays.removeByPredicate(doEntity._contributions, contrib => predicate(contrib, contributionClassOrType));
+    if (doEntity._contributions?.length === 0) {
+      delete doEntity._contributions;
+    }
+  }
+};
+
+function predicate(contribution: DoEntity, contributionClassOrType: DoContributionClassOrType<DoEntity>): boolean {
+  if (typeof contributionClassOrType === 'string') {
```
mvilliger, line 61: this way the if in this predicate is executed for every element in the array, even though it is already known which condition must be used. Maybe extract this so that only once the condition type must be checked?

eclipse-scout-core/src/dataobject/dataObjects.ts
```diff
@@ -0,0 +1,59 @@
...
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+import {arrays, DoEntity, objects, scout} from '..';
+
+export const dataObjects = {
+  /**
```
mvilliger, line 14: Maybe the functions of this utility should be null safe?

eclipse-scout-core/src/desktop/DisplayChildController.ts
cguglielmo: FormController works a bit differently and does not need to implement them

eclipse-scout-core/src/util/objects.ts
cguglielmo: Yes, we should migrate existing code. I would rather remove this method with 25.1. completely

mvilliger: Fine with me

eclipse-scout-core/src/dataobject/dataObjects.ts
cguglielmo: I don't think it is worth it because there won't be many contributions

eclipse-scout-core/src/util/objects.ts
cguglielmo: I'll add the new function on 25.1 only and use $.isPlainObject now

eclipse-scout-core/src/util/objects.ts
cguglielmo: true, added for 25.1 branch

PR #1195, opened 2024-09-23
AccessControl.ts: let app handle bootstrap errors

cguglielmo:
With Microsoft Edge, if a tab is closed and reopened after the http session has expired, the index.html is loaded from cache without revalidating which starts the bootstrap requests. These requests will fail and may return 401 (Unauthorized). In that case, the app would trigger a page reload because the login site needs to be displayed. Unfortunately, the status code gets lost for the permission call, so the message "Permissions were not synchronized successfully." is shown and the user has to manually reload the page.

395126

### DIFF
--- a/eclipse-scout-core/src/App.ts
+++ b/eclipse-scout-core/src/App.ts
@@ -9,8 +9,8 @@
  */
 
 import {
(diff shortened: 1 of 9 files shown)


fschinkel (approved):

cguglielmo:
jenkins run tests

PR #1199, opened 2024-09-24
UI Callback Support

mvilliger:
Adds API to call functions on the UI (browser) and wait for the result.

389633


eclipse-scout-core/src/desktop/hybrid/uicallback/UiCallbacks.ts
```diff
@@ -0,0 +1,77 @@
...
+ */
+export interface UiCallbackHandler {
+  /**
+   * Called when a UiCallback is sent from the client backend.
+   * @param callbackId The ID of the callback.
+   * @param owner The owner {@link Widget} the call belongs to.
+   * @param request The optional callback request data as sent from the client backend. May be {@code null}.
+   * @returns A {@link JQuery.Promise} holding the result {@link DoEntity} in case it is resolved or the {@link UiCallbackErrorDo} in case it is rejected.
+   */
+  handle(callbackId: string, owner: Widget, request: DoEntity): JQuery.Promise<DoEntity, UiCallbackErrorDo>;
```
cguglielmo, line 70: {@code} and the @returns won't be shown correctly by IntelliJ when pressing ctrl+q. Use markdown syntax for code block.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/hybrid/uicallback/UiCallbacks.java
```diff
@@ -0,0 +1,480 @@
...
+    assertNotNull(owner);
+    assertModelThread();
+
+    if (StringUtility.isNullOrEmpty(callbackId)) {
+      callbackId = UUID.randomUUID().toString();
+    }
+
+    boolean isNewId;
+    P_UiCallback<R> callback;
+    synchronized (m_pendingUiCallbacks) {
```
cguglielmo, line 210: Why does it need to be synchronized? Isn't it always called from a model job?

mvilliger: So that the future can be used outside a ModelThread (e.g. when calling cancel)

PR #1200, opened 2024-09-25
Add shake animation when unchecking last visible calendar

Knicksaas:
When the user tries to hide the last visible calendar, the tree node shakes and is not uncheckable. This commit includes also a bugfix to correctly update the visibility of a calendar when the calendar is initialized but not rendered.

387614


Knicksaas:
Run tests

PR #1205, opened 2024-09-27
Calendar: incorrect scroll shadow when on month view

Knicksaas:
The calendar-scrollable-components class indicates divs, which are scrollable and are holding components. With the introduction of the resource calendar, the components are located in a div with class resource-column. The scrollable class was still on the parent div. This was fixed in this commit. Also, the full day index stack key calculation was not working. It should not separate the full day appointments when the display mode is week or workweek.

394666


Knicksaas:
Cherry-pick of #1168

PR #1207, opened 2024-09-30
Features/cgu/25.1/empty

cguglielmo:


eclipse-scout-core/src/form/fields/FormFieldValidationResultProvider.ts
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {fields, FormField, InitModelOf, ObjectModel, ObjectWithType, SomeRequired, Status, ValidationResult} from '../..';
+
+/*
```
mvilliger: Remove this doubled file header

eclipse-scout-core/src/form/fields/FormFieldValidationResultProvider.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+export class FormFieldValidationResultProvider implements ObjectWithType {
```
mvilliger: Maybe call this FormFieldValidator or something like that?

eclipse-scout-core/src/form/fields/FormFieldValidationResultProvider.ts
```diff
@@ -0,0 +1,51 @@
...
+  declare initModel: SomeRequired<this['model'], 'field'>;
+  field: FormField;
+  objectType: string;
+
+  init(model: InitModelOf<this>) {
+    $.extend(this, model);
+  }
+
+  provide(errorStatus: Status): ValidationResult {
+    const validByErrorStatus = !errorStatus || errorStatus.isValid();
```
mvilliger: Maybe off topic, but I just saw the following:
Currently FormField.errorStatus is public while FormField._errorStatus() is not. Isn't this wrong? Is it allowed to use FormField.errorStatus? Or should always FormField._errorStatus() be used?
Because currently there are lots of places directly using FormField.errorStatus. But what happens if the function would be overwritten (like e.g. in SmartField?)
Or should we make FormField.errorStatus (the field) protected but FormField._errorStatus() (the function) public instead? Because then, it would no longer be necessary to pass it here into the validator?

eclipse-scout-core/src/form/fields/FormFieldValidationResultProvider.ts
cguglielmo: It does actually not validate but only create the validation result based on the empty and error status properties. Because of that and because of the naming conflict with the actual validators of the value field, I would rather not name it validator.

eclipse-scout-core/src/form/fields/FormFieldValidationResultProvider.ts
cguglielmo: You are right, I saw that as well and it is not ideal. But instead of doing a refactoring I decided to pass it into the validation result provider.

PR #1209, opened 2024-10-01
Do not use filterMenuRec for Calendar menus

nsteger:
MenuUtility.filterMenusRec(List<IMenu>, Predicate<IMenu>) wrappes menus with child actions. In the case of the calendar, no wrapper menus should be used, since wrapper menus won't propagate the owner value change event.
Added new method MenuUtility.filterMenus(List<IMenu>, Predicate<IMenu>) which is used by the calendar instead

332126


nsteger:
Cherry Pick from https://github.com/eclipse-scout/scout.rt/pull/1126

PR #1211, opened 2024-10-02
Remove REST IApiVersion and implementations

paolobazzi:
- API version was not used active in REST resources and therefore was removed without replacement.
- Introduced IModuleBuildDateProvider allowing to provide a build date for a concrete server module which is offering application meta data using MetaResource.

 392974


org.eclipse.scout.rt.api/src/main/java/org/eclipse/scout/rt/api/meta/MetaResource.java
```diff
@@ -52,29 +53,34 @@ public class MetaResource implements IRestResource {
...
-        .withBuildDate(BEANS.get(JarManifestHelper.class).getBuildDateAttribute(apiVersion.getClass()));
+        .withBuildDate(getBuildDate());
   }
 
   @GET
   @Path("version")
   @Produces(MediaType.TEXT_PLAIN)
-  @ApiDocDescription(text = "Returns the API version in text format. If 'verbose' is true, the application name and version are returned as well.")
-  public String getApiVersion(@QueryParam("verbose") boolean verbose) {
+  @ApiDocDescription(text = "Returns the application version in text format. If 'verbose' is true, the application name and module build date are returned as well.\")")
```
fschinkel: copy-past-error in the end?

fschinkel (approved):

org.eclipse.scout.rt.api/src/main/java/org/eclipse/scout/rt/api/meta/MetaResource.java
paolobazzi: done

PR #1212, opened 2024-10-04
tileUtil: moveOtherTilesDown can lead to overlaps

cguglielmo:
A tile that spans columns must consider the bottom grid data of every spanned column, not just the first one.

395112

### DIFF
--- a/eclipse-scout-core/src/tile/tileUtil.ts
+++ b/eclipse-scout-core/src/tile/tileUtil.ts
@@ -113,7 +113,7 @@ export const tileUtil = {
             // Ignore tiles without explicit positions, they will be arranged automatically by the logical grid
             continue;
           }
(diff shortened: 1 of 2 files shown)


bschwarzent (commented):
Fix was tested successfully.

Is it necessary to expose `lowestGridData()` as a public function on tileUtil? It seems to be specially tailored to work with the bottomGridDataPerColumn structure, which only exists within `moveOtherTilesDown()`. It could therefore be written as inner function, like `updateBottomGridData()`. There are probably not a lot of other use cases that could make use of the function. If it _was_ public, it should get its own jasmine spec.

bschwarzent (approved):

cguglielmo:
It could be used elsewhere because it just takes a list of gridData objects and returns the lowest one, so it is a generic function. But  it is probably unlikely that anyone is needing it so I made it now private.

PR #1213, opened 2024-10-04
REST HTTP Client Keep Alive and socket/connect timeout NOK

paolobazzi:
397411

### DIFF
--- a/org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/client/RestClientProperties.java
+++ b/org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/client/RestClientProperties.java
@@ -29,8 +29,8 @@ public final class RestClientProperties {
   public static final String ENABLE_COOKIES = "scout.rest.client.enableCookies";
 
   /**
-   * Standard cookie specification used by underlying http client. See {@code org.apache.hc.client5.http.cookie.StandardCookieSpec}
-   * for details.
+   * Standard cookie specification used by underlying http client. See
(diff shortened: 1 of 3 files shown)


org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/client/RestClientProperties.java
```diff
@@ -265,6 +266,18 @@ public enum LoggerVerbosity {
...
   public static final String VALIDATE_CONNECTION_AFTER_INACTIVITY = "scout.rest.client.http.validateAfterInactivity";
 
+  /**
+   * Specifies the total span of time in milliseconds connections can be kept alive or execute requests. May be zero if the
+   * connection does not have an expiry deadline.
+   * <p>
+   * The value MUST be an instance convertible to {@link java.lang.Long}.
+   * </p>
+   * <p>
+   * The default value is 30 * 60 * 1000 (e.g. 30 minutes).
```
matthiaso: Is this correct? Isn't the default value defined by the config property (which could be set to another value)?

matthiaso (approved):

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/client/RestClientProperties.java
paolobazzi: Adjusted comment.

paolobazzi:
please run tests

PR #1217, opened 2024-10-07
getGrantedPermissionLevel: Ignore null attribute

matthiaso:
In 66924c59a7a8f899ee8ba51acf6f239300c19a1d a warning was introduced if getGrantedPermissionLevel was called with a null input; however this null input is valid after all, see javadoc on IPermissionCollection; remove incorrect warning.

397579


matthiaso:
See https://github.com/eclipse-scout/scout.rt/pull/1215

PR #1226, opened 2024-10-18
Add Java representation for Switch widget (ToggleSwitch)

bschwarzent:
Note: classes and packages use the name "toggle switch", because "switch" is a reserved keyword in Java.

398685


paolobazzi (approved):
Looks good!

PR #1231, opened 2024-10-21
UiNotifications: don't poll during tests

cguglielmo:
If a component subscribes for a ui notification topic, the ui notification polling starts. These poll requests will fail and new poll requests will be scheduled, which leads to a lot of noise.

During tests, ui notifications are now disabled by default. If a specific test needs it to be enabled, the mock can be removed by calling UiNotificationMock.unregister(). To not influence other specs, it should be registered again afterward.

Also:
- uiNotifications is now an instance of the class UiNotifications, so it can be replaced using the object factory.
- UiNotificationSystem can now be replaced using the object factory.

388703

### DIFF
--- a/eclipse-scout-core/src/uinotification/UiNotificationSystem.ts
+++ b/eclipse-scout-core/src/uinotification/UiNotificationSystem.ts
@@ -12,6 +12,7 @@ import {Event, EventHandler, EventListener, EventSupport, InitModelOf, ObjectMod
 export class UiNotificationSystem implements UiNotificationSystemModel, ObjectWithType {
   declare model: UiNotificationSystemModel;
   declare initModel: UiNotificationSystemModel;
+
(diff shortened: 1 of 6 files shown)


eclipse-scout-core/src/uinotification/UiNotificationSystem.ts
```diff
@@ -7,18 +7,25 @@
...
+  objectType: string;
 
-  constructor(name: string) {
-    this.name = name;
+  constructor() {
     this.events = new UiNotificationEventSupport(this);
   }
 
+  init(model: InitModelOf<this>) {
+    this.name = model.name;
```
fschinkel: Can the model be null or undefined? Is the name mandatory?

fschinkel (changes requested):

fschinkel (approved):

PR #1234, opened 2024-10-24
Tree: Do not send unnecessary events to server

Knicksaas:
- When Tree is not checkable, do not send any nodesChecked events
- Do not send empty nodesChecked events at all

### DIFF
--- a/eclipse-scout-core/src/tree/Tree.ts
+++ b/eclipse-scout-core/src/tree/Tree.ts
@@ -1257,25 +1257,40 @@ export class Tree extends Widget implements TreeModel, Filterable<TreeNode> {
   }
 
   protected _updateMarkChildrenChecked(node: TreeNode) {
-    let treeNodeUpdate = this._checkParentsRecursive(node);
+    let checkResult = this._checkParentsRecursive(node);
 
     if (!node.initialized) {
       // No rendering of nodes which have not been initialized yet
-      treeNodeUpdate.removeNode(node);
+      checkResult.removeNode(node);
     }
 
     // Trigger events and render
-    this._renderNodesChecked(treeNodeUpdate.getNodesForRendering());
-    this.trigger('nodesChecked', {
-      nodes: treeNodeUpdate.getNodesForEventTrigger()
-    });
+    this._processTreeCheckNodesResult(checkResult);
   }
 
-  protected _renderNodesChecked(treeNodes: TreeNode[]) {
-    treeNodes.forEach(node => {
-      node._renderChecked();
-      node._renderChildrenChecked();
-    });
+  /**
+   * Processes a {@link TreeCheckNodesResult} object. It renders all updated nodes and triggers
+   * a `nodesChecked` event.
+   *
+   * @param checkNodesResult object with tree node check update
+   * @param triggerEvent indicates whether events should be triggered or not. Default is true.
+   */
+  protected _processTreeCheckNodesResult(checkNodesResult: TreeCheckNodesResult, triggerEvent = true) {
+    // Render
+    if (this.rendered) {
+      checkNodesResult.getNodesForRendering().forEach(node => {
+        node._renderChecked();
+        node._renderChildrenChecked();
+      });
+    }
+
+    // Trigger event
+    let eventTriggerNodes = checkNodesResult.getNodesForEventTrigger();
+    if (this.checkable && triggerEvent && eventTriggerNodes.length) {
+      this.trigger('nodesChecked', {
+        nodes: eventTriggerNodes
+      });
+    }
   }
 
   protected _installNodeTooltipSupport() {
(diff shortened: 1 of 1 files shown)


eclipse-scout-core/src/tree/Tree.ts
```diff
@@ -1257,25 +1257,38 @@ export class Tree extends Widget implements TreeModel, Filterable<TreeNode> {
...
-      nodes: treeNodeUpdate.getNodesForEventTrigger()
-    });
+    this._processTreeCheckNodesResult(checkResult);
   }
 
-  protected _renderNodesChecked(treeNodes: TreeNode[]) {
-    treeNodes.forEach(node => {
+  /**
+   * Processes a {@link TreeCheckNodesResult} object. It renders all updated nodes and triggers
+   * an {@code nodesChecked} event to send updated nodes to java.
```
bschwarzent:
```suggestion
   * a `nodesChecked` event.
```

* an -> a
* The tree widget is not aware of the java part, it just emits the event and the TreeAdapter is responsible to forward that event to java.
* In TypeScript, we use backticks to format inline code in jsdoc

eclipse-scout-core/src/tree/Tree.ts
```diff
@@ -1257,25 +1257,38 @@ export class Tree extends Widget implements TreeModel, Filterable<TreeNode> {
...
+   *
+   * @param checkNodesResult object with tree node check update
+   * @param triggerEvent indicates, if events should be triggered or not
```
bschwarzent:
```suggestion
   * @param triggerEvent indicates whether events should be triggered or not. Default is true.
```

eclipse-scout-core/src/tree/Tree.ts
```diff
@@ -2614,17 +2627,8 @@ export class Tree extends Widget implements TreeModel, Filterable<TreeNode> {
...
 
-    // Trigger update event
-    if (opts.triggerNodesChecked && updatedNodes.getNodesForEventTrigger().length > 0) {
-      this.trigger('nodesChecked', {
-        nodes: updatedNodes.getNodesForEventTrigger()
-      });
-    }
-
-    // Render
-    if (this.rendered) {
```
bschwarzent, line 2625: Did you remove this check intentionally? It seems appropriate to keep it around the first part of `_processTreeCheckNodesResult`. Or can this only be called while the tree is rendered?

eclipse-scout-core/src/tree/TreeAdapter.ts
```diff
@@ -85,6 +85,11 @@ export class TreeAdapter extends ModelAdapter {
   }
 
   protected _sendNodesChecked(nodes: TreeNode[]) {
+    if (!this.widget.checkable) {
```
bschwarzent: Why is this flag checked here and not already in the Tree widget? It seems to make more sense not to trigger the `nodesChecked` event in the first place, e.g. in `_processTreeCheckNodesResult`.

eclipse-scout-core/src/tree/Tree.ts
```diff
@@ -1257,25 +1257,38 @@ export class Tree extends Widget implements TreeModel, Filterable<TreeNode> {
...
+  protected _processTreeCheckNodesResult(checkNodesResult: TreeCheckNodesResult, triggerEvent = true) {
+    // Render
+    checkNodesResult.getNodesForRendering().forEach(node => {
       node._renderChecked();
       node._renderChildrenChecked();
     });
+
+    // Trigger event
+    let eventTriggerNodes = checkNodesResult.getNodesForEventTrigger();
+    if (triggerEvent && eventTriggerNodes.length) {
```
bschwarzent: Is that really correct? What if the last checked node is unchecked? An observer of the tree would want to be notified about any change in the list of checked nodes, even when the new list is empty.

bschwarzent (changes requested):

eclipse-scout-core/src/tree/Tree.ts
Knicksaas: Yes, when the tree is initialized, the checked state of the nodes can already change. At this time, the tree is not rendered, but you have to notify the server about the state change.

eclipse-scout-core/src/tree/TreeAdapter.ts
Knicksaas: I agree. Otherwise, event listeners in TS would have been notified, but not those in java.

eclipse-scout-core/src/tree/Tree.ts
Knicksaas: The list of nodes does not contain all checked nodes, but all nodes which have encountered a checked-state change. When the last node was unchecked, the event holds the value: 
```json
"node": [{
  "id": "ui12",
  "...": "..."
  "checked": false
}]
```
An empty list means no nodes have changed its checked-state. Therefore, I believe this is correct.

eclipse-scout-core/src/tree/TreeAdapter.ts
bschwarzent: 👍

eclipse-scout-core/src/tree/Tree.ts
bschwarzent: Ah, I misunderstood the meaning of the `nodes` data. Thanks for the clarification!

bschwarzent:
Looks good!

bschwarzent (approved):

PR #1236, opened 2024-10-24
CalendarComponent: Allow empty toDate

Knicksaas:
When no toDate is provided, don't show a toDate in the UI and set the visual size of the component using defaultComponentDuration property

362615


org.eclipse.scout.rt.ui.html/src/main/resources/org/eclipse/scout/rt/ui/html/texts/Texts_it.properties
```diff
@@ -10,6 +10,7 @@
 ui.Active=Attivi
 ui.Aggregation=Aggregazione
 ui.All=Tutti
+ui.AtX=alle {0}
```
nsteger, line 13: Capital A? i.e. "Alle {0}"

Edit: no, but all other languages should also start with lowercase letters ;) (like from {0} to {0})

eclipse-scout-core/src/calendar/CalendarComponent.ts
```diff
@@ -50,15 +53,24 @@ export class CalendarComponent extends Widget implements CalendarComponentModel
...
   }
 
-  protected _syncCoveredDaysRange(coveredDaysRange: JsonDateRange) {
-    if (coveredDaysRange) {
-      this.coveredDaysRange = new DateRange(
-        dates.parseJsonDate(coveredDaysRange.from),
-        dates.parseJsonDate(coveredDaysRange.to));
+  protected _initCoveredDaysRange(coveredDaysRange: JsonDateRange | DateRange) {
+    if (coveredDaysRange && coveredDaysRange.from && !coveredDaysRange.to) {
+      coveredDaysRange.to = this._calculateDefaultToDate(coveredDaysRange.from);
```
nsteger: coveredDaysRange.from is a truncated date (no time), i.e. if appointment later or equal to 23:30:00 is inserted without end date, this wouldn't work correctly!

eclipse-scout-core/src/calendar/CalendarComponent.ts
```diff
@@ -25,17 +25,20 @@ export class CalendarComponent extends Widget implements CalendarComponentModel
...
 
   constructor() {
     super();
+    this.toDate = null;
     this.selected = false;
     this.fullDay = false;
     this.fullDayIndex = -1;
     this.item = null;
     this._$parts = [];
+    this.defaultComponentDuration = 30;
```
nsteger: maybe rename to defaultComponentDurationInMinutes ?

eclipse-scout-core/src/calendar/CalendarComponent.ts
Knicksaas: I have overseen that. Now it is fixed and when calculating the uiToDate, the component cannot overlap into the next day.

eclipse-scout-core/src/calendar/CalendarComponent.ts
Knicksaas: Done. I also changed the default duration to 60 minutes. Everywhere else, an event without toDate has a duration of one hour.

eclipse-scout-core/src/calendar/CalendarComponent.ts
```diff
@@ -57,16 +57,20 @@ export class CalendarComponent extends Widget implements CalendarComponentModel
   }
 
   protected _initCoveredDaysRange(coveredDaysRange: JsonDateRange | DateRange) {
-    if (coveredDaysRange && coveredDaysRange.from && !coveredDaysRange.to) {
-      coveredDaysRange.to = this._calculateDefaultToDate(coveredDaysRange.from);
-    } else if (!coveredDaysRange) {
-      coveredDaysRange = new DateRange(dates.ensure(this.fromDate), this.getUiToDate());
+    if (!coveredDaysRange || !coveredDaysRange.from) {
+      coveredDaysRange = new DateRange(dates.ensure(this.fromDate));
```
nsteger: can we have coveredDaysRange where only to is set? in this case the given data is lost here.. maybe use this instead:

if (!coveredDaysRange) {
  coveredDaysRange = new DateRange();
}
if (!coveredDaysRange.from) {
  coveredDaysRange.from = dates.ensure(this.fromDate);
}
if (!coveredDaysRange.to) {
  coveredDaysRange.to = this.toDate || coveredDaysRange.from;
}

eclipse-scout-core/src/calendar/CalendarComponent.ts
```diff
@@ -278,7 +282,12 @@ export class CalendarComponent extends Widget implements CalendarComponentModel
 
   protected _calculateDefaultToDate(fromDate: string | Date = this.fromDate) {
     fromDate = dates.ensure(fromDate);
-    fromDate.setMinutes(fromDate.getMinutes() + this.defaultComponentDuration);
+    let newMinutes = fromDate.getMinutes() + this.defaultComponentDurationInMinutes;
+    if (fromDate.getHours() === 23 && newMinutes >= 60) {
```
nsteger: this is not safe for all defaultComponentDurations. For example, if someone uses 120 minutes as default and the component has fromDate 22:30, the component will still overlap into next day.

PR #1243, opened 2024-10-28
Improve DeferredOperationQueue

fschinkel:
The queue processes elements immediately if there are more than its batch size. If there are fewer elements it waits for a specified time for more elements to come and processes a last batch after this delay. Then, if the queue is empty, the flush job completes and is started again after new elements are added to the queue.

In order to minimize the number of loops that are executed, check at the beginning of each loop if the next batch and therefore the whole queue is empty and complete the flush job if it is. Otherwise, the flush job will wait the specified delay unnecessarily before completing.

The job itself used the run context created by the specified supplier. This results in one run context for all batches that are executed. As a queue may run forever, this can lead to an enormous amount of executions that are not committed in the same transaction. In addition, note that the job itself does not really need a run context as it just schedules the batch executions, which are the elements here that need an own run context.
Therefore, use an empty run context for the flush job and the run context created by the supplier for the batch executions.

391689


andibur (approved):
Looks good, thx

PR #1253, opened 2024-11-05
IdFactoryTest: Invalid name for non-static of method

matthiaso:


matthiaso:
According to the class name / javadoc comment I would not expect a wrong named `of` method for this (test) class (see `FixtureStringId_WrongRawTypesOfMethodName` where I do expect the wrong name).

I suggest this fix only for 25.1+ (no tests fail after this change); if we do break this test on older releases it will be detected as soon as the corresponding change is cherry-picked for 25.1.

paolobazzi (approved):
lgtm

PR #1257, opened 2024-11-05
IdFactory: Fix fallback branch for findOfByTypesMethod

matthiaso:
IRootId does not have any generic parameter, however AbstractRootId does; use this class for fallback.

PR #1261, opened 2024-11-12
IdFactory: Fix fallback branch for findOfByTypesMethod

matthiaso:
IRootId does not have any generic parameter, however AbstractRootId does; use this class for fallback.


matthiaso:
See https://github.com/eclipse-scout/scout.rt/pull/1257

PR #1172, opened 2024-09-10
Improve logging for HttpProxy async requests

matthiaso:
Copy run monitor (or create a new one) to include correlation id and further diagnostics information.

374142

PR #1263, opened 2024-11-12
Improve logging for HttpProxy async requests

matthiaso:
* Copy run monitor (or create a new one) to include correlation id and further diagnostics information.

374142


matthiaso:
See https://github.com/eclipse-scout/scout.rt/pull/1172

PR #1264, opened 2024-11-13
Malware scanner with support for scanning single file.

christiandoebele:


org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/security/FileSystemMalwareScannerImplementor.java
```diff
@@ -69,8 +81,74 @@ public void scan(BinaryResource res) {
...
+        LOG.debug("Exception while checking for malware", e);
+      }
+      throwUnsafeResource(path);
+    }
+    finally {
+      IOUtility.deleteFile(tmpFile);
+    }
+  }
+
+  protected long compareFiles(Path path1, Path path2) throws IOException {
```
matthiaso, line 124: Is there any need for a long return value? Wouldn't `true`/`false` be sufficient?

matthiaso (approved):

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/security/FileSystemMalwareScannerImplementor.java
christiandoebele: Return type is similar to Files.mismatch (java API level 12). Custom implementation may be replaced by the standard implementation someday.

PR #1270, opened 2024-11-14
Improve StreamUtility concat

csehlke:
- StreamUtility#concat can now handle streams of different type similar to Stream#concat
- StreamUtility#concat is now null safe


avandorp:
String- or StreamUtility?

PR #1278, opened 2024-11-26
Update xml namespace -> jakarta

matthiaso:
com.sun.xml.ws.handler.HandlerChainsModel does assert naming.

400816


matthiaso:
https://github.com/eclipse-scout/scout.rt/pull/1275?new_mergebox=false

PR #1279, opened 2024-11-26
Add servlet handler extended resource lookup support

matthiaso:
400816


org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/Application.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
```
LukasHuser: We probably should update the license header to 2025 ;)

LukasHuser (approved):

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/Application.java
matthiaso:
```suggestion
 * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
```

PR #1280, opened 2024-11-27
Malware scanner with support for scanning single file.

matthiaso:


matthiaso:
https://github.com/eclipse-scout/scout.rt/pull/1264

PR #1281, opened 2024-11-27
Tree: update node size on updateNodes

andibur:
The size of a node must be recomputed when the text is changed and if scrollbars are used.

402635

### DIFF
--- a/eclipse-scout-core/src/tree/Tree.ts
+++ b/eclipse-scout-core/src/tree/Tree.ts
@@ -2408,6 +2408,15 @@ export class Tree extends Widget implements TreeModel {
       updatedNodes.push(oldNode);
     });
 
+    // recompute node size if horizontal scrolling is enabled
+    if (this.isHorizontalScrollingEnabled()) {
+      let nodesToResize = updatedNodes.filter(n => n.rendered);
+      if (arrays.hasElements(nodesToResize)) {
+        this._updateNodeSize(nodesToResize);
+        this.invalidateLayoutTree();
+      }
+    }
+
     this.trigger('nodesUpdated', {
       nodes: updatedNodes
     });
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/src/tree/Tree.ts
```diff
@@ -3056,6 +3065,10 @@ export class Tree extends Widget implements TreeModel {
   }
 
   protected _installNodes(nodes: TreeNode[]) {
+    this._updateNodeSize(nodes);
+  }
+
+  protected _updateNodeSize(nodes: TreeNode[]) {
```
cguglielmo, line 3071: Maybe name it updateNodeSizes (plural) because it takes an array of nodes. Also please move it up to the other update size related methods (updateNodeHeights, _updateNodeDimensions).

eclipse-scout-core/test/tree/TreeSpec.ts
```diff
@@ -598,6 +598,19 @@ describe('Tree', () => {
         expect($checkbox(child0).isEnabled()).toBe(false);
       });
     });
+
+    it('updates node size', () => {
+      tree.render();
+
+      let origWidth = node0.width;
+      let origMaxNodeWidth =tree.maxNodeWidth;
```
cguglielmo, line 606: Code does not look properly formatted

eclipse-scout-core/src/tree/Tree.ts
```diff
@@ -2408,6 +2408,15 @@ export class Tree extends Widget implements TreeModel {
       updatedNodes.push(oldNode);
     });
 
+    // recompute node size if horizontal scrolling is enabled
+    if (this.isHorizontalScrollingEnabled()) {
```
cguglielmo, line 2412: I remembered again that updateNodes is probably not the right method to adjust. Unfortunately, there is another method with a similar purpose: changeNode. changeNode should be called if a property changes that affects presentation but not the tree structure (see TreeEvent.TYPE_NODE_CHANGED). So we should add it there (to make it work for Scout Classic as well), but it works a bit differently... Unfortunately, it is not a batch method, instead it uses a microTask to coalesce the update. I tried to implement it here: https://github.com/eclipse-scout/scout.rt/tree/features/cgu/25.1/tree_update_node_size_402635 But it needs some more work.

I think in the future we should combine updateNodes and changeNode because it is confusing.

PR #1282, opened 2024-11-29
Features/cgu/25.1/isd undo redo

cguglielmo:


cguglielmo:
jenkins run tests

eclipse-scout-core/src/desktop/hybrid/DisposeWidgetsHybridActionDo.ts
```diff
@@ -0,0 +1,15 @@
...
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+import {DoEntity} from '../..';
+
+export interface DisposeWidgetsHybridActionDo extends DoEntity {
```
fschinkel, line 13: DOs for hybridAction need to have the _type property set, so maybe we should add it here.

fschinkel (changes requested):

eclipse-scout-core/src/desktop/hybrid/DisposeWidgetsHybridActionDo.ts
cguglielmo: This is not necessary at the moment, it will be added anyway when DOs will be migrated to classes

PR #1285, opened 2024-12-02
Add LookupColumn

fschinkel:
The column is a multivalued smart column that displays
a LookupBox (List or Tree) when editable.
The column supports distinct selection, i.e. the same value is not
selectable in more than one row.
Move shared code for SmartColumn and LookupColumn to LookupCallColumn.

Implement StaticLookupCall._getByKeys as it is used by e.g.
LookupColumn.formatValues.


eclipse-scout-core/src/lookup/StaticLookupCall.ts
```diff
@@ -187,6 +187,24 @@ export class StaticLookupCall<TKey> extends LookupCall<TKey> implements StaticLo
...
 
+  protected override _getByKeys(keys: TKey[]): JQuery.Promise<LookupResult<TKey>> {
+    this._deferred = $.Deferred();
+    setTimeout(() => this._queryByKeys(keys), this.delay);
+    return this._deferred.promise();
+  }
+
+  protected _queryByKeys(keys: TKey[]) {
+    const lookupRows = arrays.ensure(keys).map(key => this._lookupRowByKey(key)).filter(row => !!row);
+    if (lookupRows?.length) {
```
cguglielmo: How can lookupRows be undefined? LookupRows is always an array -> remove question mark

eclipse-scout-core/src/table/columns/LookupColumn.ts
```diff
@@ -0,0 +1,172 @@
...
+      return '';
+    }
+    lookupCall = lookupCall.clone();
+    prepareLookupCall?.(lookupCall);
+    return lookupCall.textsByKeys(values)
+      .then(result => strings.join(', ', ...Object.values(result)));
+  }
+}
+
+export interface LookupColumnModel<TValue> extends ColumnModel<TValue[]> {
```
cguglielmo: Add CodeType and adjust js docs according to https://github.com/eclipse-scout/scout.rt/pull/1286 (or I'll do it with the other PR)

eclipse-scout-core/src/table/columns/LookupColumn.ts
```diff
@@ -0,0 +1,172 @@
...
+   *
+   * Default is SmartField.DEFAULT_BROWSE_MAX_COUNT.
+   */
+  browseMaxRowCount?: number;
+  /**
+   * Whether the same value can be selected in multiple rows or not.
+   *
+   * Default is `false`.
+   */
+  distinct?: boolean;
```
cguglielmo: Better create a separate file for Model and EventMap so it is consistent between all scout columns

eclipse-scout-core/src/table/columns/LookupColumn.ts
```diff
@@ -0,0 +1,172 @@
...
+  protected _setLookupCall(lookupCall: LookupCallOrModel<TValue>) {
+    let call = LookupCall.ensure(lookupCall, this.session);
+    this._setProperty('lookupCall', call);
+  }
+
+  setBrowseHierarchy(browseHierarchy: boolean) {
+    this.setProperty('browseHierarchy', browseHierarchy);
+  }
+
+  protected _setBrowseHierarchy(browseHierarchy: boolean) {
```
cguglielmo: the protected setters are not necessary if they only contain a call to this._setProperty -> can be removed

eclipse-scout-core/src/table/columns/LookupColumn.ts
```diff
@@ -0,0 +1,172 @@
...
+      this.table.rows.forEach(r => {
+        if (r === row) {
+          return;
+        }
+        valuesInOtherRows.push(...arrays.ensure(this.cellValue(r)));
+      });
+      arrays.removeAll(valuesInOtherRows, this.cellValue(row));
+      e.result.lookupRows.forEach(lookupRow => {
+        if (scout.isOneOf(lookupRow.key, valuesInOtherRows)) {
+          lookupRow.enabled = false;
```
cguglielmo: Add missing setter setEnabled() on LookupRow and call it

eclipse-scout-core/src/table/columns/LookupColumn.ts
```diff
@@ -0,0 +1,172 @@
...
+      row
+    }));
+    editor.on('lookupCallDone', e => this.trigger('lookupCallDone', {
+      result: e.result
+    }));
+
+    return editor;
+  }
+
+  override isContentValid(row: TableRow): { valid: boolean; validByMandatory: boolean; errorStatus: Status } {
```
cguglielmo: Add new type ColumnValidationResult for the return type to reduce copy & paste

eclipse-scout-core/src/table/columns/LookupEditor.ts
```diff
@@ -0,0 +1,192 @@
...
+    const lookupBox = this._popup.widget('LookupBox', LookupBox<TValue>);
+
+    // sync value
+    lookupBox.on('propertyChange:value', e => this.setValue(e.newValue));
+
+    // pass popup events to editor
+    lookupBox.on('prepareLookupCall', e => this.trigger('prepareLookupCall', {
+      lookupCall: e.lookupCall
+    }));
+
```
cguglielmo, line 98: Remove blank line (belongs to block "pass popup events to editor") and format both event handlers the same way.

eclipse-scout-core/src/table/columns/LookupEditor.ts
```diff
@@ -0,0 +1,192 @@
...
+}
+
+export interface LookupEditorEventMap<TValue> extends ValueFieldEventMap<TValue[]> {
+  'lookupCallDone': LookupEditorLookupCallDoneEvent<TValue>;
+  'prepareLookupCall': LookupEditorPrepareLookupCallEvent<TValue>;
+}
+
+export class LookupEditorTreeLayout extends TreeLayout {
+
+  constructor(tree: Tree) {
```
cguglielmo: Constructor should not be necessary

eclipse-scout-core/src/table/columns/LookupEditor.ts
```diff
@@ -0,0 +1,192 @@
...
+    super(tree);
+  }
+
+  override preferredLayoutSize($container: JQuery, options?: HtmlCompPrefSizeOptions): Dimension {
```
cguglielmo: Explain that we need this to get the preferred width of the tree nodes. A fixed height limits the view range which would only render the first few nodes.

cguglielmo (changes requested):
Commit message
- Subject should be an action, not just a noun -> Add LookupColumn
- Remove 'Add LooupupColumn.' 
- Seperate last sentence with a blank line and explain why getByKeys had to be implemented.

Testing results
- The remove animation does not seem to work, popup is closed immediately
- The cell flickers when the editor opens and closes. This is not the case with the SmartColumn
- TAB and Shift TAB does not work when the editor is open, it should accept the value and focus the next cell
- I somehow expect ENTER to do something, for smart and date column it accepts the input and closes the popup. I think it should close the editor like ESC. 

Other issues that need to be tracked seperately 
- filtering a tree should not remove the child nodes if a parent node accepts the filter. E.g. filter for latin -> it is not possible to select any value

PR #1288, opened 2024-12-03
Calendar: Fix layouting when not rendered

Knicksaas:
When the calendar is not rendere, no layouting should be triggered when setting new components.

401857


Knicksaas:
Cherry pick of #1287

PR #1289, opened 2024-12-03
HybridManager: add support for context elements in hybrid events

bschwarzent:
The "data" attribute on hybrid events can hold static information. But sometimes, it would be helpful if we could reference a widget (e.g. Tree) or any other model element (e.g. TreeNode). Since the IDs of those elements are only known in the JSON layer, we cannot send them as part of the "data" object

A new property "contextElements" is introduced for that purpose. It can hold named lists of context elements. A context element consists of a widget and optionally any other element owned by that widget. In JSON form, a widget is represented by its JSON adapter ID. The optional element is represented by a widget-specific ID. These transformations are automatically applied by the JSON layer using an extensible list of converters (HybridActionContextElementConverter).

Other notable changes:
[HybridManagerAdapter.ts]
  Rename HybridEvent -> HybridRemoteEvent
[HybridManager.ts]
  Remove deprecated methods
[JsonHybridManagerTest.java]
  Simplify existing test cases. Use DataObjectMapper instance instead
  of custom mock logic. This used to be necessary, because no
  implementation existed at runtime, but that has changed since.
[UiSessionMock.java]
  Correctly register spy instance in the JSON adapter registry.
  Otherwise, resolving an adapter by ID might return an unexpected
  instance (the original adapter instead of the spy).
[scout.ts]
  Add Constructor type.

389630


eclipse-scout-core/src/desktop/hybrid/HybridActionContextElement.ts
```diff
@@ -0,0 +1,82 @@
...
+   */
+  static of(widget: Widget, element?: any): HybridActionContextElement {
+    return scout.create(HybridActionContextElement, {
+      widget,
+      element
+    });
+  }
+
+  constructor() {
+    this.widget = null;
```
fschinkel: initialize directly and remove constructor

eclipse-scout-core/src/desktop/hybrid/HybridActionContextElement.ts
```diff
@@ -0,0 +1,82 @@
...
+    let widget = scout.assertValue(this.widget);
+    return widgetType ? scout.assertInstance(widget, widgetType) : scout.assertInstance(widget, Widget) as W;
+  }
+
+  /**
+   * @returns the typed or untyped {@link element}, never `null`
+   * @throws Error if the element is `null` or not an instance of the given type. Use {@link optElement} to return `null` instead.
+   */
+  getElement<E>(elementType?: Constructor<E>): E {
+    let element = scout.assertValue(this.element);
```
fschinkel:
```suggestion
    return scout.assertValue(this.optElement(elementType));
```

eclipse-scout-core/src/desktop/hybrid/HybridActionContextElements.ts
```diff
@@ -0,0 +1,115 @@
...
+  isEmpty(): boolean {
+    return this.map.size === 0;
+  }
+
+  /**
+   * Returns a list of context elements for the given key. If the given key does not exist, an error is thrown.
+   * Use {@link optList} to return `null` instead. Use {@link getSingle} if only a single context element is expected.
+   */
+  getList(key: string): HybridActionContextElement[] {
+    let contextElements = this.optList(key);
```
fschinkel: why do we not use assertValue here?

eclipse-scout-core/src/desktop/hybrid/HybridActionContextElements.ts
```diff
@@ -0,0 +1,115 @@
...
+  optList(key: string): HybridActionContextElement[] {
+    return this.map.get(key);
+  }
+
+  /**
+   * Returns the context elements for the given key. If the given key does not exist, an error is thrown.
+   * Use {@link optSingle} to return `null` instead. Use {@link getList} if more than one context element is expected.
+   */
+  getSingle(key: string): HybridActionContextElement {
+    let contextElement = this.optSingle(key);
```
fschinkel: why do we not use assertValue here?

eclipse-scout-core/src/desktop/hybrid/converter/HybridActionContextElementConverter.ts
```diff
@@ -0,0 +1,87 @@
...
+  static all(): HybridActionContextElementConverter[] {
+    return HybridActionContextElementConverter._registry.map(registration => {
+      registration.converter = registration.converter || scout.create(registration.converterConstructor);
+      return registration.converter;
+    });
+  }
+
+  static convertFromJson(adapter: ModelAdapter, jsonElement: any): HybridActionContextElement {
+    let converters = HybridActionContextElementConverter.all();
+    for (let i = 0; i < converters.length; i++) {
```
fschinkel:
```suggestion
    for (let converter of converters) {
```

eclipse-scout-core/src/desktop/hybrid/converter/HybridActionContextElementConverter.ts
```diff
@@ -0,0 +1,87 @@
...
+      if (element) {
+        return element;
+      }
+    }
+    return null;
+  }
+
+  static convertToJson(adapter: ModelAdapter, element: any): JsonHybridActionContextElement {
+    let converters = HybridActionContextElementConverter.all();
+    for (let i = 0; i < converters.length; i++) {
```
fschinkel:
```suggestion
    for (let converter of converters) {
```

eclipse-scout-core/src/desktop/hybrid/converter/TreeNodeContextElementConverter.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+import {HybridActionContextElementConverter, ModelAdapter, objects, TreeAdapter, TreeNode} from '../../../index';
+
+export class TreeNodeContextElementConverter extends HybridActionContextElementConverter {
+
+  override jsonToElement(adapter: ModelAdapter, jsonElement: any): any {
+    if (adapter instanceof TreeAdapter && objects.isString(jsonElement)) {
+      let tree = adapter.widget;
+      let treeNode = tree.nodesMap[jsonElement];
+      if (!treeNode) {
```
fschinkel: why do we not use assertValue here?

eclipse-scout-core/src/desktop/hybrid/converter/HybridActionContextElementConverter.ts
```diff
@@ -0,0 +1,87 @@
...
+ * instance typically only handles a specific type of element. Because elements have no common structure, they are
+ * always accompanied by the widget that owns them (here represented by the corresponding {@link ModelAdapter}).
+ *
+ * How to add a converter:
+ * - Create a subclass of this abstract class.
+ * - Implement the abstract methods {@link #jsonToElement} and {@link #elementToJson}. Return `null` if the
+ *   converter cannot handle the given arguments (in which case the task is delegated to the next converter).
+ * - Register the class using the static {@link register} method.
+ */
+export abstract class HybridActionContextElementConverter {
```
fschinkel: Do we want to separate the registry from the converter, e.g. util for all static methods + abstract class?

eclipse-scout-core/src/desktop/hybrid/converter/HybridActionContextElementConverter.ts
```diff
@@ -0,0 +1,87 @@
...
+    return null;
+  }
+
+  /**
+   * @param adapter owner widget (e.g. TreeAdapter)
+   * @param jsonElement JSON representation of the element (e.g. String)
+   * @return model representation of the element (e.g. TreeNode), or `null` if this instance cannot handle the
+   * given arguments (in which case the task is delegated to the next converter)
+   */
+  abstract jsonToElement(adapter: ModelAdapter, jsonElement: any): any;
```
fschinkel: Do we want to add some generics and accept-functions? The registry then calls the accept-functions first and continues with the next converter if it does not accept. We still can pass it to the next converter if null is returned.

eclipse-scout-core/src/desktop/hybrid/HybridManager.ts
```diff
@@ -140,42 +141,48 @@ export class HybridManager extends Widget {
...
+   * @returns a promise that will be resolved with the entire result object once the corresponding hybridActionEnd event arrives.
    * @see IHybridAction
    * @see AbstractHybridAction.fireHybridActionEndEvent
    */
-  callActionAndWait(actionType: string, data?: object): JQuery.Promise<object> {
-    const id = this.callAction(actionType, data);
-    return this.when(`hybridActionEnd:${id}`).then(event => event.data);
+  callActionAndWaitWithContext(actionType: string, data?: object, contextElements?: HybridActionContextElements): JQuery.Promise<HybridManagerActionEndEventResult> {
+    const id = this.callAction(actionType, data, contextElements);
+    return this.when(`hybridActionEnd:${id}`).then(event => {
```
fschinkel: .then(event => ({data: ...

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
```diff
@@ -39,12 +40,111 @@ export class HybridManagerAdapter extends ModelAdapter {
   }
 
   protected _onWidgetHybridAction(event: HybridActionEvent) {
-    this._send('hybridAction', event.data);
+    this._send('hybridAction', {
+      actionType: event.data.actionType, // add as first property (devtools sometimes show properties in that order)
+      id: event.data.id,
+      data: event.data.data,
+      contextElements: this._contextElementsToJson(event.data.contextElements) || undefined
```
fschinkel, line 47: why is `|| undefined` necessary?

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
```diff
@@ -39,12 +40,111 @@ export class HybridManagerAdapter extends ModelAdapter {
...
+    };
   }
 }
 
-interface HybridEvent<TObject = object> extends RemoteEvent {
+interface HybridRemoteEvent<TObject = object> extends RemoteEvent {
   id: string;
   eventType: string;
   data: TObject;
+  contextElements: object;
```
fschinkel:
```suggestion
    contextElements: { [property: string]: JsonHybridActionContextElement[] };
```

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
```diff
@@ -39,12 +40,111 @@ export class HybridManagerAdapter extends ModelAdapter {
...
+    this._send('hybridAction', {
+      actionType: event.data.actionType, // add as first property (devtools sometimes show properties in that order)
+      id: event.data.id,
+      data: event.data.data,
+      contextElements: this._contextElementsToJson(event.data.contextElements) || undefined
+    });
+  }
+
+  protected _jsonToContextElements(jsonContextElements: object): HybridActionContextElements {
+    if (!jsonContextElements) {
```
fschinkel, line 52: when do we want to return null? only input is null or undefined? or also if it is an empty object or all keys map to empty arrays?

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
```diff
@@ -39,12 +40,111 @@ export class HybridManagerAdapter extends ModelAdapter {
...
+    Object.keys(jsonContextElements).forEach(key => {
+      let list = this._jsonToContextElementList(jsonContextElements[key]);
+      contextElements.withElements(key, list);
+    });
+    return contextElements;
+  }
+
+  protected _jsonToContextElementList(jsonContextElements: JsonHybridActionContextElement[]): HybridActionContextElement[] {
+    if (!jsonContextElements) {
+      return null;
```
fschinkel, line 65: when do we want to return null? see above

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
```diff
@@ -39,12 +40,111 @@ export class HybridManagerAdapter extends ModelAdapter {
...
+    });
+    return list;
+  }
+
+  protected _jsonToContextElement(jsonContextElement: JsonHybridActionContextElement): HybridActionContextElement {
+    if (!jsonContextElement) {
+      return null;
+    }
+    let adapter = this.session.getModelAdapter(jsonContextElement.widget);
+    if (!(adapter instanceof ModelAdapter)) {
```
fschinkel: is it possible that we get something that is not a ModelAdapter? maybe use assertValue or assertInstance

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
```diff
@@ -39,12 +40,111 @@ export class HybridManagerAdapter extends ModelAdapter {
...
+    let element = null;
+    if (jsonContextElement.element) {
+      element = HybridActionContextElementConverter.convertFromJson(adapter, jsonContextElement.element);
+      if (!element) {
+        throw new Error('Unable to convert JSON to context element');
+      }
+    }
+    return scout.create(HybridActionContextElement, {
+      widget: adapter.widget,
+      element: element
```
fschinkel:
```suggestion
   element
```

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
```diff
@@ -39,12 +40,111 @@ export class HybridManagerAdapter extends ModelAdapter {
...
+    if (!jsonContextElement) {
+      return null;
+    }
+    let adapter = this.session.getModelAdapter(jsonContextElement.widget);
+    if (!(adapter instanceof ModelAdapter)) {
+      throw new Error(`No adapter found for '${jsonContextElement.widget}'`);
+    }
+    let element = null;
+    if (jsonContextElement.element) {
+      element = HybridActionContextElementConverter.convertFromJson(adapter, jsonContextElement.element);
```
fschinkel: assertValue?

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
```diff
@@ -39,12 +40,111 @@ export class HybridManagerAdapter extends ModelAdapter {
...
+      element: element
+    });
+  }
+
+  protected _contextElementsToJson(contextElements: HybridActionContextElements): object {
+    if (!contextElements || contextElements.isEmpty()) {
+      return null;
+    }
+    let json = {};
+    contextElements.map.forEach((list, key) => {
```
fschinkel: order of arguments seems wrong at first, maybe for...of is easier to read
```suggestion
    for (const [key, list] of contextElements.map) {
```

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
```diff
@@ -39,12 +40,111 @@ export class HybridManagerAdapter extends ModelAdapter {
...
+    }
+    return scout.create(HybridActionContextElement, {
+      widget: adapter.widget,
+      element: element
+    });
+  }
+
+  protected _contextElementsToJson(contextElements: HybridActionContextElements): object {
+    if (!contextElements || contextElements.isEmpty()) {
+      return null;
```
fschinkel, line 104: see above

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
```diff
@@ -39,12 +40,111 @@ export class HybridManagerAdapter extends ModelAdapter {
...
+    let json = {};
+    contextElements.map.forEach((list, key) => {
+      json[key] = this._contextElementListToJson(list);
+    });
+    return json;
+  }
+
+  protected _contextElementListToJson(contextElements: HybridActionContextElement[]): JsonHybridActionContextElement[] {
+    if (!contextElements) {
+      return null;
```
fschinkel, line 115: see above

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
```diff
@@ -39,12 +40,111 @@ export class HybridManagerAdapter extends ModelAdapter {
...
+    }
+    return contextElements.map(contextElement => this._contextElementToJson(contextElement));
+  }
+
+  protected _contextElementToJson(contextElement: HybridActionContextElement): JsonHybridActionContextElement {
+    if (!contextElement) {
+      return null;
```
fschinkel, line 122: see above

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
```diff
@@ -39,12 +40,111 @@ export class HybridManagerAdapter extends ModelAdapter {
...
+
+  protected _jsonToContextElement(jsonContextElement: JsonHybridActionContextElement): HybridActionContextElement {
+    if (!jsonContextElement) {
+      return null;
+    }
+    let adapter = this.session.getModelAdapter(jsonContextElement.widget);
+    if (!(adapter instanceof ModelAdapter)) {
+      throw new Error(`No adapter found for '${jsonContextElement.widget}'`);
+    }
+    let element = null;
```
fschinkel: assignment not necessary

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
```diff
@@ -39,12 +40,111 @@ export class HybridManagerAdapter extends ModelAdapter {
...
+
+  protected _contextElementToJson(contextElement: HybridActionContextElement): JsonHybridActionContextElement {
+    if (!contextElement) {
+      return null;
+    }
+    let adapter = contextElement.widget.modelAdapter;
+    if (!(adapter instanceof ModelAdapter)) {
+      throw new Error('Widget does not have a model adapter');
+    }
+    let jsonElement = undefined;
```
fschinkel: assignment not necessary

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
```diff
@@ -39,12 +40,111 @@ export class HybridManagerAdapter extends ModelAdapter {
...
+    }
+    return contextElements.map(contextElement => this._contextElementToJson(contextElement));
+  }
+
+  protected _contextElementToJson(contextElement: HybridActionContextElement): JsonHybridActionContextElement {
+    if (!contextElement) {
+      return null;
+    }
+    let adapter = contextElement.widget.modelAdapter;
+    if (!(adapter instanceof ModelAdapter)) {
```
fschinkel: assertValue or assertInstance?

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
```diff
@@ -39,12 +40,111 @@ export class HybridManagerAdapter extends ModelAdapter {
...
+      throw new Error('Widget does not have a model adapter');
+    }
+    let jsonElement = undefined;
+    if (contextElement.element) {
+      jsonElement = HybridActionContextElementConverter.convertToJson(adapter, contextElement.element);
+      if (!jsonElement) {
```
fschinkel: assertValue?

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/hybrid/AbstractHybridAction.java
```diff
@@ -78,9 +126,7 @@ protected void addWidget(IWidget widget) {
    */
   protected void addWidgets(Map<String, ? extends IWidget> widgets) {
     assertInitialized();
-    hybridManager().addWidgets(Optional.ofNullable(widgets).orElse(Map.of())
-        .entrySet().stream()
-        .collect(Collectors.toMap(entry -> getId() + entry.getKey(), Entry::getValue)));
+    hybridManager().addWidgets(Optional.ofNullable(widgets).orElse(Map.of()).entrySet().stream().collect(Collectors.toMap(entry -> getId() + entry.getKey(), Entry::getValue)));
```
fschinkel: explicitly formatted?

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/hybrid/AbstractHybridAction.java
```diff
@@ -96,10 +142,7 @@ protected void removeWidget() {
    */
   protected void removeWidgetsById(Collection<String> ids) {
     assertInitialized();
-    hybridManager().removeWidgetsById(Optional.ofNullable(ids).orElse(Collections.emptySet())
-        .stream()
-        .map(id -> getId() + id)
-        .collect(Collectors.toSet()));
+    hybridManager().removeWidgetsById(Optional.ofNullable(ids).orElse(Collections.emptySet()).stream().map(id -> getId() + id).collect(Collectors.toSet()));
```
fschinkel: explicitly formatted?

eclipse-scout-core/src/desktop/hybrid/HybridActionContextElement.ts
```diff
@@ -0,0 +1,82 @@
...
+    scout.assertProperty(model, 'widget', Widget);
+    $.extend(this, model);
+  }
+
+  /**
+   * @returns the typed {@link widget}, never `null`
+   * @throws Error if the widget is not an instance of the given type
+   */
+  getWidget(): Widget;
+  getWidget<W extends Widget>(widgetType: Constructor<W>): W;
```
fschinkel, line 46: why are the overload necessary?

org.eclipse.scout.rt.client.test/src/test/java/org/eclipse/scout/rt/client/ui/desktop/hybrid/HybridActionContextElementsTest.java
```diff
@@ -0,0 +1,154 @@
...
+    IForm form1 = new P_DummyForm("form1");
+    IForm form2 = new P_DummyForm("form2");
+    IForm form3 = new P_DummyForm("form3");
+    ITree tree = new AbstractTree() {
+    };
+    ITreeNode treeNode = new AbstractTreeNode() {
+    };
+
+    Map<String, List<HybridActionContextElement>> baseMap = Map.of(
+        "zero", Collections.emptyList(),
```
fschinkel:
```suggestion
        "zero", List.of(),
```

org.eclipse.scout.rt.client.test/src/test/java/org/eclipse/scout/rt/client/ui/desktop/hybrid/HybridActionContextElementsTest.java
```diff
@@ -0,0 +1,154 @@
...
+    HybridActionContextElements contextElements = BEANS.get(HybridActionContextElements.class);
+    assertTrue(contextElements.isEmpty());
+
+    contextElements
+        .withElement("existing", HybridActionContextElement.of(form0))
+        .withElement("one", HybridActionContextElement.of(form0))
+        .withMap(baseMap)
+        .withElement("three", tree)
+        .withElement("four", tree, treeNode)
+        .withElements("five", Collections.emptyList())
```
fschinkel:
```suggestion
        .withElements("five", List.of())
```

org.eclipse.scout.rt.ui.html.test/src/test/java/org/eclipse/scout/rt/ui/html/json/desktop/hybrid/JsonHybridManagerTest.java
```diff
@@ -283,7 +266,265 @@ public void testCreateAndDisposeHybridWidget() {
...
+
+    JsonTestUtility.endRequest(m_uiSession);
+
+    // -------------------
+
+    String id1 = createId();
+    hybridManager.fireHybridEvent(id1, "foo", BEANS.get(HybridActionContextElements.class)
+        .withElement("form", form)
+        .withElement("node", tree, treeNode)
+    );
```
fschinkel: move to previous line

fschinkel (changes requested):

eclipse-scout-core/src/desktop/hybrid/HybridActionContextElement.ts
bschwarzent: Done (also for HybridAcitonContextElements.ts)

eclipse-scout-core/src/desktop/hybrid/HybridActionContextElement.ts
bschwarzent: To allow the argument `widgetType` to be optional, without allowing unchecked type inference:

- `element.getWidget()` --> should return type `Widget` and nothing else
- `element.getWidget(T)` --> should return type `T`, where T must be a subtype of Widget

For example, the following should not be allowed: `let form: Form = element.getWidget();`

eclipse-scout-core/src/desktop/hybrid/HybridActionContextElement.ts
bschwarzent: assertValue() will fail if the value is null, therefore the ternary expression on the next line is needed.

eclipse-scout-core/src/desktop/hybrid/HybridActionContextElements.ts
bschwarzent: Done

eclipse-scout-core/src/desktop/hybrid/HybridActionContextElements.ts
bschwarzent: Done

eclipse-scout-core/src/desktop/hybrid/HybridManager.ts
bschwarzent: Not a great benefit, but ok.

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
bschwarzent: Not technically necessary, but helpful when inspecting REST traffic, because it reduces unnecessary "noise" on the transport channel. Undefined values are automatically omitted during JSON serialization.

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
bschwarzent: I tend to ignore only null or absent arguments, but treat deliberately empty lists as normal.

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
bschwarzent: Unfortuantely, session.getModelAdapter() does not return the type `ModelAdapter` but `ModelAdapterLike`. Checking the variable like this has the benefit that it catches the null case, but also implicitly casts the variable to `ModelAdapter`.

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
bschwarzent: The intention was to use "null" as a default value during deserialization but "undefined" during serialization (again, to remove the attribute from the JSON).

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
bschwarzent: Done

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
bschwarzent: I'd rather use the long form to make it consistent with the previous line. There are no benefits from using the shorter form.

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
bschwarzent: Yes, I was confused, too. The API designers clearly didn't think that through... 😢 

Your proposal is better, I changed it.

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
bschwarzent: Done

eclipse-scout-core/src/desktop/hybrid/converter/HybridActionContextElementConverter.ts
bschwarzent: Done

eclipse-scout-core/src/desktop/hybrid/converter/HybridActionContextElementConverter.ts
bschwarzent: Done

eclipse-scout-core/src/desktop/hybrid/converter/HybridActionContextElementConverter.ts
bschwarzent: I tried this, but it did not make really make the implementation easier. In fact, the code is longer, because it needs to overwrite 4 methods instead of 2.

eclipse-scout-core/src/desktop/hybrid/converter/TreeNodeContextElementConverter.ts
bschwarzent: To save the necessary string interpolation in the non-null case. But this is probably negligible.

org.eclipse.scout.rt.client.test/src/test/java/org/eclipse/scout/rt/client/ui/desktop/hybrid/HybridActionContextElementsTest.java
bschwarzent: Done

org.eclipse.scout.rt.client.test/src/test/java/org/eclipse/scout/rt/client/ui/desktop/hybrid/HybridActionContextElementsTest.java
bschwarzent: Done

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/hybrid/AbstractHybridAction.java
bschwarzent: No, this was an accident (formatted twice). I reverted it.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/hybrid/AbstractHybridAction.java
bschwarzent: No, this was an accident (formatted twice). I reverted it.

org.eclipse.scout.rt.ui.html.test/src/test/java/org/eclipse/scout/rt/ui/html/json/desktop/hybrid/JsonHybridManagerTest.java
bschwarzent: Done.

eclipse-scout-core/src/desktop/hybrid/converter/HybridActionContextElementConverter.ts
bschwarzent: Done. The registry code was extracted to HybridActionContextElementConverters.ts.

eclipse-scout-core/src/desktop/hybrid/converter/HybridActionContextElementConverter.ts
bschwarzent: Ok, I'm convinced. The implementation was adjusted to match the Java equivalent.

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
bschwarzent: This code converts the JSON 1:1 to a JS object and vice-versa, without changing the structure. So an empty list will stay an empty list and a list of a null element will stay a list with a null element. No implicit "cleanup" logic should take place here.

eclipse-scout-core/src/desktop/hybrid/HybridManagerAdapter.ts
bschwarzent: Changed to adapter = scout.assertInstance(...)

PR #1296, opened 2024-12-09
Update JS build and test dependencies

cguglielmo:
scout.ts
typescript compiler "improved" conditional type constraints. ObjectFactory.ts had an error: object does not match {model: any} which is true but it should fallback to object, which does not happen anymore.

@types/jquery
traverse functions now return TElement instead of this.

karma-defaults.js
outputModule needs to be deleted otherwise it creates a runtime error: Uncaught SyntaxError: Unexpected token 'export'
happens since webpack 5.95.0

typescript and eslint cannot be updated to the latest version yet because they require IntelliJ 2024.

typescript-eslint v8 removed formatting rules. Update to this version is postponed until we decided whether we need the rules.

389554


cguglielmo:
Already merged

PR #1301, opened 2024-12-12
Fix incorrect rendering of components on saturday

Knicksaas:
When switching from DisplayMode Day to WorkWeek, the components on Saturday were rendered on the right hand side. This is not desired, the overflow should be invisible.

404100


eclipse-scout-core/src/calendar/Calendar.ts
```diff
@@ -2317,6 +2328,14 @@ export class Calendar extends Widget implements CalendarModel {
     }
   }
 
+  /**
+   * Removes range selection and triggers a selectedRangeChange event
+   */
+  removeRangeSelection() {
```
cguglielmo: This does only remove the selection in the DOM but the model state won't change. It will send selectedRangeChange events with the old range to the server. Better use _setSelectedRange instead, you can make it public

PR #1305, opened 2024-12-16
More robust /unload/ request handling.

christiandoebele:
- Prevent new session (session store) creations for unload requests.
- Handle /unload/ request in the trivial filter to avoid unwanted side effects in other filters (e.g. saml, oidc).

393576


mvilliger:
run test

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/TrivialAccessController.java
```diff
@@ -52,6 +53,8 @@
  */
 public class TrivialAccessController implements IAccessController {
 
+  private static final Pattern UNLOAD_PATH_PATTERN = Pattern.compile("^/unload/(.+)$");
```
mvilliger: This regex has been duplicated. UnloadRequestHandler uses the same. Make this here public and remove the other one?

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/UnloadRequestHandler.java
```diff
@@ -57,8 +57,14 @@ public boolean handlePost(HttpServletRequest req, HttpServletResponse resp) thro
   protected void handleUnloadRequest(HttpServletRequest req, HttpServletResponse resp, String uiSessionId) {
     LOG.info("Unloading UI session with ID {} (requested by UI)", uiSessionId);
 
-    final HttpSession httpSession = req.getSession();
-    final ISessionStore sessionStore = BEANS.get(HttpSessionHelper.class).getSessionStore(httpSession);
+    final HttpSession httpSession = req.getSession(false);
```
mvilliger, line 58: remove the UNLOAD_PATH_PATTERN regex in this class and use the one from the TrivialAccessController. Furthermore, I would remove all these unnecessary "final" keywords in this class to remove noise.

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/TrivialAccessController.java
```diff
@@ -254,4 +267,30 @@ public TrivialAuthConfig withLoginPageInstalled(final boolean loginPageInstalled
...
   }
+
+  /**
+   * Verify if the request is a browser /unload/ request. Unload requests are handled by {@code UnloadRequestHandler}.
+   */
+  protected boolean isUnloadRequest(final HttpServletRequest request) {
+    return UNLOAD_PATH_PATTERN.matcher(getTarget(request)).matches();
+  }
+
+  protected boolean isAuthenticatedRequest(final HttpServletRequest request) {
```
mvilliger: This method could be changed to e.g. getAuthenticationStatus returning e.g. {principal, status: "validSubject" or "none" or "principalValid" or "principalInvalid" as enum} 
Then this method could be used in handleRequest() to 
1. detect if authenticated and cancel unload
2. continue chain if subject is valid
3. abort if principal is invalid
4. continueChainAsSubject if principal is valid
Currently the logic is very hard to read and duplicated.

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/TrivialAccessController.java
```diff
@@ -254,4 +267,30 @@ public TrivialAuthConfig withLoginPageInstalled(final boolean loginPageInstalled
       return this;
     }
   }
+
+  /**
+   * Verify if the request is a browser /unload/ request. Unload requests are handled by {@code UnloadRequestHandler}.
+   */
+  protected boolean isUnloadRequest(final HttpServletRequest request) {
```
mvilliger: this method could be public static and then be used in UnloadRequestHandler?
Furthermore, I would remove all these unnecessary "final" keywords in this class to remove noise.

mvilliger (changes requested):
Thank you very much, looks good.
I only have some minor code style remarks.

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/UnloadRequestHandler.java
```diff
@@ -57,8 +57,14 @@ public boolean handlePost(HttpServletRequest req, HttpServletResponse resp) thro
   protected void handleUnloadRequest(HttpServletRequest req, HttpServletResponse resp, String uiSessionId) {
     LOG.info("Unloading UI session with ID {} (requested by UI)", uiSessionId);
```
bschwarzent, line 56: We should probably improve the logging in this method. This first line here is ok, but it might be useful for log analysis if the outcome would be logged as well.

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/UnloadRequestHandler.java
bschwarzent: I am not a fan of referencing a constant in the TrivialAccessController. The UnloadRequestHandler is the "owner" of the /unload path, so it is appropriate that it also defines the constant. It should not rely on the existence of the  TrivialAccessController, the application should be free to reimplement the authentication completely differently.  The relation should be inverse, but this can't be done because of the direction of the module dependency. 

In my view, it is not a problem that these constants are duplicated. We have other magic strings that are hard coded in the access controllers (e.g. "login").

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/UnloadRequestHandler.java
```diff
@@ -57,8 +57,14 @@ public boolean handlePost(HttpServletRequest req, HttpServletResponse resp) thro
...
+    if (httpSession == null) {
+      return;
+    }
+    final ISessionStore sessionStore = BEANS.get(HttpSessionHelper.class).optSessionStore(httpSession);
+    if (sessionStore == null) {
+      return;
+    }
     IUiSession uiSession = sessionStore.getUiSession(uiSessionId);
 
     if (uiSession != null) {
```
bschwarzent: Switch this if() as well, so it uses the same pattern for all early return points --> `if (uiSession == null) { return; }`

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/SessionStore.java
```diff
@@ -443,7 +443,7 @@ protected void removeClientSessionInsideWriteLock(final IClientSession clientSes
 
   @Override
   public void valueBound(HttpSessionBindingEvent event) {
-    // ignore notifications about being bound to an HTTP session
+    Assertions.assertTrue(m_httpSessionValid, "Binding to an invalid session is not supported.");
```
bschwarzent: This message is not entirely accurate, which might lead to confusion when this error happens. The new session is not invalid, the old one is.

Maybe something like this? `"Binding to new HTTP session {} is not supported (was already bound to {})", event.getSession().getId(), m_httpSessionId)`

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/SessionStore.java
```diff
@@ -443,7 +443,7 @@ protected void removeClientSessionInsideWriteLock(final IClientSession clientSes
 
   @Override
   public void valueBound(HttpSessionBindingEvent event) {
-    // ignore notifications about being bound to an HTTP session
```
bschwarzent, line 446: Consider adding a similar assertion to `preregisterUiSession` and `registerUiSession` to ensure nothing can be added to an invalidated store. Should never happen, but it would throw an early and explicit error in case it does.

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/HttpSessionHelper.java
```diff
@@ -60,6 +60,22 @@ public ISessionStore getSessionStore(HttpSession httpSession) {
...
+   * @return The {@link ISessionStore}. If there is no {@link ISessionStore} registered for the given
+   *         {@link HttpSession} {@code null} will be returned.
+   * @throws AssertionException
+   *           if the given HTTP session is <code>null</code>.
+   * @throws IllegalStateException
+   *           if the given HTTP session is invalid.
+   *
+   */
+  public ISessionStore optSessionStore(HttpSession httpSession) {
+    return getSessionStoreFromHttpSession(httpSession);
```
bschwarzent, line 76: Add `assertNotNull()` check. The JavaDoc states that it would be thrown. (Currently, a NPE would be thrown, which I find very ugly.)

Additionally, the first line of `getSessionStore` can then be simplified to this:
```
    ISessionStore sessionStore = optSessionStore(httpSession);
```

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/HttpSessionHelper.java
```diff
@@ -60,6 +60,22 @@ public ISessionStore getSessionStore(HttpSession httpSession) {
...
+   * Gets an existing session store from the given HTTP session. If there is no store available for the given session no new one
+   * will be created.
+   *
+   * @return The {@link ISessionStore}. If there is no {@link ISessionStore} registered for the given
+   *         {@link HttpSession} {@code null} will be returned.
+   * @throws AssertionException
+   *           if the given HTTP session is <code>null</code>.
+   * @throws IllegalStateException
+   *           if the given HTTP session is invalid.
+   *
```
bschwarzent: Remove unecessary empty line

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/TrivialAccessController.java
bschwarzent: I would keep this method, see my argumentation in UnloadRequestHandler below.

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/TrivialAccessController.java
```diff
@@ -254,4 +267,30 @@ public TrivialAuthConfig withLoginPageInstalled(final boolean loginPageInstalled
...
+      return true;
+    }
+
+    // Is already authenticated?
+    // Keep logic in sync with TrivialAccessController.handleRequest(HttpServletRequest, HttpServletResponse, FilterChain).
+    final Principal principal = BEANS.get(ServletFilterHelper.class).findPrincipal(request, m_config.getPrincipalProducer());
+    if (principal == null) {
+      return false;
+    }
+    if (m_config.m_principalVerifier == null) {
```
bschwarzent, line 297: Why not use m_config.getPrincipalVerifier() like above? Also, I agree with @mvilliger that we should try to combine this code into a single implementation. It should not be necessary to keep two different places in-sync within the same class.

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/TrivialAccessController.java
```diff
@@ -65,6 +68,14 @@ public boolean handle(final HttpServletRequest request, final HttpServletRespons
       return false;
     }
 
+    if (isUnloadRequest(request) && !isAuthenticatedRequest(request)) {
+      // Unload requests are handled by the {@code UnloadRequestHandler}.
+      // The unload request is supposed to close a UI session by the browser. If the current session is not authenticated then there is no session to unload.
+      // Answering the unauthenticated unload request here prevents saml or oidc filters to produce unwanted redirect urls after authentication (we want to avoid the /unload/ url as redirect url).
```
bschwarzent:
```suggestion
      // Answering the unauthenticated unload request here prevents other controllers (e.g. saml or oidc) to produce unwanted redirect urls after authentication (we want to avoid the /unload/ url as redirect url).
```

Use a more general wording here. SAML or OIDC are entirely implementation-dependent and not part of the Scout platform.

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/UnloadRequestHandler.java
mvilliger: I agree, it should be the other way round. But this is not possible. I prefer having it in one place. Otherwise they will diverge. 
If anybody would like to reimplement the authentitcation, this still works if the constant is only there.
As an alternative, all the constants (also the others like "login") could be extracted to an extra interface. Then it can be used by both clients.

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/TrivialAccessController.java
mvilliger: I disagree. I don't like it to have the logic "what is an unload request" in two places.
Its for sure wrong in the TrivialAccessController. But still it should be in one place.
Could be in a new class.

bschwarzent: Agreed! Lets find a good place for that logic. How about `ServletFilterHelper`?

mvilliger: Yes, sounds good!

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/TrivialAccessController.java
christiandoebele: This regex is now moved to ServletFilterHelper (only defined once now)

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/UnloadRequestHandler.java
christiandoebele: This regex is now moved to ServletFilterHelper (only defined once now)

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/TrivialAccessController.java
christiandoebele: new shared method added for getting the authentication status as recommended.

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/TrivialAccessController.java
christiandoebele: Method moved to ServletFilterHelper

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/UnloadRequestHandler.java
christiandoebele: The logging covers all the unexpected cases now

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/UnloadRequestHandler.java
christiandoebele: return pattern now also for if (uiSession == null)

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/SessionStore.java
christiandoebele: Log message is now more precise.

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/SessionStore.java
christiandoebele: preregisterUiSession and registerUiSession also assert a valid session now.

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/HttpSessionHelper.java
christiandoebele: done

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/HttpSessionHelper.java
christiandoebele: done

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/TrivialAccessController.java
christiandoebele: New shared method added for getting the authentication status

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/TrivialAccessController.java
christiandoebele: Log message changed as recommended.

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/TrivialAccessController.java
```diff
@@ -254,4 +257,83 @@ public TrivialAuthConfig withLoginPageInstalled(final boolean loginPageInstalled
...
+
+  /**
+   * Checks if the request is authenticated by either a valid subject or a valid principal.
+   *
+   * @param request
+   *     Http request
+   * @return {@code true} if request is authenticated. Otherwise {@code false}.
+   */
+  protected boolean isAuthenticatedRequest(HttpServletRequest request) {
+    return ObjectUtility.isOneOf(getAuthenticationStatus(request).getStatus(), AuthenticationStatusType.SUBJECT_VALID, AuthenticationStatusType.PRINCIPAL_VALID);
```
mvilliger: This method should get an AuthenticationStatusType instead of the request and only do the isOneOf. Otherwise getAuthenticationStatus(request) must be called twice.

mvilliger (changes requested):

mvilliger (approved):

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/TrivialAccessController.java
```diff
@@ -263,4 +265,83 @@ public TrivialAuthConfig withLoginPageInstalled(final boolean loginPageInstalled
...
+   *     authentication status type
+   * @return {@code true} if authenticated. Otherwise {@code false}.
+   */
+  protected boolean isAuthenticatedRequest(AuthenticationStatusType authenticationStatusType) {
+    return ObjectUtility.isOneOf(authenticationStatusType, AuthenticationStatusType.SUBJECT_VALID, AuthenticationStatusType.PRINCIPAL_VALID);
+  }
+
+  /**
+   * Gets the authentication status of the http request.
+   * <br/><br/>
```
bschwarzent: Use `<p>` instead

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/TrivialAccessController.java
```diff
@@ -263,4 +265,83 @@ public TrivialAuthConfig withLoginPageInstalled(final boolean loginPageInstalled
...
   }
+
+  /**
+   * Checks if the authentication status type is authenticated by either a valid subject or a valid principal.
+   *
+   * @param authenticationStatusType
+   *     authentication status type
+   * @return {@code true} if authenticated. Otherwise {@code false}.
+   */
+  protected boolean isAuthenticatedRequest(AuthenticationStatusType authenticationStatusType) {
```
bschwarzent: Argument should be `AuthenticationStatus` instance instead.

Is this method expected to be overwritten? Otherwise, consider moving it to the AuthenticationStatus class itself.

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/TrivialAccessController.java
```diff
@@ -263,4 +265,83 @@ public TrivialAuthConfig withLoginPageInstalled(final boolean loginPageInstalled
       return this;
     }
   }
+
+  /**
+   * Checks if the authentication status type is authenticated by either a valid subject or a valid principal.
+   *
+   * @param authenticationStatusType
+   *     authentication status type
```
bschwarzent: This `@param` description is useless (remove it entirely)

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/TrivialAccessController.java
```diff
@@ -66,7 +69,17 @@ public boolean handle(final HttpServletRequest request, final HttpServletRespons
       return false;
     }
 
-    switch (getTarget(request)) {
+    ServletFilterHelper helper = BEANS.get(ServletFilterHelper.class);
+    final AuthenticationStatus authenticationStatus = getAuthenticationStatus(request);
```
bschwarzent: Remove `final` keyword?

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/authentication/ServletFilterHelper.java
```diff
@@ -424,4 +426,28 @@ public boolean redirectAfterLogin(HttpServletRequest request, HttpServletRespons
     }
     return false;
   }
+
+  /**
+   * Get http request target.
```
bschwarzent: This method needs a better explanation.

Does it have to be public?

bschwarzent (approved):
Looks good. Thank you!

PR #1308, opened 2024-12-18
Calendar: fix static range selection

Knicksaas:
The range selection was only able to support a resolution of 30 minutes. When the numberOfHourDivisions were set to a value > 2, the selection was rounded up to the next half hour.

404446


Knicksaas:
Cherry pick #1304

PR #1311, opened 2024-12-19
DataObject support for Scout JS

mvilliger:
- Introduce a DataObjectTransformer which adds annotations during TypeScript compile time to all attributes of DataObjects. These annotations contain the data type of the attribute to be retrieved at runtime.
- Add a base class for DataObjects: BaseDoEntity
- Add ID classes in TS (number, string uuid, composite)
- Add a DoRegistry to retrieve all DataObjects known in TS
- Migrate existing DOs to the new base class including decorators.
- Add convenience methods to ajax utility to post and get DataObjects
- Add a DataObject (De)Serializer for TS. It uses the data type annotations added by the compile time transformer and supports:
  - Arrays
  - Maps
  - Sets
  - Records
  - Dates
  - Ids
- Enable experimental decorators for Scout JS
- Allow the unqualified ID deserializer in Java to accept over-qualified instances (a typeName is available event not necessary because the ID is concrete).
- Migrate from deprecated 'type' jQuery ajax property to the new 'method' property.

389583


eclipse-scout-core/src/id/CompositeId.ts
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {Id, IdParser, ObjectType, scout} from '../index';
+
+export abstract class CompositeId extends Id<Id<any>[], 'scout.CompositeId'> {
```
fschinkel: Is this class intended to be the pendant of `AbstractCompositeId`? If it is wee need to pass the typeName to the implementing class.
Did you consider introducing generics for the value type? The implementing class can then use tupel types which will increase comfort for the developers using these composite ids.

eclipse-scout-core/src/id/IdParser.ts
```diff
@@ -0,0 +1,94 @@
...
+    const rawValue = signatureSplit[0];
+    const signature = signatureSplit[1];
+    const rawIdComponents = rawValue.split(IdParser.COMPONENT_DELIMITER);
+    const idModel = {value: rawIdComponents, typeName, signature} as InitModelOf<TIdType>;
+    const objectTypeOverwrite = this.getObjectTypeForTypeName(typeName, idObjectType) as ObjectType<TIdType>;
+    return scout.create(objectTypeOverwrite, idModel, {ensureUniqueId: false});
+  }
+
+  protected _splitSignature(value: string): string[] {
+    let splitPos = value.lastIndexOf(IdParser.SIGNATURE_DELIMITER);
```
fschinkel, line 83: is this split-logic really necessary? splitting the string will result in an array with length 1 or 2 and there is no such thing as IndexOutOfBounds in JS

eclipse-scout-core/test/uinotification/uiNotificationsSpec.ts
```diff
@@ -92,9 +92,10 @@ describe('uiNotifications', () => {
         receivedMsgCcc = event.message;
       });
 
-      let response: UiNotificationResponse = {
+      let response = scout.create(UiNotificationResponse, {
         notifications: [{
           id: '1',
+          objectType: UiNotificationDo,
```
fschinkel: Can we eliminate this?

eclipse-scout-core/src/ajax/ajax.ts
```diff
@@ -192,13 +192,40 @@ export const ajax = {
...
     let opts = $.extend({}, {
-      type: 'POST',
+      method: 'POST',
       dataType: 'json',
       contentType: 'application/json; charset=UTF-8'
     }, options);
     return ajax.createCall(opts, model);
   },
 
+  getDataObject<TDoOut extends BaseDoEntity>(url?: string | JQuery.UrlAjaxSettings, model?: AjaxCallModel): JQuery.Promise<TDoOut, AjaxError> {
```
fschinkel: please add js doc

eclipse-scout-core/test/dataobject/DoDeserializerSpec.ts
```diff
@@ -0,0 +1,218 @@
...
+    const withConstructorObjectType = {
+      objectType: Fixture03Do,
+      nestedNestedDate: '2024-07-25 08:41:10.708Z'
+    };
+    const resultFromConstructorObjectType = new DoDeserializer().deserialize(withConstructorObjectType) as Fixture03Do;
+    expect(resultFromConstructorObjectType).toBeInstanceOf(Fixture03Do);
+    expect(resultFromConstructorObjectType.nestedNestedDate).toEqual(dates.parseJsonDate('2024-07-25 08:41:10.708Z'));
+  });
+
+  it('Uses BaseDoEntity if no type information is available', () => {
```
fschinkel: uses

eclipse-scout-core/src/dataobject/DoRegistry.ts
```diff
@@ -0,0 +1,94 @@
...
+  protected _jsonTypeByDataObject = new Map<string, string>();
+  protected _objectTypeByJsonType = new Map<string, string>();
+
+  init() {
+    ObjectFactory.get()
+      .getSubClassesOf(BaseDoEntity)
+      .forEach(doClass => this.add(doClass));
+  }
+
+  add(DoClass: Constructor<BaseDoEntity>, jsonType?: string) {
```
fschinkel: doClass, also see other methods

eclipse-scout-core/src/dataobject/serialize/doValueMetaData.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  resolveToConstructor<T>(objectType: ObjectType<T>): Constructor<T> {
+    return TypeDescriptor.resolveType(objectType, {variantLenient: true});
+  },
+
+  /**
+   * Checks if `actualType` is instanceof `declaredType`.
+   */
+  assertTypesCompatible(actualType: Constructor, declaredType: Constructor) {
+    if (actualType && declaredType && actualType !== declaredType && !declaredType.isPrototypeOf(actualType)) {
```
fschinkel, line 105: flip if?

eclipse-scout-core/src/dataobject/serialize/IdDoNodeSerializer.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    // TypeName is part of the datatype declaration (E.g. UuId<'scout.MyIdType'>): parse from metadata type argument
+    const typeNameFromMetaData = this.parseTypeName(metaData);
+    if (typeNameFromMetaData) {
+      return typeNameFromMetaData;
+    }
+
+    // Type name cannot be computed. Check if ID type is fixed on instance (e.g. using the idTypeName annotation or other constructor implementation)
+    const IdType = metaData.type;
```
fschinkel, line 60: idType

eclipse-scout-core/src/dataobject/serialize/ArrayDoNodeSerializer.ts
```diff
@@ -0,0 +1,38 @@
...
+    return objects.isArray(value);
+  }
+
+  deserialize(array: Array<any>, metaData: DoValueMetaData, deserializer: DoDeserializer): Array<any> {
+    const arrayValueType = ArrayDoNodeSerializer.getArrayValueType(metaData);
+    return array.map(e => deserializer.deserialize(e, arrayValueType));
+  }
+
+  static getArrayValueType(metaData: DoValueMetaData): DoValueMetaData {
+    if (metaData?.args?.length) {
```
fschinkel:
```suggestion
return metaData?.args?.[0];
```

eclipse-scout-core/src/dataobject/dataObjects.ts
```diff
@@ -7,10 +7,54 @@
...
-
-import {arrays, DoEntity, objects, scout} from '..';
+import {
+  ArrayDoNodeSerializer, arrays, DateDoNodeSerializer, DoDeserializer, DoEntity, DoNodeSerializer, DoSerializer, doValueMetaData, IdDoNodeSerializer, MapDoNodeSerializer, objects, ObjectType, scout, SetDoNodeSerializer
+} from '../index';
 
 export const dataObjects = {
+
+  serializers: [
+    new DateDoNodeSerializer(),
```
fschinkel, line 25: how do other layers/projects contribute new serializers? do they simply inject them into this array at the correct position?

eclipse-scout-core/src/id/CompositeId.ts
mvilliger: Good point is fixed.

eclipse-scout-core/src/id/IdParser.ts
mvilliger: I prefer lastIndexOf because it searches from the end while split would search from the start.

eclipse-scout-core/test/uinotification/uiNotificationsSpec.ts
mvilliger: We could yes, but for this fixture it is required to have the same structure as it would come from the deserializer

eclipse-scout-core/src/ajax/ajax.ts
mvilliger: JS doc is completely missing and will be added later on. Same for documentation

eclipse-scout-core/test/dataobject/DoDeserializerSpec.ts
mvilliger: Fixed

eclipse-scout-core/src/dataobject/DoRegistry.ts
mvilliger: lowercase is correct here

eclipse-scout-core/src/dataobject/serialize/doValueMetaData.ts
mvilliger: Would be possible yes, but makes it not really better: one more line vs one more indent

eclipse-scout-core/src/dataobject/serialize/IdDoNodeSerializer.ts
mvilliger: Capitalized because a new instance is created. necessary to make eslint happy.

eclipse-scout-core/src/dataobject/serialize/ArrayDoNodeSerializer.ts
mvilliger: Is applied

eclipse-scout-core/src/dataobject/dataObjects.ts
mvilliger: Yes

eclipse-scout-core/src/dataobject/serialize/DoDeserializer.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export class DoDeserializer {
+
+  deserialize<T extends object>(value: any, valueMetaData?: DoValueMetaData<T>): T {
+    const deserializer = dataObjects.serializers.find(d => d.canDeserialize(value, valueMetaData));
+    if (deserializer) {
+      // use custom deserializer
+      return deserializer.deserialize(value, valueMetaData, this);
+    }
+    if (objects.isNullOrUndefined(value)) {
```
cguglielmo: Shouldn't this be the first statement? Or are there serializers that can serialize null or undefined?

eclipse-scout-core/src/id/IdParser.ts
```diff
@@ -0,0 +1,94 @@
...
+      throw new Error('Id value is mandatory.');
+    }
+    const isQualified = strings.contains(value, IdParser.ID_TYPENAME_DELIMITER);
+    if (isQualified) {
+      return this.fromQualified(idObjectType, value);
+    }
+    return this.fromUnqualified(idObjectType, value, typeNameProvider());
+  }
+
+  fromUnqualified<TIdType extends Id<TValueType, TTypeName>, TTypeName extends string, TValueType>(idObjectType: ObjectType<TIdType>, value: string, typeName: string): TIdType {
```
cguglielmo, line 69: is it necessary to set ensureUniqueId: false ? ObjectFactory excludes Id type

eclipse-scout-core/src/id/NumberId.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * LongId
+ */
+export class NumberId<TTypeName extends string> extends Id<number, TTypeName> {
+
+  protected override _initIdValue(value: number | string[]) {
+    if (typeof value === 'number') {
+      this.value = value;
+    } else {
+      const invalidNumberId = () => new Error('Invalid NumberId: ' + arrays.format(value, IdParser.COMPONENT_DELIMITER));
```
cguglielmo: maybe extract to method instead of creating function on the fly?

eclipse-scout-core/src/id/StringId.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+import {arrays, Id, IdParser, scout} from '../index';
+
+export class StringId<TTypeName extends string> extends Id<string, TTypeName> {
+
+  protected override _initIdValue(value: string | string[]) {
+    if (typeof value === 'string') {
+      this.value = value;
+    } else {
+      const invalidStringId = () => new Error('Invalid StringId: ' + arrays.format(value, IdParser.COMPONENT_DELIMITER));
```
cguglielmo: see NumberId comment

eclipse-scout-core/src/id/UuId.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+import {arrays, Id, IdParser, scout} from '../index';
+
+export class UuId<TTypeName extends string> extends Id<string, TTypeName> {
+
+  protected override _initIdValue(value: string | string[]) {
+    if (typeof value === 'string') {
+      this.value = value;
+    } else {
+      const invalidUuId = () => new Error('Invalid UuId: ' + arrays.format(value, IdParser.COMPONENT_DELIMITER));
```
cguglielmo: see NumberId comment, maybe move to Id to reuse it

eclipse-scout-core/test/dataobject/dataObjectsSpec.ts
```diff
@@ -23,10 +23,10 @@ describe('dataObjects', () => {
 
   describe('addContribution', () => {
     it('adds a contribution to the DO', () => {
-      let doEntity: DoEntityWithContributions = {};
+      const doEntity: DoEntityWithContributions = {};
```
cguglielmo, line 26: we don't have a guideline for let vs const (yet). No need to replace let with const everywhere. I personally prefer let over const

eclipse-scout-core/src/dataobject/serialize/DoDeserializer.ts
mvilliger: No, it is intended like this. If the customizable deserializers come first, it gives them the power to e.g. replace an undefined or null value with e.g. an empty string if an object may require that.
If the null/undefined case would be handled first, such flexibility would get lost.

eclipse-scout-core/src/id/IdParser.ts
mvilliger: No it's not necessary. Its just another layer of security in case we don't want to keep the change in the ObjectFactory.

eclipse-scout-core/src/id/NumberId.ts
mvilliger: Agree. Fixed.

eclipse-scout-core/src/id/StringId.ts
mvilliger: Agree. Fixed.

eclipse-scout-core/src/id/UuId.ts
mvilliger: Agree. Fixed.

eclipse-scout-core/test/dataobject/dataObjectsSpec.ts
mvilliger: Ok. fine with me. Especially in specs for me const is useful to not by accident overwrite fixture variables from previous expectations.

eclipse-scout-core/src/ObjectFactory.ts
```diff
@@ -164,10 +164,11 @@ export class ObjectFactory {
     options.model = model;
 
     // Create object
-    let scoutObject = this._createObjectByType(objectType, options);
+    const scoutObject = this._createObjectByType(objectType, options);
+    const ensureUniqueId = scout.nvl(options.ensureUniqueId, !(scoutObject instanceof BaseDoEntity || scoutObject instanceof Id) /* don't create unique IDs for DOs and IDs by default */);
```
bschwarzent: Maybe extract this logic (or the actual assignment of the `id`) to a method, so it can be customized?

eclipse-scout-core/src/ObjectFactory.ts
```diff
@@ -242,6 +243,24 @@ export class ObjectFactory {
     return this._objectTypeMap.get(Class);
   }
 
+  /**
+   * @param baseClass The base class (exclusive) for which all known subclasses should be returned.
+   * @returns All classes that have the given class in their super hierarchy. The given baseClass is not part of the result.
+   * More formally: all constructors known to this factory that have the given class in their prototype chain.
+   */
+  getSubClassesOf<T extends object>(baseClass: Constructor<T> | AbstractConstructor<T>): Constructor<T>[] {
```
bschwarzent, line 264: Is `extends object` necessary?

eclipse-scout-core/src/ObjectFactory.ts
```diff
@@ -242,6 +243,24 @@ export class ObjectFactory {
...
+    const result: Constructor<T>[] = [];
+    if (!baseClass) {
+      return result;
+    }
+    for (let obj of this._objectTypeMap.keys()) {
```
bschwarzent, line 269: Can the type of `_objectTypeMap` be simplified to `Map<Constructor, string>`?

eclipse-scout-core/src/ajax/ajax.ts
bschwarzent: Why can `url` also be an UrlAjaxSettings object? The other `get*` methods in this utility only use `string`. I would use the same pattern here. If the caller wants to set other options, they should use `callDataObject`.

eclipse-scout-core/src/ajax/ajax.ts
```diff
@@ -192,13 +192,40 @@ export const ajax = {
...
+    const urlAjaxSettings = typeof url === 'string' ? {url: url} : url;
+    const opts: JQuery.UrlAjaxSettings = $.extend({}, {method: 'GET'}, urlAjaxSettings);
+    return ajax.callDataObject(null, opts, model);
+  },
+
+  postDataObject<TDoIn extends BaseDoEntity | void, TDoOut extends BaseDoEntity | void>(dataObject: TDoIn, url?: string | JQuery.UrlAjaxSettings, model?: AjaxCallModel): JQuery.Promise<TDoOut, AjaxError> {
```
bschwarzent: We should probably offer the same convenice functions for "DataObject" as for "Json" -> add `putDataObject` and `removeDataObject`.

eclipse-scout-core/src/ajax/ajax.ts
```diff
@@ -192,13 +192,40 @@ export const ajax = {
...
+    const opts: JQuery.UrlAjaxSettings = $.extend({}, {method: 'POST'}, urlAjaxSettings);
+    return ajax.callDataObject(dataObject, opts, model);
+  },
+
+  callDataObject<TDoIn extends BaseDoEntity | void, TDoOut extends BaseDoEntity | void>(dataObject: TDoIn, options?: JQuery.UrlAjaxSettings, model?: AjaxCallModel): JQuery.Promise<TDoOut, AjaxError> {
+    return ajax.createCallDataObject(dataObject, options, model).call();
+  },
+
+  createCallDataObject(dataObject: DoEntity | void, urlAjaxSettings?: JQuery.UrlAjaxSettings, model?: AjaxCallModel): AjaxCall {
+    const json = dataObject ? dataObjects.stringify(dataObject) : undefined;
```
bschwarzent: "undefined" is only used to omit the property in the result of the extend() call. I would simplify this here (`json = dataObjects.stringify()`) and use `data: json || undefined` below instead.

eclipse-scout-core/src/code/CodeTypeCache.ts
```diff
@@ -189,14 +190,52 @@ export class CodeTypeCache extends EventEmitter implements ObjectModel<CodeTypeC
...
 }
 
-export interface CodeTypeRequest extends DoEntity {
+@typeName('scout.CodeTypeRequest')
+export class CodeTypeRequest extends BaseDoEntity {
   codeTypeIds?: string[];
 }
 
+@typeName('scout.Code')
+export class CodeDo extends BaseDoEntity implements CodeModel<any> {
```
bschwarzent: Is this defined in the correct file? Why not in Code.ts? Or even in a dedicated file like we do for widget models.

eclipse-scout-core/src/dataobject/DoEntity.ts
```diff
@@ -8,12 +8,51 @@
  * SPDX-License-Identifier: EPL-2.0
  */
 
+import {Constructor, objects, scout} from '../index';
+
 /**
- * Base interface for all data object entities.
+ * Base interface for all data objects.
```
bschwarzent: Encourage readers to extend the BaseDoEntity class instead of implementing this interface. This mainly exists for legacy reasons, right?

eclipse-scout-core/src/dataobject/DoEntity.ts
```diff
@@ -8,12 +8,51 @@
...
  *
  * @see "org.eclipse.scout.rt.dataobject.IDoEntity"
  */
 export interface DoEntity {
   _type?: string;
   _typeVersion?: string;
 }
+
+export class BaseDoEntity implements DoEntity, BaseDoEntityModel {
+  declare model: Partial<this> | BaseDoEntityModel;
```
bschwarzent: - What is the intention of "Partial<this>"?
- Why is the model argument in the init() method typed as "any" instead of InitModelOf<this>? What other options are there? init() does nothing if the argument is not a POJO. Assertion?

eclipse-scout-core/src/dataobject/DoRegistry.ts
bschwarzent: Is it really that useful to use uppercase for the `DoClass` argument, just because it happens to be of type "Constructor"? I would reserve uppercase names for things that are actually classes.

eclipse-scout-core/src/dataobject/DoRegistry.ts
```diff
@@ -0,0 +1,94 @@
...
+    if (!DoClass) {
+      return;
+    }
+    jsonType = jsonType || this._readJsonType(DoClass);
+    if (!jsonType) {
+      return;
+    }
+
+    this._constructorByJsonType.set(jsonType, DoClass);
+    const objectType = ObjectFactory.get().getObjectType(DoClass);
```
bschwarzent: General remark: should we really default to "const" instead of "let" for local variables? Or is that just the default inserted by the IDE? I fail to see the benefit of defining everything as const. It is similar to the "final" Keyword in Java which we used excessively at some point in time but have since stopped using.

eclipse-scout-core/src/dataobject/DoRegistry.ts
```diff
@@ -0,0 +1,94 @@
...
+      this._jsonTypeByDataObject.delete(objectType);
+    }
+    this._objectTypeByJsonType.delete(jsonType);
+  }
+
+  protected _readJsonType(DoClass: Constructor<BaseDoEntity>): string {
+    return new DoClass()._type;
+  }
+
+  allDataObjects(): IterableIterator<Constructor<BaseDoEntity>> {
```
bschwarzent: Check if there is a better name. The method does not really return data objects (only data object _classes_).

eclipse-scout-core/src/dataobject/serialize/ArrayDoNodeSerializer.ts
```diff
@@ -0,0 +1,35 @@
...
+  canDeserialize(value: any, metaData: DoValueMetaData): boolean {
+    return objects.isArray(value);
+  }
+
+  deserialize(array: Array<any>, metaData: DoValueMetaData, deserializer: DoDeserializer): Array<any> {
+    const arrayValueType = ArrayDoNodeSerializer.getArrayValueType(metaData);
+    return array.map(e => deserializer.deserialize(e, arrayValueType));
+  }
+
+  static getArrayValueType(metaData: DoValueMetaData): DoValueMetaData {
```
bschwarzent: More suitable in doValueMetaData.ts?

eclipse-scout-core/src/dataobject/serialize/DoDeserializer.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+    return value;
+  }
+
+  protected _deserializeObject<T extends object>(rawObj: Record<string, any>, metaData?: DoValueMetaData<T>): T {
+    const detectedClass = this._detectClass(rawObj) as Constructor<T>;
+
+    let constructor = metaData?.type as Constructor<T>;
+    if (constructor && detectedClass) {
+      // if both is present: check compatibility and use the one from _type attribute (might be a subclass of the one in the source code)
```
bschwarzent: "both is present" -> "both are present"

eclipse-scout-core/src/dataobject/serialize/DoDeserializer.ts
```diff
@@ -0,0 +1,74 @@
...
+      constructor = detectedClass;
+    } else if (detectedClass) {
+      // no metadata: use from _type attribute
+      constructor = detectedClass;
+    } else if (!constructor) {
+      // default DO if no _type attribute and no metadata is available
+      constructor = BaseDoEntity as Constructor<T>;
+    }
+
+    const resultObj = scout.create(constructor, null /* must always be possible to create a DO without model */, {ensureUniqueId: false});
```
bschwarzent: Is the "ensureUniqueId" argument still needed? ObjectFactory will automatically skip id assignment if object is a BaseDoEntity.

eclipse-scout-core/src/dataobject/serialize/DoSerializer.ts
```diff
@@ -0,0 +1,139 @@
...
+export class DoSerializerCycleDetector {
+  protected _stack = [];
+  protected _attributeNames: string[] = [];
+
+  push(value: any) {
+    if (!this._canHaveChildren(value)) {
+      return;
+    }
+    if (this.contains(value)) {
+      const path = '[' + strings.join(',', ...this._attributeNames) + ']';
```
bschwarzent: `this._attributeNames.join(',')` ?

eclipse-scout-core/src/dataobject/serialize/DoSerializer.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  protected _convertFieldValue(proto: object, fieldName: string, value: any): any {
+    this.cycleDetector.pushAttributeName(fieldName);
+    const fieldMetaData = doValueMetaData.getFieldMetaData(proto, fieldName);
+    const result = this.serialize(value, fieldMetaData);
+    this.cycleDetector.popAttributeName(fieldName);
+    return result;
+  }
+
+  protected _setJsonTypeTo(obj: any, constructor: Constructor) {
```
bschwarzent: `_ensureJsonType` ?

eclipse-scout-core/src/dataobject/serialize/IdDoNodeSerializer.ts
```diff
@@ -0,0 +1,78 @@
...
+      return false;
+    }
+    return this.isIdClass(metaData.type) || Id.isPrototypeOf(metaData.type);
+  }
+
+  deserialize(rawId: string, metaData: DoValueMetaData, deserializer: DoDeserializer): Id<any> {
+    return IdDoNodeSerializer._getIdParser().parse(metaData.type, rawId, () => this.detectConcreteTypeName(metaData));
+  }
+
+  isIdClass(candidate: Constructor) {
```
bschwarzent: Missing return type.

Also: what is the expected ts-error here?

eclipse-scout-core/src/dataobject/serialize/IdDoNodeSerializer.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    // Type name cannot be computed. Check if ID type is fixed on instance (e.g. using the idTypeName annotation or other constructor implementation)
+    const IdType = metaData.type;
+    const id = new IdType() as Id<any>;
+    return id.typeName;
+  }
+
+  parseTypeName(metaData: DoValueMetaData): string {
+    return metaData.type['parseTypeName'](metaData);
```
bschwarzent: Why is bracket notation necessary here? Is this not part of the official API?

eclipse-scout-core/src/dataobject/serialize/IdDoNodeSerializer.ts
```diff
@@ -0,0 +1,78 @@
...
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {Constructor, DoDeserializer, DoNodeSerializer, DoSerializer, DoValueMetaData, Id, IdParser, scout} from '../../index';
+
+export class IdDoNodeSerializer implements DoNodeSerializer<Id<any>> {
+
+  protected static _ID_PARSER: IdParser; // created on first use. Access using IdDoNodeSerializer._getIdParser()
```
bschwarzent: Alternatively, we could provide a getSingleton() method on the IdParser class itself.

eclipse-scout-core/src/ajax/ajax.ts
```diff
@@ -26,7 +26,7 @@ export const ajax = {
   get(url: string, options?: JQuery.AjaxSettings, model?: AjaxCallModel): JQuery.Promise<any, AjaxError> {
     let opts = $.extend({}, {
       url: url,
-      type: 'GET'
+      method: 'GET'
```
bschwarzent, line 29: Consider a separate commit for this change, it has nothing to do with data objects.

eclipse-scout-core/src/id/Id.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  toQualified(): string {
+    if (this.typeName) {
+      return this.typeName + IdParser.ID_TYPENAME_DELIMITER + this.toUnqualified();
+    }
+    return this.toUnqualified();
+  }
+
+  static parseTypeName(idMetaData: DoValueMetaData): string {
```
bschwarzent: No parsing takes place in this function. Is it even called somewhere?

eclipse-scout-core/src/id/Id.ts
```diff
@@ -0,0 +1,111 @@
...
+      }
+      for (let i = 0; i < a.length; i++) {
+        if (!this._equals(a[i], b[i])) {
+          return false;
+        }
+      }
+      return true;
+    }
+
+    if (a instanceof Date && b instanceof Date) {
```
bschwarzent: This seems odd and should at least be explained. Do we really use Ids that contain a Date? Shouldn't objects.equals() already handle this?

eclipse-scout-core/src/id/IdParser.ts
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {Id, InitModelOf, ObjectType, ObjectWithType, scout, strings} from '../index';
+
+const registry: Map<string, ObjectType<Id<any>>> = new Map();
+
+export class IdParser implements ObjectWithType {
+
+  static ID_TYPENAME_DELIMITER = ':';
```
bschwarzent: What does this constant have the ID_ prefix, but the other two don't? Maybe this should be called TYPE_NAME_DELIMITER?

eclipse-scout-core/src/id/Id.ts
```diff
@@ -0,0 +1,111 @@
...
+  objectType: string;
+
+  init(model: InitModelOf<this>) {
+    const value = scout.assertParameter('value', model?.value);
+    this.typeName = scout.nvl(model.typeName, this.typeName);
+    this.signature = model.signature;
+    this._initIdValue(value);
+  }
+
+  protected abstract _initIdValue(value: TIdValueType | string[]);
```
bschwarzent, line 36: When is the argument a string array? Is this only needed for the CompositeId? But is the value of the composite id not already an array?

eclipse-scout-core/src/ObjectFactory.ts
mvilliger: Agree. Fixed

eclipse-scout-core/src/ObjectFactory.ts
mvilliger: Basically no as it already must be a constructor. But it is the same as e.g. in the create function and makes it more clear that it makes no sense for primitives.

eclipse-scout-core/src/ObjectFactory.ts
mvilliger: Yes. Fixed.

eclipse-scout-core/src/ajax/ajax.ts
mvilliger: Agree. Fixed.

eclipse-scout-core/src/ajax/ajax.ts
mvilliger: Agree. Fixed.

eclipse-scout-core/src/ajax/ajax.ts
mvilliger: Ok. Changed.

eclipse-scout-core/src/code/CodeTypeCache.ts
mvilliger: Well, the class is actually only used in the cache. That's why it is declared there. But I extracted it into own files. Then the cache has less noise.

eclipse-scout-core/src/dataobject/DoEntity.ts
mvilliger: Yes, JsDoc is completely missing and needs to be added/updated anyway.

eclipse-scout-core/src/dataobject/DoEntity.ts
mvilliger: - Partial ensures all properties are optional in the model. In case a property in the DO is not.
- Argument changed to InitModelOf<this>. I would not add an assertion because if a subclass would like to change that behavior, it cannot make a super call anymore. Futhermore the model may also be null which is correct.

eclipse-scout-core/src/dataobject/DoRegistry.ts
mvilliger: Well, it is used in new (e.g. in _readJsonType). And ESLint compains if lowerCase variables are used with new. But I limited the uppercase variables to the minimum now.

eclipse-scout-core/src/dataobject/DoRegistry.ts
mvilliger: I prefer const over let to ensure I don't override the variable by accident. But it is longer. I think in the end it doesn't matter. As a reader of the code const helps me to know, that this variable will not be changed anywhere below which helps to follow the code flow in my opinion. The difference to Java is that a variable can be declared without keyword while in JS one keyword is required anyway. So I prefer the one which explains the variable better. But it's no strong opinion towards const.

eclipse-scout-core/src/dataobject/DoRegistry.ts
mvilliger: Agree. Fixed.

eclipse-scout-core/src/dataobject/serialize/ArrayDoNodeSerializer.ts
mvilliger: Doesn't really matter. I changed it to the utility.

eclipse-scout-core/src/dataobject/serialize/DoDeserializer.ts
mvilliger: Correct. Fixed.

eclipse-scout-core/src/dataobject/serialize/DoDeserializer.ts
mvilliger: No, it's not needed. It was only there in case we don't want to keep the new rule in the ObjectFactory. But as this rule seems to be fine, i removed the option here already.

eclipse-scout-core/src/dataobject/serialize/DoSerializer.ts
mvilliger: Yes, fine with me. But is the behavior of the native join function really the same as the one from Scout?
Especially because the scout version uses strings.asString which has some extra logic?

eclipse-scout-core/src/dataobject/serialize/DoSerializer.ts
mvilliger: Agree. Fixed.

eclipse-scout-core/src/dataobject/serialize/IdDoNodeSerializer.ts
mvilliger: Correct. Fixed.

The ts error is TS2367: unintentional comparison.

eclipse-scout-core/src/dataobject/serialize/IdDoNodeSerializer.ts
mvilliger: It is part of the API. But the call is by reflection here. Is there a better way to write this? As this method could be overwritten.

eclipse-scout-core/src/dataobject/serialize/IdDoNodeSerializer.ts
mvilliger: Agree. Fixed.

eclipse-scout-core/src/ajax/ajax.ts
mvilliger: That's true. But the extra work to separate this now does not pay out here I think.

eclipse-scout-core/src/id/Id.ts
mvilliger: Right. I changed the name. Yes, it is used by reflection in IdDoNodeSerializer

eclipse-scout-core/src/id/Id.ts
mvilliger: Yes, we really have IDs with dates. This is officially supported and used.

Yes objects.equals SHOULD handle this, but it does not. And we don't think that this is a good moment to change that. After discussions with cgu about the same topic, we decided to introduce an equals2 and equalsRecursively2 which handles these cases better. In the next release it is the plan to switch to the new implementation so that we have time to stabilize again.

eclipse-scout-core/src/id/IdParser.ts
mvilliger: Agree. Fixed.

eclipse-scout-core/src/id/Id.ts
mvilliger: See e.g. DomainKey.ts
The value type of the ID (type arg) can be different than the serialized model.

cguglielmo:
Merged with #1413

PR #1317, opened 2024-12-24
fix clipboard form selection for first text

PossessedC0bra:
- make clipboard field positioning static so double click text selection works properly for the first text in the field

405372


eclipse-scout-core/src/form/fields/clipboardfield/ClipboardField.less
```diff
@@ -13,6 +13,7 @@
   padding: @text-field-padding-y @text-field-padding-x;
   #scout.user-select(text);
   white-space: pre-wrap;
+  position: static;
```
cguglielmo: Any explanation why this works? position: static will be overridden by position: relative, so it actually has no effect.

PR #1318, opened 2025-01-03
Remove SharedVariableMap version property

mvilliger:
This property is not necessary. Compare map contents instead of version numbers.

405215


stephan-merkli (commented):
Code looks good to me.

PR #1323, opened 2025-01-09
LookupBox: make customizing easier

cguglielmo:
The table inside a list box uses other default values for some properties than the regular table, for example checkable is set to true. If such a table needs to be customized, the default values should be preserved, so they don't have to be copied.

385612

### DIFF
--- a/eclipse-scout-core/src/form/fields/LookupBox.ts
+++ b/eclipse-scout-core/src/form/fields/LookupBox.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
--- a/eclipse-scout-core/src/table/columns/LookupEditor.ts
+++ b/eclipse-scout-core/src/table/columns/LookupEditor.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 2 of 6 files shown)


eclipse-scout-core/src/table/columns/LookupEditor.ts
```diff
@@ -152,7 +152,6 @@ export class LookupEditor<TValue> extends ValueField<TValue[]> implements Lookup
         statusVisible: false,
         tree: {
           objectType: Tree,
```
fschinkel, line 154: Do we still need the objectType, or is it only added for code completion?

eclipse-scout-core/src/table/columns/LookupEditor.ts
```diff
@@ -166,11 +165,6 @@ export class LookupEditor<TValue> extends ValueField<TValue[]> implements Lookup
       statusVisible: false,
       table: {
         objectType: Table,
```
fschinkel, line 167: Do we still need the objectType, or is it only added for code completion?

eclipse-scout-core/src/form/fields/LookupBox.ts
```diff
@@ -67,6 +67,16 @@ export abstract class LookupBox<TValue> extends ValueField<TValue[], TValue | TV
 
   protected abstract _initStructure(value: TValue[]);
 
+  protected abstract _createStructureModel(): ModelOf<Widget>;
+
+  override _createChild<T extends Widget>(widgetOrModel: ObjectOrChildModel<T> | string): T {
```
fschinkel: This does not work, as it enhances every model of a child with the structureModel. LookupBoxes in scout.rt have two children, the tree/table and the filterBox. With this e.g. the filterBox of a ListBox is enhanced with columns.
For scout.rt this might not be problematic as the properties of a tree/table do not collide with the ones of a RadioButtonGroup, which is normally used as filterBox. But as the type of the filterBox is Widget, it can also be e.g. a Table, and then there will be a lot of unexpected behaviour.

fschinkel (changes requested):

PR #1324, opened 2025-01-10
SequenceBox: mark as focused when a field without label is focused

bschwarzent:
FormFields are marked as "focused" when they have the focus. This causes their form field label to be highlighted. If a field without a label is used in a SequenceBox with a label, that label was not highlighted, although this is expected by the user (because the label appears to belong to the field while it technically belongs to the sequence box).

This change listens for the "focused" property of inner fields and propagates it to the sequence box, as long as the inner field does not have a visible label of its own. As soon as a visible label is encountered, the sequence box is no longer marked. (Special case: labels that only contain a single punctuation character in are treated as if their label was empty. This is useful for "from - to" sequence boxes that only use a "-" sign as label for the second field).

The default behavior can be overwritten by adding marker classes:
- sequence-box-focused-always (sequence box)
- sequence-box-focused-never (sequence box)
- sequence-box-focused-assume-label (inner field)
- sequence-box-focused-assume-no-label (inner field)

390352

### DIFF
--- a/eclipse-scout-core/src/form/fields/sequencebox/SequenceBox.ts
+++ b/eclipse-scout-core/src/form/fields/sequencebox/SequenceBox.ts
@@ -8,8 +8,8 @@
  * SPDX-License-Identifier: EPL-2.0
  */
 import {
-  CheckBoxField, CloneOptions, CompositeField, DateField, dates, EventHandler, FormField, FormFieldSuppressStatus, HorizontalGrid, HtmlComponent, InitModelOf, LogicalGrid, LogicalGridData, LogicalGridLayout, LogicalGridLayoutConfig, Menu,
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/src/form/fields/sequencebox/SequenceBox.ts
```diff
@@ -216,6 +217,59 @@ export class SequenceBox extends CompositeField implements SequenceBoxModel {
...
+  protected _computeSequenceBoxFocused(focusedField: FormField) {
+    // Special marker classes to override the default behavior
+    if (this.hasCssClass('consider-inner-focus')) {
+      return true;
+    }
+    if (this.hasCssClass('never-consider-inner-focus')) {
+      return false;
+    }
+
+    // Traverse the list of visible inner fields until we reach the focused field. If we did not encounter a fields with a visible label
```
cguglielmo: Small typo: If we did not encounter a field[s]

bschwarzent:
Go Jenkins

bschwarzent:
Run tests

PR #1330, opened 2025-01-15
IdCodec: change signature delimiter and improve signature assertion

fschinkel:
In some projects IIds are used to transport decimal formats. As these formats can contain multiple # the IId is considered to be signed during deserialization which leads to exceptions because these IIds are typically marked with @IdSignature(false).
To improve the signature assertion split the unqualified id only at the last occurrence of the delimiter and not at every occurrence. Also check that the unqualified id has not changed after splitting at the delimiter if it needs to be unsigned (e.g. an unsigned id ends with the delimiter).

406658


fschinkel:
@paolobazzi As we need to change the delimiter, do we want to use the opportunity to choose a delimiter that does need to be encoded in URLs? What do you think of `~sig~`?

paolobazzi (approved):

PR #1339, opened 2025-01-17
App: reload on any error not just session timeout

cguglielmo:
400259

### DIFF
--- a/eclipse-scout-core/src/App.ts
+++ b/eclipse-scout-core/src/App.ts
@@ -242,7 +242,7 @@ export class App extends EventEmitter {
   }
 
   protected _bootstrapDone(options: AppBootstrapOptions) {
-    webstorage.removeItemFromSessionStorage('scout:timeoutPageReload');
+    webstorage.removeItemFromSessionStorage('scout:bootstrapErrorPageReload');
     this.trigger('bootstrap', {
       options: options
     });
@@ -256,7 +256,7 @@ export class App extends EventEmitter {
    *               - a {@link JsonErrorResponseContainer} if a successful response contained a {@link JsonErrorResponse} which was transformed to an error (e.g. using {@link App.handleJsonError}).
    */
   protected _bootstrapFail(options: AppBootstrapOptions, vararg: AjaxError | JQuery.jqXHR | JsonErrorResponseContainer, textStatus?: JQuery.Ajax.ErrorTextStatus, errorThrown?: string,
-    requestOptions?: JQuery.AjaxSettings): JQuery.Promise<any> | void {
+    requestOptions?: JQuery.AjaxSettings): JQuery.Promise<any> {
     $.log.isInfoEnabled() && $.log.info('App bootstrap failed');
 
     // If one of the bootstrap ajax call fails due to a session timeout, the index.html is probably loaded from cache without asking the server for its validity.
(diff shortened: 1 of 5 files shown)


eclipse-scout-core/src/App.ts
```diff
@@ -266,47 +266,51 @@ export class App extends EventEmitter {
...
-    } else if (objects.isObject(vararg) && vararg.error) {
+    } else if (jsonError) {
       // Json based error
       // Json errors (normally processed by Session.js) are returned with http status 200
-      if (vararg.error.code === Session.JsonResponseError.SESSION_TIMEOUT) {
-        this._handleBootstrapTimeoutError(vararg.error, vararg.url);
-        return;
-      }
+      url = jsonError.url;
+      message = `JSON error code ${jsonError.error.code}`;
```
bschwarzent: Shouldn't we also include the jsonError.error.message?

> Resource res/texts.json could not be loaded, even after a page reload. Reason: JSON error code 10

Also, connection issues are now reported differently:

> **Old:** AJAX call "GET res/texts.json" failed [Connection error]
> **New:** Resource res/texts.json could not be loaded, even after a page reload. Reason: 0

eclipse-scout-core/src/App.ts
```diff
@@ -266,47 +266,51 @@ export class App extends EventEmitter {
...
       // Prevent loop in case reloading did not solve the problem
-      $.log.isWarnEnabled() && $.log.warn('Prevented automatic reload, startup will likely fail', error, url);
-      webstorage.removeItemFromSessionStorage('scout:timeoutPageReload');
-      throw new Error('Resource ' + url + ' could not be loaded due to a session timeout, even after a page reload');
+      $.log.isWarnEnabled() && $.log.warn('Prevented automatic reload, startup will likely fail.');
+      webstorage.removeItemFromSessionStorage('scout:bootstrapErrorPageReload');
+      throw new Error('Resource ' + url + ' could not be loaded, even after a page reload. Reason: ' + message);
     }
-    webstorage.setItemToSessionStorage('scout:timeoutPageReload', true + '');
+    webstorage.setItemToSessionStorage('scout:bootstrapErrorPageReload', true + '');
```
bschwarzent:
```suggestion
    webstorage.setItemToSessionStorage('scout:bootstrapErrorPageReload', 'true');
```

While we are changing the lines in this method, we could generally convert them to template strings.

eclipse-scout-core/src/main.less
```diff
@@ -504,13 +504,16 @@ a,
...
 .startup-error {
+  position: relative;
   #scout.user-select(text);
-  #scout.drop-shadow();
   margin: 100px 50px;
   padding: 20px;
-  border: 1px solid @error-border-color;
   background-color: @error-background-color;
-  border-radius: @border-radius-large;
+  border-radius: @border-radius-medium;
```
bschwarzent, line 512: Padding should not be uniform on all sides, does not loook good.

grafik

eclipse-scout-core/src/App.ts
```diff
@@ -266,47 +266,51 @@ export class App extends EventEmitter {
...
+  protected _handleBootstrapError(url: string, message: string) {
+    $.log.isInfoEnabled() && $.log.info('Error for resource ' + url + '. Reloading page...');
+    if (webstorage.getItemFromSessionStorage('scout:bootstrapErrorPageReload')) {
       // Prevent loop in case reloading did not solve the problem
-      $.log.isWarnEnabled() && $.log.warn('Prevented automatic reload, startup will likely fail', error, url);
-      webstorage.removeItemFromSessionStorage('scout:timeoutPageReload');
-      throw new Error('Resource ' + url + ' could not be loaded due to a session timeout, even after a page reload');
+      $.log.isWarnEnabled() && $.log.warn('Prevented automatic reload, startup will likely fail.');
+      webstorage.removeItemFromSessionStorage('scout:bootstrapErrorPageReload');
+      throw new Error('Resource ' + url + ' could not be loaded, even after a page reload. Reason: ' + message);
```
bschwarzent: "even after a page reload": Is this information relevant? I would only log this, but not show it to the user.

PR #1344, opened 2025-01-21
Fix PlainText calculation for html with ">" in tag attribute value

nsteger:
Fix regex for html tags such that > inside quotes (tag attribute value) are ignored

406821


nsteger:
Cherry pick from https://github.com/eclipse-scout/scout.rt/pull/1340

PR #1350, opened 2025-01-27
Features/bsh/25.1/378432 cell editor popup

bschwarzent:


cguglielmo (commented):
This enhancement also improves the LookupColumn so that pressing enter in the filter field no longer closes the lookup box popup. Could you mention that in the commit comment as another example?

PR #1355, opened 2025-01-30
UiNotifications: wait for subscribe

cguglielmo:
The subscribe request needs to be awaited to avoid race conditions and to improve error handling.

Example what happens without waiting for subscribe():
1. Call subscribe(). A notification for that topic would reload a local cache.
2. Subscribe request is still pending.
3. Load data to fill a local cache.
4. Publish a notification to invalidate the cache.
4. Subscribe request now ends. -> The notification is not received because subscribe() finished after
   the notification was published. The local cache is not invalidated
   and contains old data.

Error handling:
Subscribe now rejects the promise if the request fails. Previously, the request was retried but may have failed again resulting in an unsuccessful subscriptions -> notification for that topic were never received.

Also: UiNotificationsMock now registers a local system which makes it possible to simulate publishing notifications without needing ajax mocks.

408087

### DIFF
--- a/eclipse-scout-core/src/testing/uinotification/UiNotificationsMock.ts
+++ b/eclipse-scout-core/src/testing/uinotification/UiNotificationsMock.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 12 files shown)


eclipse-scout-core/src/testing/uinotification/UiNotificationsMock.ts
```diff
@@ -26,19 +29,34 @@ export class UiNotificationsMock {
 
     ObjectFactory.get().unregister(UiNotificationSystem);
   }
+
+  static putNotification(notification: SomePartial<UiNotificationDo, 'id' | 'nodeId' | 'creationTime'>, system?: string) {
+    system = system || System.MAIN_SYSTEM;
+    let systemObj = uiNotifications.systems.get(system);
+    if (!(systemObj instanceof LocalUiNotificationSystem)) {
```
bschwarzent, line 42: Alternatively, we could use `scout.assertInstanceOf()` here.

eclipse-scout-core/src/testing/uinotification/UiNotificationsMock.ts
```diff
@@ -26,19 +29,34 @@ export class UiNotificationsMock {
...
+export class LocalUiNotificationSystem extends UiNotificationSystem {
 
-  override subscribe(topic: string, handler: UiNotificationHandler): JQuery.Promise<string> {
+  override whenSubscriptionStart(topic: string): JQuery.Promise<string> {
     return $.resolvedPromise(topic);
   }
 
-  override subscribeOne(topic: string, handler: UiNotificationHandler): JQuery.Promise<string> {
-    return $.resolvedPromise(topic);
+  put(notification: SomePartial<UiNotificationDo, 'id' | 'nodeId' | 'creationTime'>) {
```
bschwarzent, line 63: Move this custom method to the end, so both overwritten methods are close to each other.

eclipse-scout-core/src/testing/uinotification/UiNotificationsMock.ts
```diff
@@ -26,19 +29,34 @@ export class UiNotificationsMock {
 
     ObjectFactory.get().unregister(UiNotificationSystem);
   }
+
+  static putNotification(notification: SomePartial<UiNotificationDo, 'id' | 'nodeId' | 'creationTime'>, system?: string) {
```
bschwarzent, line 39: A short description would be helpful.

eclipse-scout-core/src/uinotification/UiNotificationSystem.ts
```diff
@@ -52,13 +58,20 @@ export class UiNotificationSystem implements UiNotificationSystemModel, ObjectWi
       return $.resolvedPromise(topic);
     }
     let deferred = $.Deferred();
+    let errorHandler = event => {
```
bschwarzent: Maybe declare `errorHandler` after `handler`? (Same order as they are registered as event handlers.)

eclipse-scout-core/src/uinotification/UiNotificationSystem.ts
```diff
@@ -52,13 +58,20 @@ export class UiNotificationSystem implements UiNotificationSystemModel, ObjectWi
...
+      deferred.reject(event.error);
+      this.poller.off('subscriptionStart', handler);
+      this.poller.off('error', errorHandler);
+    };
     let handler = event => {
```
bschwarzent: Alternative name: `subscriptionStartHandler` (consistent with event names)

eclipse-scout-core/src/uinotification/UiNotificationPoller.ts
```diff
@@ -17,10 +17,15 @@ export class UiNotificationPoller extends PropertyEventEmitter {
   declare eventMap: UiNotificationPollerEventMap;
   declare self: UiNotificationPoller;
 
+  /**
+   * Configures in milliseconds the time to wait after an error occurs, before the polling will be retried.
+   */
+  static RESPONSE_ERROR_RETRY_INTERVAL = 30000;
```
bschwarzent:
```suggestion
  static RESPONSE_ERROR_RETRY_INTERVAL = 30_000;
```

eclipse-scout-core/src/uinotification/UiNotificationPoller.ts
```diff
@@ -199,24 +207,34 @@ export class UiNotificationPoller extends PropertyEventEmitter {
...
+      }
+      this.setStatus(BackgroundJobPollingStatus.FAILURE);
+
+      // If poller is supposed to run, reschedule it
+      this._schedulePoll(UiNotificationPoller.RESPONSE_ERROR_RETRY_INTERVAL);
       return;
     }
     this.setStatus(BackgroundJobPollingStatus.FAILURE);
 
     if (scout.isOneOf(error.jqXHR.status, 401, 403)) {
```
bschwarzent, line 229: Do we need something like _AjaxCall.isAuthError()_ here (analog to `isOfflineError`)?

eclipse-scout-core/src/uinotification/UiNotificationPoller.ts
```diff
@@ -17,10 +17,15 @@ export class UiNotificationPoller extends PropertyEventEmitter {
...
+  /**
+   * Configures in milliseconds the time to wait after an error occurs, before the polling will be retried.
+   */
+  static RESPONSE_ERROR_RETRY_INTERVAL = 30000;
+
   /**
    * Configures in milliseconds the time to wait after a connection error occurs, before the polling will be retried.
    */
-  static RETRY_INTERVAL = 5000;
+  static CONNECTION_ERROR_RETRY_INTERVALS = [300, 500, 1000, 5000];
```
bschwarzent, line 28: Why are these constants public, but `DEFAULT_BACKEND_TIMEOUT` and `BACKEND_TIMEOUT_OFFSET` are protected? Should all constants be protected?

bschwarzent:
During application startup, the poller is restarted multiple times, as more and more topics are subscribed with each additional bootstrapper. When the previous call is aborted, the `_onError` method is called. It should do nothing in case the call was already restarted. However, this does not work properly when the "Call" class encountered an offline error and automatically scheduled a retry. In this case, the `pendingCall` property is not set. Instead, we should also check for the scheduled call via `callTimeoutId`.

Suggested change:
```
   protected _onError(error: AjaxError) {
-    if (this._call.pendingCall) {
+    if (this._call.pendingCall || this._call.callTimeoutId) {
       // Poller has already been restarted (likely after an abort) -> ignore error and don't reschedule poll
       return;
     }
-    if (error.textStatus === 'abort') {
+    if (error.textStatus === 'abort' || this._call.aborted) {
       // Don't report errors if polling was aborted
       if (this.status === BackgroundJobPollingStatus.STOPPED) {
         return;
```

eclipse-scout-core/src/uinotification/UiNotificationPoller.ts
cguglielmo: This would be the only usage, so I don't think it is necessary

eclipse-scout-core/src/testing/uinotification/UiNotificationsMock.ts
cguglielmo: It would need a new variable:

`    let localSystem = scout.assertInstance(systemObj, LocalUiNotificationSystem, 'You need to register the local system first using UiNotificationsMock.registerLocalSystem');
    localSystem.put(notification);`
    
I actually prefer the current solution.

PR #1356, opened 2025-01-30
HybridManager: add context element converters for table rows and columns

bschwarzent:
- Add a 'col:' and 'row:' prefix to element ids to be able to identify the correct converter on the other side.
- Add utility methods to add and remove prefixes/suffixes from strings.

389630

### DIFF
--- a/eclipse-scout-core/src/util/strings.ts
+++ b/eclipse-scout-core/src/util/strings.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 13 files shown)


eclipse-scout-core/src/util/strings.ts
```diff
@@ -379,20 +379,36 @@ export const strings = {
     return strings.equals(a, b, true);
   },
 
+  /**
+   * Adds the given prefix to the start of the given string and returns the result.
+   * If either of the given arguments is null or empty, the original string is returned unchanged, i.e. without prefix.
+   */
+  addPrefix(string: string, prefix: string): string {
+    return string && prefix ? prefix + string : string;
```
fschinkel, line 387: why do we not use _box_ here?

eclipse-scout-core/src/util/strings.ts
```diff
@@ -379,20 +379,36 @@ export const strings = {
...
+  },
+
+  /**
+   * Adds the given suffix to the end of the given string and returns the result.
+   * If either of the given arguments is null or empty, the original string is returned unchanged, i.e. without suffix.
+   */
+  addSuffix(string: string, suffix: string): string {
+    return string && suffix ? string + suffix : string;
```
fschinkel, line 395: why do we not use _box_ here?

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/util/StringUtility.java
```diff
@@ -1655,6 +1655,38 @@ public static String decompress(byte[] compressed) {
     return out.toString();
   }
 
+  /**
+   * Adds the given prefix to the start of the given string and returns the result.
+   * If either of the given arguments is null or empty, the original string is returned unchanged, i.e. without prefix.
+   */
+  public static String addPrefix(String s, String prefix) {
+    return isNullOrEmpty(s) || isNullOrEmpty(prefix) ? s : prefix + s;
```
fschinkel, line 1663: why do we not use _box_ here?

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/util/StringUtility.java
```diff
@@ -1655,6 +1655,38 @@ public static String decompress(byte[] compressed) {
...
+  }
+
+  /**
+   * Adds the given suffix to the end of the given string and returns the result.
+   * If either of the given arguments is null or empty, the original string is returned unchanged, i.e. without suffix.
+   */
+  public static String addSuffix(String s, String suffix) {
+    return isNullOrEmpty(s) || isNullOrEmpty(suffix) ? s : s + suffix;
```
fschinkel, line 1671: why do we not use _box_ here?

fschinkel (approved):

eclipse-scout-core/src/util/strings.ts
bschwarzent: Because StringUtility#box returns `""` when the original string is `null`. This is somewhat useful when the result is directly concatenated to another string. But otherwise, the "is-null" information about the original string is lost.

I think these methods more accurately represent the operation we want to perform. `box` should only be used when a string should be surrounded by two other strings (usually some kind of brackets), i.e. a combination of `addPrefix` and `addSuffix`. Adding a string only on one end by adding "nothing" on the other side feels unnecessarily complicated.

Also, because of the symmetry with `removePrefix`/`removeSufffix`.

PR #1362, opened 2025-01-31
Menu: show text as tooltip for shrunk menus

bschwarzent:
400639

### DIFF
--- a/eclipse-scout-core/src/action/Action.ts
+++ b/eclipse-scout-core/src/action/Action.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -278,6 +278,17 @@ export class Action extends Widget implements ActionModel {
   }
 
   /**
+   * Returns the text to show as a tooltip, which might be different from the `tooltipText` property.
+   * If this value is falsy, the tooltip is not installed.
+   *
+   * @see _configureTooltip
+   * @see _shouldInstallTooltip
+   */
+  protected _computeTooltipText(): string {
+    return this.tooltipText;
+  }
+
+  /**
    * Installs or uninstalls tooltip based on tooltipText, selected and enabledComputed.
    */
   protected _updateTooltip() {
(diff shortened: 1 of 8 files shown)


eclipse-scout-core/src/action/Action.ts
```diff
@@ -286,14 +290,14 @@ export class Action extends Widget implements ActionModel {
     } else {
       tooltips.uninstall(this.$container);
     }
-    aria.description(this.$container, this.tooltipText);
+    aria.description(this.$container, this._tooltipText());
```
cguglielmo: Maybe screen readers now read the same text twice because the aria-label and description contain the text. I am not sure about that, needs to be tested. Maybe we should only add the description if it differs from the text.

eclipse-scout-core/src/action/Action.ts
```diff
@@ -277,6 +277,10 @@ export class Action extends Widget implements ActionModel {
     this._updateTooltip();
   }
 
+  protected _tooltipText(): string {
```
cguglielmo: Alternative name: _computeTooltipText

eclipse-scout-core/src/menu/Menu.ts
```diff
@@ -731,4 +732,15 @@ export class Menu extends Action implements MenuModel {
     }
     return false;
   }
+
+  protected override _tooltipText(): string {
```
cguglielmo: Can you move the declaration of textVisibleOrig  from Action to Menu? It seems to be at the wrong place.

eclipse-scout-core/src/menu/Menu.ts
```diff
@@ -731,4 +732,15 @@ export class Menu extends Action implements MenuModel {
...
+    if (this._showTextAsTooltip()) {
+      return strings.join('\n\n', this.text, this.tooltipText);
+    }
+    return super._tooltipText();
+  }
+
+  protected _showTextAsTooltip(): boolean {
```
cguglielmo, line 758: Please add a comment or jsdoc what the purpose of this function is (e..g same as in _renderText )

cguglielmo (commented):
Can you add Jasmine Specs?

I think I would move the code from MenuBoxLayout/MenuBarLayout to Action and Menu to simplify testing. 
Action.makeCompact/undoMakeCompact
Menu.shrink/undoShrink

Maybe keep suppressInvalidate in the Layout because it only needs to be suppressed because it is called while layouting.

eclipse-scout-core/src/action/Action.ts
bschwarzent: Correct. Aria attributes should not be affected by the shrinked state.

bschwarzent:
Replaced by https://github.com/eclipse-scout/scout.rt/pull/1530

PR #1381, opened 2025-02-11
Column: set sortActive if column model declares a sortIndex

andibur:
Otherwise, the ColumnHeader shows no sort indicator and sortAscending false is ignored.

410150

### DIFF
--- a/eclipse-scout-core/src/table/columns/Column.ts
+++ b/eclipse-scout-core/src/table/columns/Column.ts
@@ -108,7 +108,7 @@ export class Column<TValue = string> extends PropertyEventEmitter implements Col
     this.primaryKey = false;
     this.mandatory = false;
     this.optimalWidthMeasurer = new ColumnOptimalWidthMeasurer(this);
-    this.sortActive = false;
+    this.sortActive = null;
     this.sortAscending = true;
     this.sortIndex = -1;
     this.summary = false;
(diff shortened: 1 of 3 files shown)


eclipse-scout-core/src/table/columns/Column.ts
```diff
@@ -161,6 +161,7 @@ export class Column<TValue = string> extends PropertyEventEmitter implements Col
     icons.resolveIconProperty(this, 'headerIconId');
     this._setTable(this.table);
     this._setAutoOptimizeWidth(this.autoOptimizeWidth);
+    this.sortActive = this.sortIndex >= 0;
```
cguglielmo: sortActive can be passed by the model -> set this.sortActive to null in the constructor and use scout.nvl(this.sortActive, this.sortIndex >= 0) here. Please also adjust the jsdoc in ColumnModel.ts for sortActive to something like this: Default is true if {@link sortIndex} is >= 0, otherwise it is false.

eclipse-scout-core/test/table/TableSpec.ts
```diff
@@ -1348,6 +1348,25 @@ describe('Table', () => {
       $header2 = $colHeaders.eq(2);
     }
 
+    it('applies initial sort spec from model', () => {
+      model = helper.createModelFixture(3, 3);
+      model.columns[0].sortIndex = 0;
+      model.columns[0].sortAscending = false;
+      adapter = helper.createTableAdapter(model);
```
cguglielmo: Please create adapter only when testing adapter specific behavior, e.g. in TableAdapterSpec.ts.
Remove variables for column1 and 2 as they are not needed.
Replace table.columns[0] with column0 or remove column0.
Assert that sortActive stays false if a column passes sortActive = false in its init model

PR #1385, opened 2025-02-12
Add boolean id and date id

tgygax:


matthiaso (approved):
Reminder: Currently freeze for 24.2

PR #1397, opened 2025-02-20
Form: don't close a form created by the ui server

cguglielmo:
Use case:
- A form is created by the hybrid manager
- The form is used as a detail form of a page
- Reloading the table page deletes the nodes which destroys the form.
- The same response contains a property change event referencing the form that was closed by JS when nodes were deleted -> form cannot be resolved anymore. -> Only the server is allowed to close the form.

In a specific case the problematic form was used as a child of a detail form. The form was explicitly closed using form.close(). A detail form is currently closed using destroy() by Page.ts. This should be adjusted in the future.

The change in HybridManager.ts ensures that the close event is not triggered twice if the form was closed by JS. Normally, closing the form with Java only triggers destroy() on JS side. The hybrid manager triggers close() explicitly to make it possible to listen for the close event for forms opened using the hybrid manager.

409972

### DIFF
--- a/eclipse-scout-core/src/desktop/hybrid/HybridManager.ts
+++ b/eclipse-scout-core/src/desktop/hybrid/HybridManager.ts
@@ -120,7 +120,7 @@ export class HybridManager extends Widget {
     widget.trigger(eventType, {data});
   }
 
-  protected _onHybridFormEvent(form: Form, eventType: string, data: object) {
+  protected _onHybridFormEvent(form: HybridForm, eventType: string, data: object) {
     if (eventType === 'reset') {
       form.setData(data);
       form.trigger('reset');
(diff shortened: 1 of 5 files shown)


bschwarzent (approved):

eclipse-scout-core/src/desktop/hybrid/HybridManager.ts
```diff
@@ -220,3 +229,10 @@ export class HybridManager extends Widget {
     return super.when(type as K);
   }
 }
+
+interface HybridForm extends Form {
+  /**
+   * @returns true if {@link FormEventMap.close} event was triggered at least once for this form.
+   */
+  __closeTriggered: boolean;
```
fschinkel: This is an optional property, isn't it? And when this is changed to optional you do not need to change the casts.

PR #1400, opened 2025-02-21
Allow access to AbstractPageWithTable#isLimitedResult property

ralph-steiner:
Used for cases where data is loaded not with an AbstractTablePageData


bschwarzent (approved):
🥇

PR #1408, opened 2025-03-04
Improve formatting of DateTimePeriodFormatter

dwl-bsi:
In order to prevent confusion with time instants, the DateTimePeriodFormatter should format time periods in a format which is distinct from the format of local (date) times.

Previously, the format was "hh:mm:ss.zzz" (e.g. 12:34:56) which was confused with the corresponding time of day. The new format makes this distinction clear by adding time units (e.g. 12h 34m 56s).

389916


org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/util/date/DateTimePeriodFormatter.java
```diff
@@ -34,62 +34,114 @@ public class DateTimePeriodFormatter {
...
+    long minutes = duration.toMinutesPart();
+    long seconds = duration.toSecondsPart();
+
+    // the highest non-zero unit can be displayed without padding-zeros
+    // all lower units are displayed with padding zeroes, even if they are zero
+    // this is done so that two different periods formatted by this algorithm can be easily compared when stacked on top of one another
+    // because the digits of the same time-unit always appear in the same position of the text
+    boolean showDays = days > 0;
+    boolean showHours = days > 0 || hours > 0;
+    boolean showMinutes = days > 0 || hours > 0 || minutes > 0;
```
nsteger: Maybe a little bit more readable:
```
    boolean showHours = showDays || hours > 0;
    boolean showMinutes = showHours || minutes > 0;
```
?

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/util/date/DateTimePeriodFormatter.java
```diff
@@ -34,62 +34,117 @@ public class DateTimePeriodFormatter {
...
+   *     If this argument is negative, the result is the same as if zero was passed.<br>
+   *     If this argument is null, the result is also null.
+   * @param includeMilliseconds
+   *     If {@code true}, the milliseconds part of the duration is included in the output.
+   * @return The period formatted in a human-readable text
    */
-  public String formatTimePeriod(BigDecimal d) {
-    if (d == null) {
+  public String formatDuration(Duration duration, boolean includeMilliseconds) {
+    // KEEP IN SYNC WITH eclipse-scout-core/src/util/dates.ts#formatDuration(number, boolean)
```
nsteger: dates.ts#formatDuration(number, boolean, locale)

eclipse-scout-core/src/util/dates.ts
```diff
@@ -465,6 +465,69 @@ export const dates = {
...
+   * @param includeMillis
+   *          If this flag is true, the milliseconds part is included in the text
+   * @param locale
+   *          The locale that should be used to format the text
+   * @returns The time duration as formatted text
+   * @see DateTimePeriodFormatter.java
+   */
+  formatDuration(durationInMilliseconds: number, includeMillis: boolean, locale: Locale): string {
+    // KEEP IN SYNC WITH org.eclipse.scout.rt.platform.util.date.DateTimePeriodFormatter.formatDuration(java.time.Duration, boolean)
+    if (durationInMilliseconds === null || durationInMilliseconds === undefined) {
```
nsteger, line 485: To make it easier to keet it in sync, I would prefer to write it as similar as possible. i.e.:

```
  formatDuration(durationInMilliseconds: number, includeMilliseconds: boolean, locale: Locale): string {
    // KEEP IN SYNC WITH org.eclipse.scout.rt.platform.util.date.DateTimePeriodFormatter.formatDuration(java.time.Duration, boolean)
    if (durationInMilliseconds === null || durationInMilliseconds === undefined) {
      return null;
    }
    if (durationInMilliseconds < 0) {
      durationInMilliseconds = 0;
    }
    let time = durationInMilliseconds;
    let milliseconds = time % 1000;
    time = Math.floor(time / 1000);
    let seconds = time % 60;
    time = Math.floor(time / 60);
    let minutes = time % 60;
    time = Math.floor(time / 60);
    let hours = time % 24;
    let days = Math.floor(time / 24);

    // the highest non-zero unit can be displayed without padding-zeros
    // all lower units are displayed with padding zeroes, even if they are zero
    // this is done so that two different periods formatted by this algorithm can be easily compared when stacked on top of one another
    // because the digits of the same time-unit always appear in the same position of the text
    let showDays = days > 0;
    let showHours = showDays || hours > 0;
    let showMinutes = showHours || minutes > 0;
    let result = '';
    if (showDays) {
      result = days + ' ' + (days > 1 ? texts.resolveText('${textKey:ui.DaysUnit}', locale.languageTag) : texts.resolveText('${textKey:ui.DayUnit}', locale.languageTag)) + ' ';
    }
    if (showHours) {
      // because the days-part has different lengths ('2 days' vs '1 day'), there is no need to pad the hours-part with zeros
      // as the digits of the day cannot possibly align anyway. Make the text more readable by leaving out this leading zero.
      result += hours + 'h ';
    }
    if (showMinutes) {
      result += (showHours ? strings.padZeroLeft(minutes, 2) : minutes) + 'm ';
    }
    // seconds are always shown
    result += showMinutes ? strings.padZeroLeft(seconds, 2) : seconds;
    if (includeMilliseconds) {
      let decimalSeparator = locale.decimalFormatSymbols.decimalSeparator;
      result += decimalSeparator + strings.padZeroLeft(milliseconds, 3);
    }
    result += 's';
    return result;
  },
```

PR #1424, opened 2025-03-12
ServiceTunnel: use signed ids when necessary

fschinkel:
DOs might only be known on the backend server and are sent to the UI server and then to the browser where they are used in rest calls again. In this case the id needs to be signed on the backend server as the UI server has no idea which parts of the DO are ids and which are not. Therefore, introduce ServiceTunnelOptions that can be used to parametrize service tunnel proxies.
If the idSignature flag of the ServiceTunnelOptions is set serializing and deserializing the request and response on both sides of the tunnel is running in a run context that has a specific property set. If this property is set, the data object wrappers created by the ServiceTunnelObjectReplacer are using a signature object mapper. The TunnelToServer annotation does support the idSignature flag as well. If set the proxy created for the interface is initialized with the flag accordingly.

413664

### DIFF
--- a/org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/ServiceTunnelServlet.java
+++ b/org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/ServiceTunnelServlet.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -13,6 +13,7 @@ import static org.eclipse.scout.rt.server.commons.opentelemetry.SpanNamePropagat
 
 import java.io.IOException;
 import java.security.AccessController;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.function.LongPredicate;
 
@@ -47,8 +48,10 @@ import org.eclipse.scout.rt.server.context.RunMonitorCancelRegistry.IRegistratio
 import org.eclipse.scout.rt.server.context.ServerRunContext;
 import org.eclipse.scout.rt.server.context.ServerRunContexts;
 import org.eclipse.scout.rt.shared.servicetunnel.IServiceTunnelContentHandler;
+import org.eclipse.scout.rt.shared.servicetunnel.ServiceTunnelOptions;
 import org.eclipse.scout.rt.shared.servicetunnel.ServiceTunnelRequest;
 import org.eclipse.scout.rt.shared.servicetunnel.ServiceTunnelResponse;
+import org.eclipse.scout.rt.shared.servicetunnel.http.HttpServiceTunnel;
 import org.eclipse.scout.rt.shared.ui.UserAgents;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -128,6 +131,7 @@ public class ServiceTunnelServlet extends AbstractHttpServlet {
       m_serverRunContextProducer.get()
           .getInnerRunContextProducer()
           .produce(servletRequest, servletResponse)
+          .withProperties(enableSignature(servletRequest) ? Map.of(ServiceTunnelOptions.ID_SIGNATURE_PROP, true) : Map.of())
           .run(() -> {
             ServiceTunnelRequest serviceRequest = deserializeServiceRequest();
             ServiceTunnelResponse serviceResponse = doPost(serviceRequest);
@@ -177,6 +181,14 @@ public class ServiceTunnelServlet extends AbstractHttpServlet {
         servletResponse.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
       }
     }
+  }
+
+  /**
+   * Check the {@link HttpServletRequest} if signature creation needs to be enabled. Default implementation checks
+   * the {@link HttpServiceTunnel#ID_SIGNATURE_HTTP_HEADER}.
+   */
+  protected boolean enableSignature(HttpServletRequest servletRequest) {
+    return Boolean.TRUE.toString().equalsIgnoreCase(servletRequest.getHeader(HttpServiceTunnel.ID_SIGNATURE_HTTP_HEADER));
   }
 
   protected ServiceTunnelResponse doPost(ServiceTunnelRequest serviceRequest) {
--- a/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/TunnelToServer.java
+++ b/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/TunnelToServer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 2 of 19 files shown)


org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/servicetunnel/AbstractServiceTunnel.java
```diff
@@ -30,12 +33,25 @@
...
+  public static final String ID_SIGNATURE_PROP = "ServiceTunnel.idSignature";
 
   @Override
   public Object invokeService(Class<?> serviceInterfaceClass, Method operation, Object[] callerArgs) {
     LOG.debug("{}.{}({})", serviceInterfaceClass, operation, callerArgs);
     ServiceTunnelRequest request = createRequest(serviceInterfaceClass, operation, callerArgs);
     interceptRequest(request);
+
+    // check the TunnelToServer annotation if idSignature is set to true
+    if (Optional.ofNullable(serviceInterfaceClass)
```
fschinkel: @paolobazzi do we want to move this code into the `ServiceTunnelProxyProducer`?

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/ServiceTunnelServlet.java
```diff
@@ -128,6 +130,7 @@ protected void doPost(HttpServletRequest servletRequest, HttpServletResponse ser
       m_serverRunContextProducer.get()
           .getInnerRunContextProducer()
           .produce(servletRequest, servletResponse)
+          .withProperty(AbstractServiceTunnel.ID_SIGNATURE_PROP, enableSignature(servletRequest))
```
fschinkel: @mvilliger how do we test this new behavior end to end?

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/TunnelToServer.java
```diff
@@ -29,4 +30,14 @@
...
+  /**
+   * Whether the interface uses signature creation for serialization/deserialization of {@link IDoEntity}s in the service tunnel.
+   * When enabled the transferred {@link IDoEntity}s are serialized using signature creation and therefore the transferred data changes.
+   * <p>
+   * Consider a {@link IDoEntity} which is transferred from the backend server to the ui server and afterward to the browser where it might be used for rest calls directly to the browser.
+   * In this case the serialization between the ui server and the browser and between the browser and the rest endpoint use signatures.
+   * If now the type of the {@link IDoEntity} is only known to the backend and not the ui server the serialization between the backend and the ui server needs to use signatures or otherwise the rest endpoint will not accept the
+   * {@link IDoEntity} as it contains unsigned attributes.
+   */
+  boolean idSignature() default false;
```
fschinkel, line 37: @fschinkel add release notes

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/ServiceTunnelServlet.java
```diff
@@ -128,6 +130,7 @@ protected void doPost(HttpServletRequest servletRequest, HttpServletResponse ser
       m_serverRunContextProducer.get()
           .getInnerRunContextProducer()
           .produce(servletRequest, servletResponse)
+          .withProperty(ServiceTunnelOptions.ID_SIGNATURE_PROP, enableSignature(servletRequest))
```
mvilliger: Maybe we could skip setting the property if it is false? Not having the property means the same as false, therefore there is no need to add it to the map in most cases?

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/servicetunnel/ServiceTunnelProxyProducer.java
```diff
@@ -47,10 +57,22 @@ public Object invoke(T instance /*will always be null*/, Method method, Object[]
...
 
+    // check the idSignature flag
+    if (getOptions().isIdSignature()) {
+      // add run context property
+      return RunContexts.copyCurrent()
+          .withProperty(ID_SIGNATURE_PROP, true)
+          .call(() -> BEANS.get(IServiceTunnel.class).invokeService(getInterfaceClass(), method, args));
+    }
+
     return BEANS.get(IServiceTunnel.class).invokeService(getInterfaceClass(), method, args);
```
mvilliger: Maybe extract this line to a callable variable instead of duplicating?

mvilliger (commented):
Looks good, thank you. Please see the comments and add test cases if possible.

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/servicetunnel/ServiceTunnelOptions.java
```diff
@@ -0,0 +1,52 @@
...
+
+  public static ServiceTunnelOptions create() {
+    return new ServiceTunnelOptions();
+  }
+
+  /**
+   * Whether the service tunnel uses signature creation for serialization/deserialization of {@link IDoEntity}s or not.
+   * When enabled the transferred {@link IDoEntity}s are serialized using signature creation and therefore the transferred data changes.
+   * <p>
+   * Consider a {@link IDoEntity} which is transferred from the backend server to the ui server and afterward to the browser where it might be used for rest calls directly to the browser.
```
paolobazzi: 'directly to the browser.'
-> shouldn't this be:
'directly to a REST endpoint in the backend server.'

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/servicetunnel/ServiceTunnelObjectReplacer.java
```diff
@@ -55,6 +63,12 @@ public Object resolveObject(Object obj) {
     if (obj instanceof DoEntityWrapper) {
       return ((DoEntityWrapper) obj).getDoEntity();
     }
+    if (obj instanceof DoEntityHolderWrapper) {
```
paolobazzi, line 66: DoEntityHolder is a subclass of DataObjectHolder -> DoEntityHolderWrapper should not be necessary.

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/servicetunnel/ServiceTunnelOptions.java
```diff
@@ -0,0 +1,52 @@
...
+  public static final String ID_SIGNATURE_PROP = "ServiceTunnelOptions.idSignature";
+
+  private Boolean m_idSignature;
+
+  public static ServiceTunnelOptions create() {
+    return new ServiceTunnelOptions();
+  }
+
+  /**
+   * Whether the service tunnel uses signature creation for serialization/deserialization of {@link IDoEntity}s or not.
```
paolobazzi: Could also be a IDataObject (e.g. toplevel list) instead of a plain IDoEntity

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/TunnelToServer.java
```diff
@@ -29,4 +31,8 @@
 @Retention(RetentionPolicy.RUNTIME)
 public @interface TunnelToServer {
 
+  /**
+   * Whether the interface uses signature creation for serialization/deserialization of {@link IDoEntity}s in the service tunnel (see {@link ServiceTunnelOptions#withIdSignature(Boolean)}).
```
paolobazzi: better IDataObject instead of IDoEntity?

paolobazzi (approved):
See inputs for minor improvements.

PR #1427, opened 2025-03-14
Property for partitioned session-cookie

egloff:
Add new config property 'scout.app.sessionCookieConfigPartitioned' which controls the partitioned-flag on the session-cookie.

326352

### DIFF
--- a/org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ApplicationProperties.java
+++ b/org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ApplicationProperties.java
@@ -172,9 +172,9 @@ public final class ApplicationProperties {
 
     @Override
     public String description() {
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ApplicationProperties.java
```diff
@@ -209,6 +209,25 @@ protected SameSite parse(String value) {
...
+    }
+
+    @Override
+    public Boolean getDefaultValue() {
+      return false;
+    }
+
+    @Override
+    public String description() {
+      return "Specifies whether any session tracking cookies created by the web application will be marked as partitioned. "
```
cguglielmo: I would rephrase it like so: Specifies whether the HTTP session cookie created  by the web... The term tracking is confusing. Please also adjust the description for scout.app.sessionCookieConfigSecure accordingly and remove the word 'tracking'

PR #1431, opened 2025-03-18
JsFormAdapter: Prefer jsFormModel data over undefined input data

gabrielstalder:
Adjust handling of falsy parameters in data object utility
- Adjusts serialize such that it returns the unaltered given data object
parameter if it is falsy.
- Adjusts stringify to return the data objects string  representation
even if the data object is falsy.
- Adjusts parse such that it returns null/undefined if the given json is
null or undefined. If the given json is an empty string, null is
returned.
- Adjusts deserialize to return the unaltered given object if it is
falsy.

389583

### DIFF
--- a/eclipse-scout-core/src/dataobject/dataObjects.ts
+++ b/eclipse-scout-core/src/dataobject/dataObjects.ts
@@ -38,11 +38,7 @@ export const dataObjects = {
    * @returns the JSON string.
    */
   stringify(dataObject: any): string {
-    const serialized = dataObjects.serialize(dataObject);
-    if (!serialized) {
-      return null;
-    }
-    return JSON.stringify(serialized);
+    return JSON.stringify(dataObjects.serialize(dataObject));
   },
 
   /**
@@ -55,11 +51,11 @@ export const dataObjects = {
    * * Properties with value 'undefined' are not part of the result (skipped).
    *
    * @param dataObject The value to serialize. Can be primitives, arrays or objects/classes. Typically, a pojo or data object class extending {@link BaseDoEntity}.
-   * @returns the input serialized. Typically, a pojo.
+   * @returns the input serialized. Typically, a pojo. Returns the unaltered input if the input is falsy.
    */
   serialize(dataObject: any): any {
     if (!dataObject) {
-      return null;
+      return dataObject;
     }
     return scout.create(DataObjectSerializer).serialize(dataObject);
   },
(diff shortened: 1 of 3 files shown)


eclipse-scout-core/src/dataobject/dataObjects.ts
```diff
@@ -40,7 +40,7 @@ export const dataObjects = {
   stringify(dataObject: any): string {
     const serialized = dataObjects.serialize(dataObject);
     if (!serialized) {
-      return null;
+      return dataObject;
```
mvilliger: - Here it should be "return serialized"?
- The parse() function below should be adapted as well?
- Please add a Spec for this utility to ensure falsy inputs stay the same
- And the commit message seems no longer to be correct?

gabrielstalder: - Adjusted to always return the string representation even if dataObject is falsy
- Adjusted
- Added specs in dataObjectSpec.ts
- Adjusted commit message

Thanks for the review!

eclipse-scout-core/test/dataobject/dataObjectsSpec.ts
```diff
@@ -163,6 +163,81 @@ describe('dataObjects', () => {
...
+      expect(parsedObject.value).toBe(13);
+      expect(dataObjects.parse('')).toBe(null);
+    });
+    it('null and undefined', () => {
+      expect(dataObjects.parse(null)).toBe(null);
+      expect(dataObjects.parse(undefined)).toBe(undefined);
+    });
+  });
+
+  describe('deserialize', () => {
```
mvilliger, line 227: In this block the deserialize function should be tested? Instead of parse?

gabrielstalder: Correct, I've adjusted the test.

eclipse-scout-core/test/dataobject/dataObjectsSpec.ts
```diff
@@ -163,6 +163,81 @@ describe('dataObjects', () => {
...
+    it('strings', () => {
+      let parsedObject = dataObjects.parse('{' +
```
mvilliger: Still uses parse here. Is this intended?

gabrielstalder: adjusted

mvilliger (approved):

PR #1434, opened 2025-03-18
Decouple ListBox value type from lookup type

crispin-kirchner:
Improve ListBox extensibility by removing the constraint on the value, which was fixed to an array of the lookup type.

#402055

### DIFF
--- a/eclipse-scout-core/src/form/fields/LookupBox.ts
+++ b/eclipse-scout-core/src/form/fields/LookupBox.ts
@@ -13,17 +13,17 @@ import {
 } from '../../index';
 import $ from 'jquery';
 
-export abstract class LookupBox<TValue> extends ValueField<TValue[], TValue | TValue[]> implements LookupBoxModel<TValue> {
-  declare model: LookupBoxModel<TValue>;
-  declare eventMap: LookupBoxEventMap<TValue>;
-  declare self: LookupBox<any>;
+export abstract class LookupBox<TLookup, TValue = TLookup[]> extends ValueField<TValue, TLookup | TValue> implements LookupBoxModel<TLookup, TValue> {
+  declare model: LookupBoxModel<TLookup, TValue>;
+  declare eventMap: LookupBoxEventMap<TLookup, TValue>;
+  declare self: LookupBox<TLookup, TValue>;
 
   filterBox: Widget;
-  lookupCall: LookupCall<TValue>;
-  codeType: string | (new() => CodeType<TValue>);
+  lookupCall: LookupCall<TLookup>;
+  codeType: string | (new() => CodeType<TLookup>);
   lookupStatus: Status;
 
-  protected _currentLookupCall: LookupCall<TValue>;
+  protected _currentLookupCall: LookupCall<TLookup>;
   protected _lookupExecuted: boolean;
   /** true when value is either syncing to table or table to value */
   protected _valueSyncing: boolean;
@@ -33,7 +33,7 @@ export abstract class LookupBox<TValue> extends ValueField<TValue[], TValue | TV
     this.filterBox = null;
     this.gridDataHints.weightY = 1.0;
     this.gridDataHints.h = 2;
-    this.value = [];
+    this.value = [] as TValue;
     this.clearable = ValueField.Clearable.NEVER;
     this.lookupCall = null;
     this._currentLookupCall = null;
@@ -56,7 +56,7 @@ export abstract class LookupBox<TValue> extends ValueField<TValue[], TValue | TV
     }
   }
 
-  protected override _initValue(value: TValue[]) {
+  protected override _initValue(value: TValue) {
     if (this.lookupCall) {
       this._setLookupCall(this.lookupCall);
     }
@@ -65,7 +65,7 @@ export abstract class LookupBox<TValue> extends ValueField<TValue[], TValue | TV
     super._initValue(value);
   }
 
-  protected abstract _initStructure(value: TValue[]);
+  protected abstract _initStructure(value: TValue);
 
   protected override _render() {
     this.addContainer(this.$parent, 'lookup-box');
@@ -97,16 +97,12 @@ export abstract class LookupBox<TValue> extends ValueField<TValue[], TValue | TV
     this.filterBox.render(this.$fieldContainer);
   }
 
-  protected override _ensureValue(value: TValue | TValue[]): TValue[] {
-    return arrays.ensure(value);
-  }
-
-  protected _lookupByAll(): JQuery.Promise<LookupResult<TValue>> {
+  protected _lookupByAll(): JQuery.Promise<LookupResult<TLookup>> {
     if (!this.lookupCall) {
       return;
     }
 
-    let deferred = $.Deferred<LookupResult<TValue>>();
+    let deferred = $.Deferred<LookupResult<TLookup>>();
 
     this._executeLookup(this.lookupCall.cloneForAll(), true)
       .done(result => {
@@ -117,7 +113,7 @@ export abstract class LookupBox<TValue> extends ValueField<TValue[], TValue | TV
     return deferred.promise();
   }
 
-  protected _executeLookup(lookupCall: LookupCall<TValue>, abortExisting?: boolean): JQuery.Promise<LookupResult<TValue>> {
+  protected _executeLookup(lookupCall: LookupCall<TLookup>, abortExisting?: boolean): JQuery.Promise<LookupResult<TLookup>> {
     this.setLoading(true);
 
     if (abortExisting && this._currentLookupCall) {
@@ -138,7 +134,7 @@ export abstract class LookupBox<TValue> extends ValueField<TValue[], TValue | TV
       });
   }
 
-  protected _lookupByAllDone(result: LookupResult<TValue>) {
+  protected _lookupByAllDone(result: LookupResult<TLookup>) {
     try {
       if (result.exception) {
         // Oops! Something went wrong while the lookup has been processed.
@@ -174,11 +170,11 @@ export abstract class LookupBox<TValue> extends ValueField<TValue[], TValue | TV
   }
 
   /** @see LookupBoxModel.lookupCall */
-  setLookupCall(lookupCall: LookupCallOrModel<TValue>) {
+  setLookupCall(lookupCall: LookupCallOrModel<TLookup>) {
     this.setProperty('lookupCall', lookupCall);
   }
 
-  protected _setLookupCall(lookupCall: LookupCallOrModel<TValue>) {
+  protected _setLookupCall(lookupCall: LookupCallOrModel<TLookup>) {
     this._setProperty('lookupCall', LookupCall.ensure(lookupCall, this.session));
     this._lookupExecuted = false;
     if (this.rendered) {
@@ -187,16 +183,16 @@ export abstract class LookupBox<TValue> extends ValueField<TValue[], TValue | TV
   }
 
   /** @see LookupBoxModel.codeType */
-  setCodeType(codeType: string | (new() => CodeType<TValue>)) {
+  setCodeType(codeType: string | (new() => CodeType<TLookup>)) {
     this.setProperty('codeType', codeType);
   }
 
-  protected _setCodeType(codeType: string | (new() => CodeType<TValue>)) {
+  protected _setCodeType(codeType: string | (new() => CodeType<TLookup>)) {
     this._setProperty('codeType', codeType);
     if (!codeType) {
       return;
     }
-    let lookupCall = scout.create(CodeLookupCall<TValue>, {
+    let lookupCall = scout.create(CodeLookupCall<TLookup>, {
       session: this.session,
       codeType: codeType
     });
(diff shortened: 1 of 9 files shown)


eclipse-scout-core/src/form/fields/listbox/ListBox.ts
```diff
@@ -7,21 +7,35 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-import {
-  arrays, Column, InitModelOf, ListBoxLayout, ListBoxModel, ListBoxTableAccessibilityRenderer, LookupBox, lookupField, LookupResult, LookupRow, ObjectOrChildModel, objects, scout, Table, TableModel, TableRowModel, TableRowsCheckedEvent,
-  Widget
-} from '../../../index';
+import {LookupBox} from '../LookupBox';
```
bschwarzent: Don't import all files individually, instead import them all from `'../../../index'`

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
```diff
@@ -7,21 +7,35 @@
...
+import {Column} from '../../../table/columns/Column';
+import {ListBoxModel} from './ListBoxModel';
 
-export class ListBox<TValue> extends LookupBox<TValue> implements ListBoxModel<TValue> {
-  declare model: ListBoxModel<TValue>;
+export class ListBox<TLookup, TValue = TLookup[]> extends LookupBox<TLookup, TValue> {
+  declare model: ListBoxModel<TLookup, TValue>;
 
   table: Table;
+  hideTableLines: boolean;
```
bschwarzent: * Declare and document this property on the `ListBoxModel`.
* Consider renaming to "tableLinesVisible" (with default value `false`). This is more inline with other Scout properties, e.g. `borderVisible`. (The CSS class can still be named `hide-table-lines`.)

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
```diff
@@ -49,6 +63,18 @@ export class ListBox<TValue> extends LookupBox<TValue> implements ListBoxModel<T
     this.$container.addClass('list-box');
   }
 
+  protected override _renderProperties() {
+    super._renderProperties();
+    this._renderHideTableLines();
+  }
+
+  protected _renderHideTableLines() {
+    if (!this.$container) {
```
bschwarzent: This check is not necessary. It is safe to assume that `this.$container` is present in all _render{prop} methods.

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
```diff
@@ -49,6 +63,18 @@ export class ListBox<TValue> extends LookupBox<TValue> implements ListBoxModel<T
     this.$container.addClass('list-box');
   }
 
+  protected override _renderProperties() {
+    super._renderProperties();
+    this._renderHideTableLines();
+  }
+
+  protected _renderHideTableLines() {
```
bschwarzent: Please move this method after the corresponding setter (so all property-related methods are together).

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
```diff
@@ -58,23 +84,45 @@ export class ListBox<TValue> extends LookupBox<TValue> implements ListBoxModel<T
...
-      return;
-    }
-    this._valueSyncing = true;
-    let valueArray = [];
+  protected _valueToTable(newValue0: TValue) {
+    let opts = {checkOnlyEnabled: false};
+
+    let rowsToCheck = [];
+
+    const newValue = newValue0 as TLookup[];
```
bschwarzent: How is this safe? Why not set the argument type accordingly?

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
```diff
@@ -58,23 +84,45 @@ export class ListBox<TValue> extends LookupBox<TValue> implements ListBoxModel<T
...
   }
 
-  protected _syncTableToValue() {
-    if (!this.lookupCall || this._valueSyncing) {
-      return;
-    }
-    this._valueSyncing = true;
-    let valueArray = [];
+  protected _valueToTable(newValue0: TValue) {
+    let opts = {checkOnlyEnabled: false};
```
bschwarzent: Why use `const` for `newValue` but not for the other variables? -> be consistent

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
```diff
@@ -58,23 +84,45 @@ export class ListBox<TValue> extends LookupBox<TValue> implements ListBoxModel<T
...
+      if (arrays.containsAny(newValue, row.lookupRow.key)) {
+        rowsToCheck.push(row);
+      }
+    }, this);
+    this.table.checkRows(rowsToCheck, opts);
+  }
+
+  protected _tableToValue(): TValue {
+    const valueArray: TLookup[] = [];
     this.table.rows.forEach(row => {
```
bschwarzent: Consider using map()

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
```diff
@@ -58,23 +84,45 @@ export class ListBox<TValue> extends LookupBox<TValue> implements ListBoxModel<T
...
       if (row.checked) {
         valueArray.push(row.lookupRow.key);
       }
     });
+    return valueArray as TValue;
```
bschwarzent: Are the constant casts throughout this class really necessary?

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
```diff
@@ -83,34 +131,31 @@ export class ListBox<TValue> extends LookupBox<TValue> implements ListBoxModel<T
     this._syncValueToTable(this.value);
   }
 
-  protected _syncValueToTable(newValue: TValue[]) {
+  protected _isEmpty(value: TValue): boolean {
+    return arrays.empty(value as TLookup[]);
+  }
+
+  protected _resetTable() {
```
bschwarzent, line 94: This method seems to be called only once - necessary?

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
```diff
@@ -119,14 +164,18 @@ export class ListBox<TValue> extends LookupBox<TValue> implements ListBoxModel<T
...
 
-  protected _populateTable(result: LookupResult<TValue>) {
-    let
-      tableRows = [],
+  protected _initTableRows(): TableRowModel[] {
+    return [];
+  }
+
+  protected _populateTable(result: LookupResult<TLookup>) {
+    const
```
bschwarzent: Consider writing this shorter:
```
    const tableRows = result.lookupRows.map(lookupRow => this._createTableRow(lookupRow));
```

eclipse-scout-core/src/form/fields/listbox/ListBox.less
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+.list-box.hide-table-lines {
```
bschwarzent: Maybe merge the three first lines to reduce the nesting level?

eclipse-scout-core/src/form/fields/LookupBox.ts
```diff
@@ -13,17 +13,17 @@ import {
 } from '../../index';
 import $ from 'jquery';
 
-export abstract class LookupBox<TValue> extends ValueField<TValue[], TValue | TValue[]> implements LookupBoxModel<TValue> {
-  declare model: LookupBoxModel<TValue>;
-  declare eventMap: LookupBoxEventMap<TValue>;
+export abstract class LookupBox<TLookup, TValue = TLookup[]> extends ValueField<TValue, TLookup | TValue> implements LookupBoxModel<TLookup, TValue> {
+  declare model: LookupBoxModel<TLookup, TValue>;
+  declare eventMap: LookupBoxEventMap<TLookup, TValue>;
   declare self: LookupBox<any>;
```
bschwarzent: No generic type here?

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiTextContributor.java
```diff
@@ -30,6 +30,7 @@ public void contribute(Set<String> textKeys) {
         "ColumnSorting",
         "CorrelationId",
         "DateIsNotAllowed",
+        "DefaultValue",
```
bschwarzent: Where is this text used?

bschwarzent (changes requested):

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
crispin-kirchner: done

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
crispin-kirchner: done

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
crispin-kirchner: done

eclipse-scout-core/src/form/fields/listbox/ListBox.less
crispin-kirchner: changed accordingly

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
crispin-kirchner: changed accordingly

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
crispin-kirchner: changed accordingly

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
crispin-kirchner: changed accordingly

eclipse-scout-core/src/form/fields/LookupBox.ts
crispin-kirchner: changed accordingly

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiTextContributor.java
crispin-kirchner: moved to the layer where it's actually used

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
crispin-kirchner: I don't see how this can be combined with the call to this._initTableRows()

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
crispin-kirchner: it's offered as an extension point

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
crispin-kirchner: see the other comment about casts

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
bschwarzent: Ah, I see. You want to allow subclasses to provide additional rows by overriding `_initTableRows`. But this only allows for a very specific use case to add custom rows at the front of the list. Consider a method `_createTableRows` instead. This allows the subclass to add or remove rows anywhere.

```
  protected _createTableRows(lookupRows: LookupRow<TLookup>[]): ObjectOrModel<TableRow>[] {
    return lookupRows.map(lookupRow => this._createTableRow(lookupRow));
  }

  protected _populateTable(result: LookupResult<TLookup>) {
    const tableRows = this._createTableRows(result.lookupRows);
    this.table.deleteAllRows();
    this.table.insertRows(tableRows);
    this._syncValueToTable(this.value);
  }
```

Example use in subclass:
```
  protected override _createTableRows(lookupRows: LookupRow<TLookup>[]): ObjectOrModel<TableRow>[] {
    return [
       scout.create(TableRow, {cells:['First row']}),
       ...super._createTableRows(lookupRows),
       scout.create(TableRow, {cells:['Last row']}),
    ];
  }
```

Alternative idea: Allow a listener to manipulate the lookup result before it is actually converted into table rows. In Java, this would be possible by overriding `AbstractListBox#execFilterLookupResult`. In JS, we don't have this method. In SmartField.ts, we can achieve this effect by listening for the `lookupCallDone` event. This has the additional benefit that such a listener can be added externally without the need for a subclass.

crispin-kirchner: Thanks for the hint! I settled with _lookupByAllDone and removed the hook method

eclipse-scout-core/src/form/fields/listbox/ListBox.ts
crispin-kirchner: isolated the method needing casts to a separate section of the class

PR #1455, opened 2025-04-02
Decouple LookupBox value type from lookup type

rogersteblerbsi:
Improve LookupBox extensibility by removing the constraint on the value, which was fixed to an array of the lookup type.

370000


rogersteblerbsi:
@bschwarzent this is basically a part of the PR that you already reviewed, that was closed again: https://github.com/eclipse-scout/scout.rt/pull/1434
I only need the decoupling for the LookupBox, not for the ListBox. So it's only a part of the original PR.

PR #1463, opened 2025-04-08
Fix NullPointerException in BinaryResource

nsteger:
java.lang.String.String(byte[], Charset) expects not null arguments, fix this call in BinaryResource.java

363702


christiandoebele:
review ok

PR #1465, opened 2025-04-10
DoEntityBuilder: improve predicate type

bschwarzent:
Use generic type to set type of predicate argument to same type as the value to put. This allows using method references as predicate.

Example: result.putIf("name", myName, StringUtility::hasText)


paolobazzi (approved):
Looks good for me! 
Add a note in the release notes about the minimal API change of putIf.

bschwarzent:
> Looks good for me! Add a note in the release notes about the minimal API change of putIf.

https://github.com/eclipse-scout/scout.docs/pull/300

PR #1468, opened 2025-04-10
MenuBox: prevent endless layout

bschwarzent:
When using the newest open sans font, the menu box may be layouted forever when the screen is small and one menu item contains an image.

Reason:
If the pref width of the menu box value is fractional (e.g. 310.00000762939453) and the browser rounds it down when the size is set to the menu box, the prefSize (menusWidth in layout()) will be bigger than the containerSize -> menu box collapses even though it should only be shrinked.

When the menu box is layouted by the desktop header, the header tries to figure out which mode is necessary for the menu box (compact, shrinked or collapsed).
The desktop header and the menu box need to agree on the mode but if the desktop header tries set the menu box in shrink mode and the menu box will instead collapse itself, the layout won't be deterministic. And because one menu has an image and invalidateLayoutTree() is called when the image loads, the layout loops forever.

414936


bschwarzent:
Run tests

PR #1469, opened 2025-04-11
DataObjects: add support for do type resolvers

cguglielmo:
Sometimes it is desired to use a specific BaseDoEntity class if the data object _type cannot be resolved rather than the generic BaseDoEntity. In that case a custom do type resolver can be added.

### DIFF
--- a/eclipse-scout-core/src/dataobject/serialize/DataObjectDeserializer.ts
+++ b/eclipse-scout-core/src/dataobject/serialize/DataObjectDeserializer.ts
@@ -38,7 +38,7 @@ export class DataObjectDeserializer implements DataObjectDeserializerModel, Obje
   }
 
   protected _deserializeObject<T extends object>(rawObj: Record<string, any>, metaData?: DoValueMetaData<T>): T {
-    const constructor = (doValueMetaData.chooseDataObjectType(rawObj, metaData) || BaseDoEntity) as Constructor<T>;
+    const constructor = this._resolveDataObjectType(rawObj, metaData) as Constructor<T>;
     const resultObj = this._createResultObject(constructor);
     const proto = Object.getPrototypeOf(constructor).prototype;
     Object.keys(rawObj)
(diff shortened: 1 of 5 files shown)


eclipse-scout-core/src/dataobject/serialize/DataObjectDeserializer.ts
```diff
@@ -49,6 +49,16 @@ export class DataObjectDeserializer implements DataObjectDeserializerModel, Obje
     return resultObj;
   }
 
+  protected _resolveDataObjectType(rawObj: Record<string, any>, metaData: DoValueMetaData): Constructor {
```
mvilliger, line 52: metaData is optional in _deserializeObject. Therefore it should be optional in _resolveDataObjectType and in DoTypeResolver.resolve as well?

eclipse-scout-core/src/dataobject/serialize/DoTypeResolver.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+import {Constructor, DoValueMetaData, doValueMetaData} from '../../index';
+
+export interface DoTypeResolver {
+  resolve(rawObj: Record<string, any>, metaData: DoValueMetaData): Constructor;
+}
+
+export class DefaultDoTypeResolver implements DoTypeResolver {
+  resolve(rawObj: Record<string, any>, metaData: DoValueMetaData): Constructor {
+    return doValueMetaData.chooseDataObjectType(rawObj, metaData);
```
mvilliger, line 18: DataObjectSerializer._serializeObject uses chooseDataObjectType as well. Is this resolver logic not necessary there?
Maybe chooseDataObjectType should loop over the resolvers itself?

cguglielmo: No, it is only necessary for deserialization

cguglielmo:
already merged

PR #1472, opened 2025-04-11
Improve logging in HttpProxy, reduce usage of logging proxy

paolobazzi:
Using logging proxy for all invoked methods causes a large overhead especially for methods which are invoked many times while handling large requests, e.g.
- org.apache.hc.core5.http.nio.AsyncDataConsumer.consume
- org.apache.hc.core5.http.nio.AsyncDataConsumer.updateCapacity

This change removes the logging proxy for all classes which are directly implemented in HttpProxy class and changes all log statements into a log method which dynamically adds correlation id to the MDC before emiting the log statement.

The proxied handler attached in DefaultAsyncHttpClientManager and H2AsyncHttpClientManager are newly guarded by a check statement and just added if log level is <= DEBUG.

416950

### DIFF
--- a/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/async/DefaultAsyncHttpClientManager.java
+++ b/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/async/DefaultAsyncHttpClientManager.java
@@ -42,6 +42,8 @@ import org.eclipse.scout.rt.shared.http.HttpConfigurationProperties.ApacheHttpTr
 import org.eclipse.scout.rt.shared.http.HttpConfigurationProperties.ApacheHttpTransportRetryOnSocketExceptionByConnectionResetProperty;
 import org.eclipse.scout.rt.shared.http.proxy.ConfigurableProxySelector;
 import org.eclipse.scout.rt.shared.http.retry.CustomHttpRequestRetryStrategy;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.metrics.Meter;
@@ -62,6 +64,8 @@ import io.opentelemetry.api.metrics.Meter;
  * @see ConfigurableProxySelector
  */
 public class DefaultAsyncHttpClientManager extends AbstractAsyncHttpClientManager<HttpAsyncClientBuilder> {
+
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultAsyncHttpClientManager.class);
 
   /**
    * @return Technical transport manager name used for metrics.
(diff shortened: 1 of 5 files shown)


org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/async/DefaultAsyncHttpClientManager.java
```diff
@@ -85,11 +89,14 @@ protected void interceptCreateClient(HttpAsyncClientBuilder builder) {
...
-        scope,
-        createAsyncInvocationHandler(AsyncExecCallback.class, asyncExecCallback)));
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("adding execution interceptor which adds context information for logging purposes.");
+      // adding execution interceptor allowing access to entityProducer and asyncExecCallback to add context information e.g. for logging purposes
+      builder.addExecInterceptorFirst(AsyncHttpInvocationHandler.class.getSimpleName(), (request, entityProducer, scope, chain, asyncExecCallback) -> chain.proceed(request,
+          createAsyncInvocationHandler(AsyncEntityProducer.class, entityProducer),
+          scope,
+          createAsyncInvocationHandler(AsyncExecCallback.class, asyncExecCallback)));
+    }
```
matthiaso: As far as I understand the `AsyncExecChainHandler` is not called as ofter as some other methods are, I would have choosen an implementation similar to:

```java
    // adding execution interceptor allowing access to entityProducer and asyncExecCallback to add context information e.g. for logging purposes
    builder.addExecInterceptorFirst(AsyncHttpInvocationHandler.class.getSimpleName(), (request, entityProducer, scope, chain, asyncExecCallback) -> {
      boolean debugEnabled = LOG.isDebugEnabled();
      chain.proceed(request,
          debugEnabled ? createAsyncInvocationHandler(AsyncEntityProducer.class, entityProducer) : entityProducer,
          scope,
          debugEnabled ? createAsyncInvocationHandler(AsyncExecCallback.class, asyncExecCallback) : asyncExecCallback);
    });
```

Especially for productive systems the restarts may be quite bothersome (it is the UI server after all, all logged in users are losing their current state).

But I understand there are also advantages for this implementation.

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/async/H2AsyncHttpClientManager.java
```diff
@@ -48,11 +52,14 @@ protected H2AsyncClientBuilder createBuilder() {
...
-        scope,
-        createAsyncInvocationHandler(AsyncExecCallback.class, asyncExecCallback)));
+    if (LOG.isDebugEnabled()) {
+      LOG.debug("adding execution interceptor which adds context information for logging purposes.");
+      // adding execution interceptor allowing access to entityProducer and asyncExecCallback to add context information e.g. for logging purposes
+      builder.addExecInterceptorFirst(AsyncHttpInvocationHandler.class.getSimpleName(), (request, entityProducer, scope, chain, asyncExecCallback) -> chain.proceed(request,
+          createAsyncInvocationHandler(AsyncEntityProducer.class, entityProducer),
+          scope,
+          createAsyncInvocationHandler(AsyncExecCallback.class, asyncExecCallback)));
+    }
```
matthiaso: See same comment for `org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/async/DefaultAsyncHttpClientManager.java`

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/async/DefaultAsyncHttpClientManager.java
paolobazzi: Changed to inline isDebugEnabled evaluation

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/async/H2AsyncHttpClientManager.java
paolobazzi: Changed to inline isDebugEnabled evaluation

matthiaso (approved):

PR #1476, opened 2025-04-11
TableMaxResultsHelper: remove MaxRowCountContributionDo if not used

fschinkel:
The data object the MaxRowCountContributionDo is added to may contain an old version of the contribution. If no contribution is necessary because max row count is 0 or undefined this old contribution needs to be removed.


mvilliger (approved):
Maybe add a unit test?

PR #1481, opened 2025-04-14
Table: add rowAction

fschinkel:
The rowAction is an Action that is executed when the rowAction event is triggered and default is not prevented.

416422

### DIFF
--- a/eclipse-scout-core/src/table/Table.ts
+++ b/eclipse-scout-core/src/table/Table.ts
@@ -129,6 +129,7 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
   defaultMenuTypes: string[];
   accessibilityRenderer: AbstractTableAccessibilityRenderer;
   organizer: TableOrganizer;
+  defaultRowAction: Action;
 
   $data: JQuery;
   $emptyData: JQuery;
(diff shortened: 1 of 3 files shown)


eclipse-scout-core/src/table/Table.ts
```diff
@@ -2907,7 +2911,12 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
     if (!row || !column) {
       return;
     }
-    this._triggerRowAction(row, column);
+
+    const event = this.trigger('rowAction', {column, row});
```
bschwarzent, line 2915: Why not leave this in _triggerRowAction? Seems to be a common pattern.

eclipse-scout-core/test/table/TableSpec.ts
```diff
@@ -1042,6 +1043,25 @@ describe('Table', () => {
       expect(jasmine.Ajax.requests.count()).toBe(0);
     });
 
+    it('executes row action if default is not prevented', () => {
```
bschwarzent: Add a test for the more common use case where one of the defined menus should be the default row action.

eclipse-scout-core/src/table/Table.ts
```diff
@@ -129,6 +129,7 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
   defaultMenuTypes: string[];
   accessibilityRenderer: AbstractTableAccessibilityRenderer;
   organizer: TableOrganizer;
+  rowAction: Action;
```
bschwarzent: When I first saw this I was a bit confused because the new property and the existing event have exactly the same name. Maybe "defaultRowAction" would be more appropriate? It would indicate that the action is only executed if the event is not prevented.

eclipse-scout-core/src/table/Table.ts
```diff
@@ -2289,13 +2300,6 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
     });
   }
 
-  protected _triggerRowAction(row: TableRow, column: Column<any>) {
```
bschwarzent, line 2292: Removing this and adding the new property needs documentation in the release notes.

eclipse-scout-core/src/table/Table.ts
fschinkel: This is not the pattern if the event is used afterwards.

PR #1483, opened 2025-04-16
Add NumberField to SliderField

reu24:
Add NumberField to SliderField

Two new properties:
- `valueEditable`: Whether there is a label or a number field to the right of the slider.
- `tabbable`: Makes this widget tabbable.

414023


eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -7,18 +7,46 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-import {InitModelOf, scout, Slider, SliderFieldModel, ValueField} from '../../../index';
-import $ from 'jquery';
+import {FormField, HtmlComponent, InitModelOf, NumberField, objects, PropertyChangeEvent, scout, Slider, SliderFieldModel, ValueField} from '../../../index';
 
-export class SliderField extends ValueField<number> {
+export class SliderField extends NumberField implements SliderFieldModel {
   declare model: SliderFieldModel;
```
bschwarzent: Use consistent order of properties in all places. I suggest making separate groups (one for inherited model properties, one for sliderfield model properties, one for widget data, one for internal properties):

minValue
maxValue
...

step
valueEditable
tabbable

slider
$valueLabel
fieldHtmlComp

_syncingValue

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -7,18 +7,46 @@
...
 
+  $valueLabel: JQuery<HTMLSpanElement>;
+  $fieldHtmlComp: HtmlComponent;
```
bschwarzent: Naming convention: remove "$" prefix, since this is not a DOM element

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -7,18 +7,46 @@
...
   slider: Slider;
+  syncingValue = false;
```
bschwarzent: This is an internal property -> make protected and add `_` prefix.

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -7,18 +7,46 @@
...
+  step: number;
+  tabbable: boolean;
+  valueEditable: boolean;
+
+  constructor() {
+    super();
+    this.$field = null;
```
bschwarzent: Remove this line (is already initialized in FormField.ts)

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -7,18 +7,46 @@
...
+    this.minValue = 0;
+    this.maxValue = 100;
+    this.tabbable = true;
+    this.valueEditable = true;
+  }
 
   protected override _init(model: InitModelOf<this>) {
     super._init(model);
+
     let sliderOptions: InitModelOf<Slider> = $.extend({
```
bschwarzent: You should **not** pass the entire `model` to the Slider, since these models are not compatible. There are _some_ properties that are the same, but they are passed explicitly (minValue, maxValue, value, step, tabbable).

By removing the $.extend() we also set the slider field as the parent of the slider. This is better because it's the slider field that owns and controls the lifecycle of the slider. Thus, we don't have to manually remove the slider or pass the enabled property.

Also, inline this variable (the type declaration does not have a benefit here).

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -7,18 +7,46 @@
...
-export class SliderField extends ValueField<number> {
+export class SliderField extends NumberField implements SliderFieldModel {
   declare model: SliderFieldModel;
 
+  $valueLabel: JQuery<HTMLSpanElement>;
+  $fieldHtmlComp: HtmlComponent;
   slider: Slider;
+  syncingValue = false;
+  step: number;
+  tabbable: boolean;
```
bschwarzent: This property only affects the "tabbability" of the slider, not the entire field. Therefore I suggest naming it `sliderTabbable`.

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -27,31 +55,115 @@ export class SliderField extends ValueField<number> {
...
     this.addMandatoryIndicator();
-    this._renderSlider();
+
+    let $fieldContainer = this.$container.appendDiv();
+    this.$fieldHtmlComp = HtmlComponent.install($fieldContainer, this.session);
+
+    this.slider.render($fieldContainer);
+    this.slider.on('propertyChange:value', this._onSliderValueChanged.bind(this));
+
+    let $field = $fieldContainer.appendElement('<input>', 'field')
```
bschwarzent: Seems to have been copied from field.makeTextField(). Why not simply use that method?

Alternatively, extract this code in NumberField.ts to a separate method that you can call here.

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -27,31 +55,115 @@ export class SliderField extends ValueField<number> {
...
+      .attr('type', 'text') // So that css rules from main.less are applied
+      .attr('autocomplete', 'NoAutocomplete') /* off and false are currently ignored in Chrome */
+      .disableSpellcheck();
+
+    this.$valueLabel = $fieldContainer
+      .appendSpan('slider-value', this.value + '');
```
bschwarzent: - Use appendDiv instead of appendSpan (DIV is the more generic element, and the 'display' property is overwritten in CSS anyway)
- Don't set the initial value here. This is already done in _renderDisplayText().

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -27,31 +55,115 @@ export class SliderField extends ValueField<number> {
...
+
+    this.addFieldContainer($fieldContainer);
+    this.addField($field);
+    this.addStatus();
+  }
+
+  protected override _remove() {
+    super._remove();
+    this.slider.remove();
```
bschwarzent: This line is not necessary when the creation of the Slider widget is fixed (see _init).

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -27,31 +55,115 @@ export class SliderField extends ValueField<number> {
...
+  }
+
+  protected override _renderProperties() {
+    super._renderProperties();
+    this._renderValueEditable();
+  }
+
+  protected override _renderEnabled() {
+    super._renderEnabled();
+    this.slider.setEnabled(this.enabledComputed);
```
bschwarzent: Setting a model property should not be done in a _render* method.

But when the creation of the Slider widget is fixed (see _init), this is not even necessary anymore, since the enabled state is inherited automatically.

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -27,31 +55,115 @@ export class SliderField extends ValueField<number> {
...
+  }
+
+  protected override _renderDisplayText() {
+    super._renderDisplayText();
+    this.syncDisplayTextToSlider(this.displayText);
+    this.$valueLabel.text(this.displayText);
+  }
+
+  protected _renderValueEditable() {
```
bschwarzent: Use consistent order of methods. The recommended order is, for each property:

- setMyProperty
- _setMyProperty
- _removeMyProperty
- _renderMyProperty

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -27,31 +55,115 @@ export class SliderField extends ValueField<number> {
...
-    return String(this.slider.value);
+  setValueEditable(valueEditable: boolean) {
+    this.setProperty('valueEditable', valueEditable);
   }
 
-  protected override _parseValue(displayText: string): number {
-    // Convert display text back to number
-    return Number(displayText);
+  protected override _setValue(value: number | string) {
+    super._setValue(scout.nvl(value, this.minValue));
```
bschwarzent: This should be done in _validateValue() instead

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -27,31 +55,115 @@ export class SliderField extends ValueField<number> {
...
+
+  _onSliderValueChanged(event: PropertyChangeEvent) {
+    if (!this.syncingValue) {
+      this.setValue(event.newValue);
+    }
+  }
+
+  protected override _setMinValue(minValue: number) {
+    super._setMinValue(minValue);
+    if (this.slider) {
```
bschwarzent: Why should "this.slider" be null? It is always created in _init. Remove this null-check here and in the methods below.

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -27,31 +55,115 @@ export class SliderField extends ValueField<number> {
...
+    this.syncDisplayTextToSlider(this.displayText);
+    this.$valueLabel.text(this.displayText);
+  }
+
+  protected _renderValueEditable() {
+    this.$field.setVisible(this.valueEditable);
+    this.$valueLabel.setVisible(!this.valueEditable);
+  }
+
+  _onSliderValueChanged(event: PropertyChangeEvent) {
```
bschwarzent: Missing "protected" modifier. Check this for all other methods starting with `_` as well.

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -27,31 +55,115 @@ export class SliderField extends ValueField<number> {
...
 
-  override setValue(value: number) {
-    this.slider.setValue(value);
+  protected override _onFieldInput() {
+    this.syncDisplayTextToSlider(this._readDisplayText());
   }
 
-  getValue(): number {
-    return this.slider.value;
+  protected syncDisplayTextToSlider(displayText: string) {
```
bschwarzent: This method is protected, therefore it should start with `_`.

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -27,31 +55,115 @@ export class SliderField extends ValueField<number> {
...
+    try {
+      this.syncingValue = true;
+      this.slider.setValue(this._parseValue(displayText));
```
bschwarzent:
```suggestion
      this.slider.setValue(this.parseValue(displayText));
```

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -27,31 +55,115 @@ export class SliderField extends ValueField<number> {
...
+    } catch (e: any) { /* empty */
```
bschwarzent:
```suggestion
    } catch (error) {
      // nop
```

- No need to declare the error type, since we are not using it
- Code does not work faster if we use shorter variable names :-)
- Avoid /* comments */, it is perfectly fine to use a full line for a line comment

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -27,31 +55,115 @@ export class SliderField extends ValueField<number> {
...
+  }
+
+  protected override _renderEnabled() {
+    super._renderEnabled();
+    this.slider.setEnabled(this.enabledComputed);
+  }
+
+  protected override _renderDisplayText() {
+    super._renderDisplayText();
+    this.syncDisplayTextToSlider(this.displayText);
```
bschwarzent: We should not update model properties in a _render* method. I would remove this line and change the rest of the code slightly. See my comment at the end of this file at _onFieldInput().

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -27,31 +55,115 @@ export class SliderField extends ValueField<number> {
...
-  protected override _parseValue(displayText: string): number {
-    // Convert display text back to number
-    return Number(displayText);
+  protected override _setValue(value: number | string) {
+    super._setValue(scout.nvl(value, this.minValue));
   }
 
-  override setValue(value: number) {
-    this.slider.setValue(value);
+  protected override _onFieldInput() {
```
bschwarzent: I don't like intercepting this method and ignoring the default behavior. In the end, the SliderField should still behave exactly like a NumberField. I assume you did this to immediately update the slider while typing in the number field. This is ok, but we should use the standard mechanisms for that.

I suggest changing the code as follows:
- Remove the call to syncDisplayText() in _renderDisplayText()
- Remove this method
- In the constructor, set `this.updateDisplayTextOnModify = true;`
- In the _init() method, add a listener for the 'displayText' property: `this.on('propertyChange:displayText', event => this.syncDisplayTextToSlider(event.newValue));`
- This adds a small delay before the value is synched, but that may even be beneficial. E.g. when entering numbers with multiple digits, the slider would "jump". If you want, the delay can also be changed by setting the member `this.updateDisplayTextOnModifyDelay`

eclipse-scout-core/src/form/fields/slider/SliderField.less
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+@slider-field-container-margin: 20px;
+
+.slider-field {
+  & > .field {
+    display: flex;
+
```
bschwarzent: Reorder the rules so that they match the DOM order (no functional difference, just easier to read).

eclipse-scout-core/src/form/fields/slider/SliderField.less
```diff
@@ -0,0 +1,39 @@
...
+    & > input {
+      flex: 20%;
+      min-width: 0;
```
bschwarzent: Consider setting a minimal width, e.g. 75px.

eclipse-scout-core/src/form/fields/slider/SliderField.less
```diff
@@ -0,0 +1,39 @@
...
+      text-align: right;
+    }
+
+    & > .slider-value {
+      flex: 1%;
```
bschwarzent: This seems a bit arbitrary. Should we really stretch the value label? If the entire field is very wide, this also gets large and we have a lot of white space.

Consider setting it to "flex: none" and set a sensible min-width, e.g. 50px. It might even be a good idea to introduce one or two CSS variables at .slider-field level, so it can more easily be customized for specific usages.

eclipse-scout-core/src/form/fields/slider/SliderField.less
```diff
@@ -0,0 +1,39 @@
...
+      display: inline-block;
+      vertical-align: middle;
```
bschwarzent: This does not work for vertically centering the text.

I would do it like this:
- Add "align-items: center;" to .slider-field > .field. This aligns all flex items vertically inside the field container.
- Remove "display: inline-block" and "vertical-align: middle".
- Remove the weird "top: 42%" rule from .slider-container.

eclipse-scout-core/src/form/fields/slider/SliderField.less
```diff
@@ -0,0 +1,39 @@
...
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+@slider-field-container-margin: 20px;
```
bschwarzent: Maybe convert this to a CSS variable

eclipse-scout-core/src/index.less
```diff
@@ -77,6 +77,7 @@
 @import "form/fields/radiobutton/RadioButton";
 @import "form/fields/radiobutton/RadioButtonGroup";
 @import "form/fields/sequencebox/SequenceBox";
+@import "form/fields/slider/SliderField";
```
bschwarzent: Rename the folder from "slider" to "sliderfield".

eclipse-scout-core/test/form/fields/slider/SliderFieldSpec.ts
```diff
@@ -40,23 +40,56 @@ describe('SliderField', () => {
...
+      field.setMinValue(0);
+      field.setMaxValue(10);
+      field.setStep(2);
+
+      field.render();
+
+      field.slider.move(2.5);
+      expect(field.value).toBe(2);
+    });
   });
```
bschwarzent, line 120: Add more tests for the new properties/behavior.

eclipse-scout-core/src/form/fields/slider/SliderField.ts
```diff
@@ -27,31 +55,115 @@ export class SliderField extends ValueField<number> {
...
+  protected override _renderDisplayText() {
+    super._renderDisplayText();
+    this.syncDisplayTextToSlider(this.displayText);
+    this.$valueLabel.text(this.displayText);
+  }
+
+  protected _renderValueEditable() {
+    this.$field.setVisible(this.valueEditable);
+    this.$valueLabel.setVisible(!this.valueEditable);
+  }
```
bschwarzent: Since the $field and the $valueLabel are not necessarily of the same width, the size of the slider may change. We have to update its layout to reflect those changes.

Add this to the end of this method:
`this.$fieldHtmlComp.invalidateLayoutTree(false);`

eclipse-scout-core/src/slider/SliderNavigationKeyStroke.ts
```diff
@@ -0,0 +1,58 @@
...
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {keys, KeyStroke, Slider} from '../index';
+
+export class SliderNavigationKeyStroke extends KeyStroke {
+  declare field: Slider;
+
+  protected static PAGE_STEPS = 10;
```
bschwarzent: I think we can safely inline this member

eclipse-scout-core/src/slider/SliderNavigationKeyStroke.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  constructor(field: Slider) {
```
bschwarzent: Rename this parameter to "slide" (only the member has to be called "this.field")

eclipse-scout-core/src/slider/SliderNavigationKeyStroke.ts
```diff
@@ -0,0 +1,58 @@
...
+    super();
+    this.field = field;
+    this.which = [keys.LEFT, keys.RIGHT, keys.PAGE_DOWN, keys.PAGE_UP, keys.HOME, keys.END];
```
bschwarzent, line 18: This KeyStroke is good! Left/Right/Home/End are very intuitive. However, I did not discover the PageUp/PageDown-Keystrokes by myself. I tried Shift-Left/Right or Ctrl-Left/Right to increase the step size, but this did not work. I think it would be more intuitive because the basic function (left/right) stays the same but can be modified by pressing a modifier key. Maybe we can support both?

eclipse-scout-core/src/slider/SliderNavigationKeyStroke.ts
```diff
@@ -0,0 +1,58 @@
...
+      case keys.PAGE_DOWN:
+        this.field.move(-this._calculatePageMoveUnits());
+        break;
+
+      case keys.PAGE_UP:
+        this.field.move(this._calculatePageMoveUnits());
+        break;
+
+      case keys.HOME:
+        this.field.move(Number.MIN_SAFE_INTEGER);
```
bschwarzent: Why not just set the value to minValue/maxValue?

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -42,72 +60,178 @@ export class Slider extends Widget implements SliderModel {
   }
 
   protected override _render() {
-    this.$container = this.$parent.appendDiv('slider');
+    this.$container = this.$parent.appendDiv('slider-container');
```
bschwarzent: Is this necessary? I think it would be more intuitive when the class would just be called "slider".

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -42,72 +60,178 @@ export class Slider extends Widget implements SliderModel {
...
     this.htmlComp = HtmlComponent.install(this.$container, this.session);
     this.htmlComp.setLayout(new SliderLayout(this));
-    this.$sliderInput = this.$container.appendElement('<input>', 'slider-input')
-      .attr('type', 'range')
-      .on('change', this._onValueChange.bind(this))
-      .addClass(Device.get().cssClassForEdge()) as JQuery<HTMLInputElement>;
+    this.$sliderThumb = this.$container.appendElement('<div>', 'slider-thumb') as JQuery<HTMLDivElement>;
```
bschwarzent: Use appendDiv() instead

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -42,72 +60,178 @@ export class Slider extends Widget implements SliderModel {
...
+    this.$sliderTrack = this.$container.appendElement('<div>', 'slider-track') as JQuery<HTMLDivElement>;
+
+    this.$container.on('mousedown', this._mouseDownHandler);
+    document.addEventListener('mousemove', this._mouseMoveHandler);
```
bschwarzent: Only add these events on 'mousedown' (see MoveSupports.ts). Otherwise, they would be active all the time.

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -42,72 +60,178 @@ export class Slider extends Widget implements SliderModel {
...
     }
   }
 
-  protected _renderMaxValue() {
-    if (this.maxValue) {
-      this.$sliderInput.attr('max', this.maxValue);
-    } else {
-      this.$sliderInput.removeAttr('max');
+  protected _onMouseUp() {
+    this._dragging = false;
```
bschwarzent: Similar to _down, delegate to a common method _up. (Maybe rename to _handlePointerUp/_handlePointerDown?)

Alternatively: is it really necessary to have separate methods? Maybe it is enough to attach both event types to the same "mouse" handler. There are some examples of code that does this using the methods in the events.ts utility.

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -42,72 +60,178 @@ export class Slider extends Widget implements SliderModel {
...
-    } else {
-      this.$sliderInput.removeAttr('step');
+  protected _onThumbDragged(mouseX) {
+    const value = this._applyPropertiesOnValue(this._localPositionToValue(mouseX - this.$container.offset().left));
+    this._moveThumb(this._valueToLocalPosition(value));
+    this.setValue(value);
+  }
+
+  protected _moveThumb(position: number) {
+    this.$sliderThumb.cssLeft(position);
```
bschwarzent: This could go wrong if a user starts dragging and then the dialog is closed (e.g. via key stroke). Use `?.` operator or check for `this.rendered` at the start of the method.

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -42,72 +60,178 @@ export class Slider extends Widget implements SliderModel {
...
+    this.$sliderTrack.cssWidth(position);
+  }
+
+  protected override _renderProperties() {
```
bschwarzent: Canonical position of this method is right after _render().

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -42,72 +60,178 @@ export class Slider extends Widget implements SliderModel {
...
+    super._renderProperties();
+    this._renderTabbable();
+  }
+
+  setValue(value: number | string) {
```
bschwarzent: Why should value be a string? Only allow numbers.

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -42,72 +60,178 @@ export class Slider extends Widget implements SliderModel {
...
+
+  setMaxValue(maxValue: number) {
+    this.setProperty('maxValue', maxValue);
+  }
+
+  setTabbable(tabbable: boolean) {
+    this.setProperty('tabbable', tabbable);
+  }
+
+  setValueEditable(valueEditable: boolean) {
```
bschwarzent: Why does this exist here? I think this is only relevant for the SliderField.

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -42,72 +60,178 @@ export class Slider extends Widget implements SliderModel {
...
-      n = scout.nvl(this.maxValue, this.minValue, 0);
+  protected _valueToLocalPosition(value: number) {
+    return (value - this.minValue) / (this.maxValue - this.minValue) * this.$container.cssWidth();
+  }
+
+  protected _localPositionToValue(position: number) {
+    return position / this.$container.cssWidth() * (this.maxValue - this.minValue) + this.minValue;
+  }
+
+  protected _applyPropertiesOnValue(value: number) {
```
bschwarzent: Apply properties _to_ value? Maybe improve this name (it is not clear which properties are ment). Maybe "normalize"?

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -42,72 +60,178 @@ export class Slider extends Widget implements SliderModel {
...
     }
-    this.setValue(n);
+    const stepsFromMin = Math.round((value - this.minValue) / this.step);
+    const steppedValue = stepsFromMin * this.step + this.minValue;
+    return Math.round(steppedValue * Slider.FLOATING_POINT_ERROR_CORRECTION) / Slider.FLOATING_POINT_ERROR_CORRECTION;
   }
 
-  setValue(value: number) {
-    this.setProperty('value', value);
+  protected override _createKeyStrokeContext(): KeyStrokeContext {
```
bschwarzent: Canonical order of those keystroke methods is after _init and before _render.

eclipse-scout-core/src/slider/Slider.less
```diff
@@ -7,137 +7,73 @@
...
  * SPDX-License-Identifier: EPL-2.0
  */
-@thumb-radius: 8px;
-@thumb-size: 16px;
-@thumb-border-width: 1px;
+@thumb-size: 20px;
+@thumb-border-radius: 2px;
+@track-border-radius: 4px;
 @thumb-border-color: @slider-thumb-border-color;
 @thumb-color: @slider-thumb-color;
```
bschwarzent: Are these "alias" colors really necessary? Consider using CSS variables instead to make it easier to change the for certain instances.

eclipse-scout-core/src/slider/Slider.less
```diff
@@ -7,137 +7,73 @@
...
-
-  &::-moz-focus-inner {
-    border: 0;
+@track-disabled-color: @slider-track-disabled-color;
+@thumb-disabled-color: @slider-thumb-disabled-color;
+@thumb-disabled-border-color: @slider-thumb-disabled-border-color;
+
+.slider-container {
+  position: relative;
+  top: 42%;
```
bschwarzent: Yay, 42! 😄

But such a top value is quite unusual. I'd prefer a different solution, see my remark in SliderField.less.

eclipse-scout-core/src/slider/Slider.less
```diff
@@ -7,137 +7,73 @@
...
+  background: @track-after-thumb-color;
+  border-radius: @track-border-radius;
+  user-select: none;
+  height: 8px;
+
+  // for a bigger hit box
+  &:after {
+    content: "";
```
bschwarzent: Our code style uses single quotes in LESS.

eclipse-scout-core/src/slider/Slider.less
```diff
@@ -7,137 +7,73 @@
...
+    position: absolute;
+    top: -14px;
+    width: 100%;
```
bschwarzent: Consider setting all four sides instead of using a "magic number" for the height. That way the height of the slider is only defined in one place. Also, I think 10px is enough.

```
top: -10px;
height: 10px;
left: 0;
right: 0;
```

eclipse-scout-core/src/slider/Slider.less
```diff
@@ -7,137 +7,73 @@
...
-    .thumb-focus();
-  }
+.slider-thumb {
+  position: absolute;
+  width: @thumb-size;
+  height: @thumb-size;
+  background: @thumb-color;
+  border: @thumb-border-radius solid @thumb-border-color;
+  border-radius: 50%;
+  transform: translate(-50%, -100%);
```
bschwarzent: These numbers are weird. Wouldn't we get the same result with this?

```
top: 50%;
transform: translate(-50%, -50%);
```

eclipse-scout-core/src/slider/Slider.less
```diff
@@ -7,137 +7,73 @@
...
+  top: 180%;
+  z-index: 1;
```
bschwarzent: Is this necessary? Just add them in the correct order into the DOM.

eclipse-scout-core/src/slider/Slider.less
```diff
@@ -7,137 +7,73 @@
...
   }
+}
 
-  &::-ms-thumb {
-    .thumb(@thumb-border-width);
-  }
+.slider-track {
+  position: absolute;
+  height: 100%;
+  left: -1px;
```
bschwarzent: Why -1px? Seems like an unnecessary hack.

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -7,19 +7,36 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-import {Device, HtmlComponent, InitModelOf, objects, scout, SliderEventMap, SliderLayout, SliderModel, Widget} from '../index';
+import {Device, HtmlComponent, InitModelOf, KeyStrokeContext, objects, SliderEventMap, SliderLayout, SliderModel, Widget} from '../index';
+import {SliderNavigationKeyStroke} from './SliderNavigationKeyStroke';
```
bschwarzent: Always import other files from the index.ts (except models).

eclipse-scout-core/src/form/fields/slider/SliderFieldModel.ts
```diff
@@ -7,8 +7,27 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-import {SliderModel, ValueFieldModel} from '../../../index';
+import {NumberFieldModel} from '../../../index';
 
-export interface SliderFieldModel extends ValueFieldModel<number>, SliderModel {
-  // all properties are inherited from SliderModel
+export interface SliderFieldModel extends NumberFieldModel {
```
bschwarzent: Some of these model properties can lead to unexpected behavior. See if you can improve this. It is not absolutely necessary that every combination works perfectly, but it should be as robust as possible.

1. When `step` is negative, dragging with the mouse still works normally, but the keys work the other way round. Maybe always use the absolute value?
2. When `step` is 0, the value becomes NaN. Maybe set step to "minValue" if it is 0?
3. If minValue=maxValue, the thumb cannot be dragged. However, it does not seem to sit exactly in the middle of the track.
4. If maxValue is null, any value can be entered but the thumb cannot be dragged.
5. If valueEditable=fase and clearable=ALWAYS or FOCUSED, the "X" is visible but not positioned correct. It might be easiest, if the clear icon was always hidden via CSS. Or maybe overwrite _renderClearable() and do nothing.
6. If valueEditable=false and the field is focused, nothing happens. Maybe overwrite getFocusableElement() and return either the $field or slider.getFocusableElement() depending on the state of the widget.

eclipse-scout-core/src/form/fields/slider/SliderField.less
```diff
@@ -0,0 +1,39 @@
...
+  & > .field {
+    display: flex;
+
+    & > input {
+      flex: 20%;
+      min-width: 0;
+      text-align: right;
+    }
+
+    & > .slider-value {
```
bschwarzent: For consistency reasons, I would name the CSS class "value-label" instead of "slider-value". The name does not have to be globally unique, it is only valid within .slider-field.

eclipse-scout-core/src/form/fields/slider/SliderField.ts
reu24: this.slider is created after super._init is called. Since minValue is a property from NumberField it will be set in super.init which would lead to a NPE.

bschwarzent: Ah, I see. Add a short comment explaining this to prevent future developers from removing this accidentally.

eclipse-scout-core/src/slider/Slider.less
reu24: In my opinion the collision box should be at least as big as the thumb. This is also the collision box of the default HTML slider.
image

This setting would be:
`top: -6px;
    height: 20px;
    left: 0;
    right: 0;`

Which matches `<input type="range">`
image

bschwarzent: Ok! 👍

eclipse-scout-core/src/form/fields/sliderfield/SliderField.less
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+.slider-field {
+
```
bschwarzent: Remove these empty lines before and between the css variables

eclipse-scout-core/src/form/fields/sliderfield/SliderField.ts
```diff
@@ -0,0 +1,168 @@
...
+    let sliderOptions = {
+      parent: this,
+      minValue: this.minValue,
+      maxValue: this.maxValue,
+      value: this.value,
+      step: this.step,
+      tabbable: this.sliderTabbable,
+      valueEditable: this.valueEditable
+    };
+    this.slider = scout.create(Slider, sliderOptions);
```
bschwarzent: Inline the "sliderOptions" variable. This is the usual pattern when creating widget instances:

```
this.slider = scout.create(Slider, {
  parent: this,
  minValue: this.minValue,
  ...
});
```

eclipse-scout-core/src/form/fields/sliderfield/SliderField.ts
```diff
@@ -0,0 +1,168 @@
...
+    let $fieldContainer = this.$container.appendDiv();
+    this.fieldHtmlComp = HtmlComponent.install($fieldContainer, this.session);
+
+    this.slider.render($fieldContainer);
+    this.slider.on('propertyChange:value', this._onSliderValueChanged.bind(this));
+
+    let $field = fields.makeTextField($fieldContainer, 'field');
+    $fieldContainer.append($field);
+
+    this.$valueLabel = $fieldContainer
```
bschwarzent: Remove unncessary line break:

```
    this.$valueLabel = $fieldContainer.appendDiv('slider-value');
```

eclipse-scout-core/src/form/fields/sliderfield/SliderField.ts
```diff
@@ -0,0 +1,168 @@
...
+    this._renderValueEditable();
+  }
+
+  protected override _renderDisplayText() {
+    super._renderDisplayText();
+    this.$valueLabel.text(this.displayText);
+  }
+
+  protected override _renderClearable() {
+  }
```
bschwarzent, line 99: Comment deliberate empty implementation of this method.

For example:
```
  protected override _renderClearable() {
    // nop (slider value should never be clearable)
  }
```

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -41,73 +56,179 @@ export class Slider extends Widget implements SliderModel {
...
+  protected _calculateSteppedValue(value: number) {
+    if (objects.isNullOrUndefined(this.step) || this.step === 0) {
+      return value;
+    }
+    const stepsFromMin = Math.round((value - this.minValue) / this.step);
+    const steppedValue = stepsFromMin * this.step + this.minValue;
+    return Math.round(steppedValue * Slider.FLOATING_POINT_ERROR_CORRECTION) / Slider.FLOATING_POINT_ERROR_CORRECTION;
+  }
+
+  protected override _initKeyStrokeContext() {
```
bschwarzent: Move this method after `_createKeyStrokeContext`

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -41,73 +56,179 @@ export class Slider extends Widget implements SliderModel {
...
     super._renderProperties();
-    this._renderValue();
-    this._renderMinValue();
-    this._renderMaxValue();
-    this._renderStep();
+    this._renderTabbable();
   }
 
   protected override _remove() {
     super._remove();
```
bschwarzent: Super call should be last (opposite direction of _render)

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -41,73 +56,179 @@ export class Slider extends Widget implements SliderModel {
...
+      this._onThumbDragged(clientX);
     }
   }
 
-  protected _onValueChange(event: JQuery.ChangeEvent) {
-    let n = Number(this.$sliderInput.val());
-    // Ensure valid number
-    if (!objects.isNumber(n)) {
-      n = scout.nvl(this.maxValue, this.minValue, 0);
+  protected _onThumbDragged(mouseX) {
```
bschwarzent: - mouseX -> pageX
- Add TS type (number)

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -41,73 +56,179 @@ export class Slider extends Widget implements SliderModel {
...
+    document.removeEventListener('mouseup', this._mouseUpHandler);
+  }
+
+  protected _onTouchUp() {
+    document.removeEventListener('touchmove', this._touchMoveHandler);
+    document.removeEventListener('touchend', this._touchUpHandler);
+  }
+
+  protected _onMouseMove(event: MouseEvent) {
+    this._move(event.clientX);
```
bschwarzent: Use pageX instead of clientX (since you are comparing it with the offset, which is always document-relative)

eclipse-scout-core/src/slider/Slider.less
```diff
@@ -7,137 +7,72 @@
...
-  &:not(.ms-edge):focus::-webkit-slider-thumb {
-    .thumb-focus();
-  }
-
-  /* ---- Firefox ---- */
-
-  &::-moz-range-track {
-    .track();
-    background: @track-color;
+    .slider-thumb {
```
bschwarzent, line 44: Add 'cursor: pointer'

We should probably also add a hitbox ::before element.

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -41,73 +56,179 @@ export class Slider extends Widget implements SliderModel {
...
+  protected _renderStep(step: number) {
+    this.update();
+  }
+
+  setTabbable(tabbable: boolean) {
+    this.setProperty('tabbable', tabbable);
+  }
+
+  protected _renderTabbable() {
+    this.$container.setTabbable(this.tabbable && this.enabledComputed && !Device.get().supportsOnlyTouch());
```
bschwarzent, line 138: Maybe add 'unfocusable' class to $container

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -41,73 +56,179 @@ export class Slider extends Widget implements SliderModel {
...
+  protected _moveThumb(position: number) {
+    this.$sliderThumb?.cssLeft(position);
+    this.$sliderTrack?.cssWidth(position);
+  }
+
+  move(moveBy: number) {
+    this.setValue(this._normalizeValue(this.value as number + moveBy));
+  }
+
+  update() {
```
bschwarzent: Should this be public?

eclipse-scout-core/src/slider/Slider.less
```diff
@@ -7,137 +7,72 @@
...
+  --thumb-disabled-border-color: @slider-thumb-disabled-border-color;
+
+  position: relative;
+  background: var(--track-after-thumb-color);
+  border-radius: var(--track-border-radius);
+  user-select: none;
+  height: 6px;
+
+  // for a bigger hit box
+  &:after {
```
bschwarzent: ::before and ::after pseudo elements should be written with two colons (they are not states like :focus)

Consider changing this to ::before, so it is behind the other elements

eclipse-scout-core/src/slider/SliderNavigationKeyStroke.ts
```diff
@@ -0,0 +1,57 @@
...
+    switch (event.which) {
+      case keys.LEFT:
+        this.field.move(-Math.abs(this.field.step));
+        break;
+
+      case keys.RIGHT:
+        this.field.move(Math.abs(this.field.step));
+        break;
+
+      case keys.PAGE_DOWN:
```
bschwarzent, line 35: Isn't the direction of PageUp/PageDown wrong?

- LEFT: decrease value
- HOME: decrease value
- PAGE_UP: increase value

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -7,19 +7,33 @@
...
+
   value: number;
   minValue: number;
   maxValue: number;
   step: number;
-  $sliderInput: JQuery<HTMLInputElement>;
-  $sliderValue: JQuery<HTMLSpanElement>;
+  tabbable: boolean;
+
+  protected _mouseMoveHandler = this._onMouseMove.bind(this);
```
bschwarzent, line 27: It would be nice, if we could reduce the number of listeners. See attached patch.

[Slider.patch](https://github.com/user-attachments/files/20242795/Slider.patch)

reu24:
Commit message:
Slider: render using \<div> for more precise adjustment

The `Slider` widget is now rendered using multiple `<div>` instead of the browser-native `<input type="range">`.
This allows for a more precise adjustment of the appearance.

The `SliderField` now also allows entering the desired value by hand.
It has two new properties:

* `valueEditable`: Whether there is a label or a number field to the right of the slider.
* `tabbable`: Makes this slider tabbable.

414023

eclipse-scout-core/src/form/fields/sliderfield/SliderField.ts
```diff
@@ -0,0 +1,168 @@
...
+export class SliderField extends NumberField implements SliderFieldModel {
+  declare model: SliderFieldModel;
+  declare eventMap: SliderFieldEventMap;
+
+  step: number;
+  valueEditable: boolean;
+  sliderTabbable: boolean;
+
+  slider: Slider;
+  $valueLabel: JQuery<HTMLSpanElement>;
```
bschwarzent: This says span, but the actual element is a div --> simplify types to just "JQuery" (it's usually not important what kind of DOM element it is, since DIVs are the most generic element anyway).

eclipse-scout-core/src/form/fields/sliderfield/SliderField.less
```diff
@@ -0,0 +1,49 @@
...
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+.slider-field {
+  --slider-margin: 20px;
+
```
bschwarzent: Remove empty lines between css variable definitions.

eclipse-scout-core/src/slider/Slider.less
```diff
@@ -7,137 +7,83 @@
...
+    border-radius: 50%;
+    top: 50%;
+    transform: translate(-50%, -50%);
+
+    // for a bigger hit box
+    &::before {
+      content: '';
+      position: absolute;
+      top: -4px;
+      height: 22px;
```
bschwarzent: Use calc() to avoid hard-coded values

eclipse-scout-core/src/slider/SliderShiftNavigationKeyStroke.ts
```diff
@@ -0,0 +1,39 @@
...
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {keys, KeyStroke, Slider} from '../index';
+import {SliderNavigationKeyStroke} from './SliderNavigationKeyStroke';
+
+export class SliderShiftNavigationKeyStroke extends KeyStroke {
```
bschwarzent: Extend SliderNavigationKeyStroke? calculatePageMoveUnits could then be a protected instance method again.

eclipse-scout-core/src/slider/Slider.ts
```diff
@@ -60,54 +83,146 @@ export class Slider extends Widget implements SliderModel {
...
+    this.setValue(value);
+  }
+
+  protected _setThumbPosition(position: number) {
+    this.$sliderThumb.cssLeft(position);
+    this.$sliderTrack.cssWidth(position);
+  }
+
+  move(moveBy: number) {
+    this.setValue(this._normalizeValue(this.value as number + moveBy));
```
bschwarzent: Why is the cast necessary? (`as number`)

(Same in _update)

bschwarzent (approved):

bschwarzent:
Run tests

bschwarzent:
Run tests

PR #1498, opened 2025-05-02
Reload tiles when form is activated and data was changed

egloff:
Buffer data-change-events on tiles if another form or the desktop is active. As soon the form is activated the buffering is disabled and the buffered events will be delivered to the listeners.

393494

### DIFF
--- a/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/datachange/ActiveFormDataChangeManager.java
+++ b/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/datachange/ActiveFormDataChangeManager.java
@@ -1,0 +1,75 @@
+package org.eclipse.scout.rt.client.ui.desktop.datachange;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.eclipse.scout.rt.client.IClientSession;
+import org.eclipse.scout.rt.client.session.ClientSessionProvider;
+import org.eclipse.scout.rt.client.ui.desktop.DesktopEvent;
+import org.eclipse.scout.rt.client.ui.desktop.IDesktop;
+import org.eclipse.scout.rt.client.ui.form.FormEvent;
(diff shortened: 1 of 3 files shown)


egloff:
go jenkins

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/datachange/ActiveFormDataChangeManager.java
```diff
@@ -0,0 +1,62 @@
...
+ * Data change manager which may be used to avoid data-change notifications for hidden forms.
+ * Events for listeners registered to this manager will be buffered until the form is activated.
+ */
+@ApplicationScoped
+public class ActiveFormDataChangeManager implements IDataChangeListener {
+
+  protected Map<IForm, IDataChangeManager> m_dataChangeManagers = new HashMap<>();
+
+  public ActiveFormDataChangeManager() {
+    IDesktop desktop = IDesktop.CURRENT.get();
```
cguglielmo: ApplicationScoped beans are singletons, the constructor is called only once for every session -> This will work only the first time.
You could move the dataChangeManagers to the clientSession. Then add a getter on this bean getDataChangeManagers () and create the map and set it using clientSession.setData the first time the getter is called for a session.

cguglielmo (commented):
Typo in commit comment:  As soon the form -> As soon as the form
Also, a test case would be nice.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/datachange/ActiveFormDataChangeManager.java
egloff: Thanks, I didn't consider that. The change-managers are now set to the client-sessions. Additionally a test ensures, the events are only delivered to the currently active form.

cguglielmo (approved):

egloff:
go jenkins

PR #1501, opened 2025-05-05
Output process ID on application start

bschwarzent:
The help distinguish between multiple running applications, print the process id (PID) along with the server ready message.


paolobazzi (approved):
looks good!

PR #1504, opened 2025-05-06
Hide correlation ID for VetoExceptions thrown by REST requests

rogersteblerbsi:
The correlation ID is relevant to trace and debug unexpected errors. VetoExceptions usually are expected errors that should not need tracing.

419475


bschwarzent (approved):

paolobazzi (approved):
looks good!

PR #1508, opened 2025-05-08
Health Check: Remember last status per category

work-lto:
419786

### DIFF
--- a/org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/healthcheck/AbstractHealthChecker.java
+++ b/org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/healthcheck/AbstractHealthChecker.java
@@ -93,10 +93,9 @@ public abstract class AbstractHealthChecker implements IHealthChecker {
 
   @Override
   public boolean checkHealth(RunContext context, HealthCheckCategoryId category) {
+    m_lastStatusMap.putIfAbsent(category == null ? Empty.ID : category, new LastStatusEntry(m_timeToLive));
     LastStatusEntry lastStatusEntry = m_lastStatusMap.get(category == null ? Empty.ID : category);
-    if (lastStatusEntry == null) {
-      lastStatusEntry = new LastStatusEntry(m_timeToLive);
-    }
+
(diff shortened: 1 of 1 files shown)


avandorp:
[EDIT: Resolved with new commit] Thanks! Shouldn't we either have the timestamp per category in the map or invalidate the map on every put operation? Now one could query for category 1 and subsequently for a long time for category 2 and then get a stale answer for category 1.

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/healthcheck/AbstractHealthChecker.java
```diff
@@ -169,12 +174,21 @@ else if (m_timeout > 0 && m_futureStart + m_timeout < System.currentTimeMillis()
    * @param result
    *     status of last executed health check
    */
-  protected void notifyHealthCheckResult(boolean result) {
+  protected void notifyHealthCheckResult(boolean result, HealthCheckCategoryId category) {
```
paolobazzi: I wouldn't change the signature (API) of this method in a minor build. Move this signature change to 25/2

org.eclipse.scout.rt.server.commons/src/main/java/org/eclipse/scout/rt/server/commons/healthcheck/AbstractHealthChecker.java
```diff
@@ -169,12 +174,21 @@ else if (m_timeout > 0 && m_futureStart + m_timeout < System.currentTimeMillis()
...
-      LOG.debug("HealthCheck[{}] was successful", getName());
+      LOG.debug("HealthCheck[{}] was successful for category {}", getName(), category);
     }
     else {
-      LOG.warn("HealthCheck[{}] failed.", getName());
+      LOG.warn("HealthCheck[{}] failed for category {}.", getName(), category);
+    }
+  }
+
+  private static final class Empty implements IHealthCheckCategory {
```
paolobazzi, line 179: This class is not available as bean as it is marked "private static final". I think this is the intended behavior, maybe add a javadoc indicating that this is just a sentinel value and not a real category available for health checking.

PR #1507, opened 2025-05-08
Print stack trace for timed-out tests (of timed-out thread)

matthiaso:
Also: Create thread dump in case of
AssertNoRunningJobsStatement failure

PR #1510, opened 2025-05-09
Print stack trace for timed-out tests (of timed-out thread)

matthiaso:
Also: Create thread dump in case of AssertNoRunningJobsStatement failure


matthiaso:
See https://github.com/eclipse-scout/scout.rt/pull/1507

PR #1515, opened 2025-05-09
ajax: don't serialize null as 'null'

bschwarzent:
When calling dataObjects.serialize(), the argument must be checked for null. Otherwise, the data would be set to the string 'null' which would be visible the URL in the case of a GET request.

Improve jsdoc:
- Use consistent naming
- Use consistent parameter description
- Add empty line between description and parameters

Reorder methods:
- First all basic methods
- Followed by the "json" flavor of the same methods
- Followed by the "dataobject" flavor of the same methods


cguglielmo:
jenkins run test

cguglielmo (approved):

PR #1517, opened 2025-05-09
Mobile: don't make main box of detail forms scrollable

cguglielmo:
Normally, makeGroupBoxScrollable is called first
and then notifyPageDetailFormChanged because pageChanging is true while a page is being selected which delays the PROP_DETAIL_FORM changed event.

But if the page is already selected and the detail form changes dynamically, it is the other way round. In that case, makeGroupBoxScrollable must not revert the changes done by notifyPageDetailFormChanged.

413994


cguglielmo:
jenkins run test

fschinkel (approved):

PR #1535, opened 2025-05-21
Table: add insertColumns, deleteColumns, setColumns

fschinkel:
Adds the utility functions to insert, delete or set columns on a Table. For insertColumns and setColumns an initial value (value or callback) can be passed. The index of each column is handled automatically and the cells array of each row is updated accordingly.
After the modification a columnStructureChanged event is triggered containing the old and new columns.

378974

### DIFF
--- a/eclipse-scout-core/src/table/Table.ts
+++ b/eclipse-scout-core/src/table/Table.ts
@@ -423,28 +423,12 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
   protected _initColumns() {
     let cols = this.columns as ObjectOrChildModel<Column<any>>[];
     this.columns = cols.map((colModel, index) => {
-      let column: Column<any>;
-      let columnOrModel = colModel as FullModelOf<Column<any>>;
-      if (columnOrModel instanceof Column) {
-        column = columnOrModel;
-        column._setParent(this);
-      } else {
-        columnOrModel.parent = this;
-        column = scout.create(columnOrModel);
-      }
-
+      const column = this._ensureColumn(colModel);
       if (column.index < 0) {
         column.index = index;
       }
-      if (column.checkable) {
-        // set checkable column if this column is the checkable one
-        this.checkableColumn = column as BooleanColumn;
-      }
       return column;
     });
-
-    // Add gui only checkbox column at the beginning
-    this._setCheckable(this.checkable);
 
     // Add gui only row icon column at the beginning
     if (this.rowIconVisible) {
@@ -452,11 +436,27 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
     }
 
     this._setCompact(this.compact);
+
+    // update checkable column, table node column and sort columns
+    this._calculateCheckableTableNodeAndSortColumns();
+
+    this.columnLayoutDirty = true;
+  }
+
+  /**
+   * Calculates the {@link Table.checkableColumn} (see {@link Table._calculateCheckableColumn}), the {@link Table.tableNodeColumn} (see {@link Table._calculateTableNodeColumn})
+   * and the {@link Table._permanentHeadSortColumns} and {@link Table._permanentTailSortColumns} (see {@link Table._setHeadAndTailSortColumns}).
+   */
+  protected _calculateCheckableTableNodeAndSortColumns() {
+    // update checkable column
+    this._calculateCheckableColumn();
+    this._setCheckable(this.checkable);
+
+    // update table node column
     this._calculateTableNodeColumn();
 
-    // Sync head and tail sort columns
+    // update sort columns
     this._setHeadAndTailSortColumns();
-    this.columnLayoutDirty = true;
   }
 
   protected override _destroy() {
@@ -912,7 +912,7 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
   onColumnVisibilityChanged() {
     this.columnLayoutDirty = true;
     this._calculateTableNodeColumn();
-    this.trigger('columnStructureChanged');
+    this._triggerColumnStructureChanged(this.columns, this.columns);
 
     // Rebuild aggregate rows. This computes missing aggregate values for previously hidden columns. It is also a convenient
     // way to fix the column indices. The aggregate table control was already updated via 'columnStructureChanged' event.
@@ -4985,16 +4985,17 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
   protected _setRowIconVisible(rowIconVisible: boolean) {
     this._setProperty('rowIconVisible', rowIconVisible);
     let column = this.rowIconColumn;
+    const columns = [...this.columns];
     if (this.rowIconVisible && !column) {
       this._insertRowIconColumn();
       this._calculateTableNodeColumn();
-      this.trigger('columnStructureChanged');
+      this._triggerColumnStructureChanged(columns, this.columns);
     } else if (!this.rowIconVisible && column) {
       column.destroy();
       arrays.remove(this.columns, column);
       this.rowIconColumn = null;
       this._calculateTableNodeColumn();
-      this.trigger('columnStructureChanged');
+      this._triggerColumnStructureChanged(columns, this.columns);
     }
   }
 
@@ -5283,19 +5284,34 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
     this._updateCheckableColumn();
   }
 
+  /**
+   * Calculates the {@link Table.checkableColumn}. It is the last column in {@link Table.columns} that is {@link Column.checkable}.
+   */
+  protected _calculateCheckableColumn() {
+    const oldCheckableColumn = this.checkableColumn;
+    this.checkableColumn = [...this.columns].reverse().find(c => c.checkable) as BooleanColumn;
+
+    // remove old checkable column if it was a guiOnly column
+    if (this.checkableColumn !== oldCheckableColumn && oldCheckableColumn?.guiOnly) {
+      oldCheckableColumn.destroy();
+      arrays.remove(this.columns, oldCheckableColumn);
+    }
+  }
+
   protected _updateCheckableColumn() {
     let column = this.checkableColumn;
     let showCheckBoxes = this.checkable && scout.isOneOf(this.checkableStyle, Table.CheckableStyle.CHECKBOX, Table.CheckableStyle.CHECKBOX_TABLE_ROW);
+    const columns = [...this.columns];
     if (showCheckBoxes && !column) {
       this._insertBooleanColumn();
       this._calculateTableNodeColumn();
-      this.trigger('columnStructureChanged');
+      this._triggerColumnStructureChanged(columns, this.columns);
     } else if (!showCheckBoxes && column && column.guiOnly) {
       column.destroy();
       arrays.remove(this.columns, column);
       this.checkableColumn = null;
       this._calculateTableNodeColumn();
-      this.trigger('columnStructureChanged');
+      this._triggerColumnStructureChanged(columns, this.columns);
     }
   }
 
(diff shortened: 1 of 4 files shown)


eclipse-scout-core/src/table/Table.ts
```diff
@@ -5713,15 +5705,172 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
     return this.visibleColumns().some(column => column instanceof NumberColumn && column.aggregationFunction !== 'none');
   }
 
+  setColumns(columnOrModels: ObjectOrChildModel<Column<any>>[], initValue?: any | ((Column, TableRow) => any)) {
```
cguglielmo, line 5907: New API methods should have JS doc

eclipse-scout-core/src/table/Table.ts
```diff
@@ -5713,15 +5705,172 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+      initValue = (column: Column, row: TableRow) => value;
+    }
+
+    // ensure columns
+    const columns = columnOrModels
+      .map(column => this._ensureColumn(column))
+      .filter(column => !!column);
+
+    const columnsOld = [...this.columns];
+    const columnsDeleted = this.columns.filter(column => column instanceof Column && !arrays.contains(columns, column));
```
cguglielmo: You can use arrays.diff function.

cguglielmo: How can this.columns contain a column that is not an instance of Column?

eclipse-scout-core/src/table/Table.ts
```diff
@@ -5713,15 +5705,172 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+        // update index of all columns whose index is greater than the index of the removed column
+        for (const c of columnsCurrent) {
+          if (c.index > column.index) {
+            c.index--;
+          }
+        }
+
+        // remove cells for all rows at the correct position (column.index)
+        this.rows.forEach(row => row.cells.splice(column.index, 1));
+
```
cguglielmo: remove empty line

eclipse-scout-core/src/table/Table.ts
```diff
@@ -5713,15 +5705,172 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+      const value = initValue;
+      initValue = (column: Column, row: TableRow) => value;
+    }
+
+    // ensure columns
+    const columns = columnOrModels
+      .map(column => this._ensureColumn(column))
+      .filter(column => !!column);
+
+    const columnsOld = [...this.columns];
```
cguglielmo: I personally prefer the human readable writing -> oldColumns instead of columnsOld

eclipse-scout-core/src/table/Table.ts
```diff
@@ -5713,15 +5705,172 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+    // add columns last if no position is given
+    if (!objects.isNumber(position)) {
+      position = this.columns.length;
+    }
+
+    // remove columns from columnOrModels that are already part of this.columns
+    arrays.removeAll(columnOrModels, this.columns);
+
+    // create new columns
+    const columnsNew = [...this.columns];
```
cguglielmo: Consider human readable writing

eclipse-scout-core/src/table/Table.ts
```diff
@@ -5713,15 +5705,172 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+      // insert new cell for all rows at the correct position (column.index)
+      this.rows.forEach(row => {
+        arrays.insert(row.cells, column.initCell(initValue(column, row), row), column.index);
+      });
+
+      // add to current columns
+      columnsCurrent.push(column);
+    }
+
+    this._setProperty('columns', columns);
```
cguglielmo, line 6011: The logic of _initColumns needs to applied here as well.
- gui only columns have to be at the beginning, and it  must not be possible to remove them using setColumns, otherwise the state of checkable, rowIconVisible etc. will be inconsistent 
- permanent sort columns have to be updated
- node icon column needs to be updated
- checkable column needs to be updated
-> extract relevant code from initColumns to a separate method so it can be called here as well

cguglielmo (changes requested):

cguglielmo (commented):
Please mention the new functions in the release notes

eclipse-scout-core/src/table/Table.ts
cguglielmo: Also, there are other column properties, that have an effect on the rows and need to be re-applied, if the new columns have one of these properties set. Sorting, grouping, filters, background-effect, maybe more. Have a look at the Table._init-function, all of these properties are applied there, directly or indirectly. Maybe it is enough to call _sort() (which also groups), _calculateValuesForBackgroundEffect() and _updateRowStructure() with applyFilters only.

eclipse-scout-core/src/table/Table.ts
```diff
@@ -5858,15 +5851,170 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
     return this.visibleColumns().some(column => column instanceof NumberColumn && column.aggregationFunction !== 'none');
   }
 
+  setColumns(columnOrModels: ObjectOrChildModel<Column<any>>[], initValue?: any | ((Column, TableRow) => any)) {
```
cguglielmo, line 5907: Maybe don't use arrays.equals, just ===. That would give us the possibility to apply the new column settings like sorting, grouping etc. even if the columns themselves did not change. Maybe useful for the reset()-functionality.

eclipse-scout-core/src/table/Table.ts
```diff
@@ -5858,15 +5851,170 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+  }
+
+  protected _triggerColumnStructureChanged(columnsOld: Column<any>[], columnsNew: Column<any>[]) {
+    // create new arrays to prevent unexpected behaviour if listeners modify the arrays
+    columnsOld = [...columnsOld];
+    columnsNew = [...columnsNew];
+    this.trigger('columnStructureChanged', {columnsOld, columnsNew});
+  }
+
+  insertColumn(columnOrModel: ObjectOrChildModel<Column<any>>, position?: number, initValue?: any | ((Column, TableRow) => any)) {
```
bschwarzent: It would be convenient if we could also pass a reference column instead of the position. That way, the caller would not need to compute the index by itself.

```suggestion
  insertColumn(columnOrModel: ObjectOrChildModel<Column<any>>, positionOrInsertAfterColumn?: number | Column<any>, initValue?: any | ((Column, TableRow) => any)) {
```

eclipse-scout-core/src/table/Table.ts
fschinkel: `this.columns` contains only `ColumnModel` initially. Therefore, a call of `insertColumn` during the initialization of the `Table` will delete the cells of all other columns otherwise.

eclipse-scout-core/src/table/Table.ts
```diff
@@ -5884,15 +5903,239 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
     return this.visibleColumns().some(column => column instanceof NumberColumn && column.aggregationFunction !== 'none');
   }
 
+  /**
+   * Set the given {@link Column}s. All columns that are removed will also be removed from {@link TableRow.cells}. All new columns are initialized with the given initValue.
```
cguglielmo: Set -> Sets

eclipse-scout-core/src/table/Table.ts
```diff
@@ -5884,15 +5903,239 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+      return;
+    }
+    this._setColumns(columnOrModels, initValue);
+    if (this.rendered) {
+      this._renderColumns();
+    }
+  }
+
+  /**
+   * see {@link setColumns}.
```
cguglielmo: see -> @see setColumns

eclipse-scout-core/src/table/Table.ts
```diff
@@ -5884,15 +5903,239 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+   */
+  protected _triggerColumnStructureChanged(oldColumns: Column<any>[], newColumns: Column<any>[]) {
+    // create new arrays to prevent unexpected behaviour if listeners modify the arrays
+    oldColumns = [...oldColumns];
+    newColumns = [...newColumns];
+    this.trigger('columnStructureChanged', {oldColumns, newColumns});
+  }
+
+  /**
+   * Insert the given {@link Column}.
```
cguglielmo: Insert > Inserts

eclipse-scout-core/src/table/Table.ts
```diff
@@ -5884,15 +5903,239 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+   * The column will be inserted into {@link Table.columns} at the specified position or after the given anchor {@link Column}.
+   * The inserted column is initialized with the given initValue.
+   */
+  insertColumn(columnOrModel: ObjectOrChildModel<Column<any>>, positionOrInsertAfterColumn?: number | Column<any>, initValue?: any | ((Column, TableRow) => any)) {
+    this.insertColumns([columnOrModel], positionOrInsertAfterColumn, initValue);
+  }
+
+  /**
+   * Insert the given {@link Column}s.
```
cguglielmo: Insert > Inserts

eclipse-scout-core/src/table/Table.ts
```diff
@@ -5884,15 +5903,239 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+
+    // create new columns
+    const newColumns = [...this.columns];
+    arrays.insertAll(newColumns, columnOrModels, position);
+
+    this.setColumns(newColumns, initValue);
+  }
+
+  /**
+   * Delete the given {@link Column} from {@link Table.columns}.
```
cguglielmo: Delete -> Deletes

eclipse-scout-core/src/table/Table.ts
```diff
@@ -5884,15 +5903,239 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+   */
+  deleteColumn(column: Column) {
+    this.deleteColumns([column]);
+  }
+
+  /**
+   * Delete the given {@link Column}s from {@link Table.columns}.
```
cguglielmo: Delete -> Deletes

eclipse-scout-core/src/table/Table.ts
```diff
@@ -5283,19 +5284,37 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+  protected _calculateCheckableColumn() {
+    const oldCheckableColumn = this.checkableColumn;
+    this.checkableColumn = [...this.columns].reverse().find(c => c.checkable) as BooleanColumn;
+
+    // remove old checkable column if it was a guiOnly column
+    if (this.checkableColumn !== oldCheckableColumn && oldCheckableColumn?.guiOnly) {
+      const columns = [...this.columns];
+      oldCheckableColumn.destroy();
+      arrays.remove(this.columns, oldCheckableColumn);
+      this._calculateTableNodeColumn();
```
cguglielmo: calculateTableNodeColumn and triggerColumnStructureChanged are not necessary as this is already done by the caller.

eclipse-scout-core/src/table/Table.ts
```diff
@@ -5884,15 +5903,239 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
     return this.visibleColumns().some(column => column instanceof NumberColumn && column.aggregationFunction !== 'none');
   }
 
+  /**
+   * Set the given {@link Column}s. All columns that are removed will also be removed from {@link TableRow.cells}. All new columns are initialized with the given initValue.
+   * Column properties like e.g. {@link Column.sortIndex}, {@link Column.grouped} and {@link NumberColumn.backgroundEffect} and {@link Table.filters} are re-applied.
```
cguglielmo: replace the first and with a comma

PR #1536, opened 2025-05-21
Improve AutoLeafPageWithNodes

fschinkel:
The AutoLeafPageWithNodes must not use the text of the first cell of the corresponding row as this cell holds e.g. a primary key and is not displayable. Therefore, use the result of computeTextForRow as this combines the values of all summary columns.
If the Table is compact computeTextForRow now returns the compactValue of the row.

378974

### DIFF
--- a/eclipse-scout-core/src/desktop/outline/pages/Page.ts
+++ b/eclipse-scout-core/src/desktop/outline/pages/Page.ts
@@ -517,15 +517,23 @@ export class Page extends TreeNode implements PageModel, ObjectWithUuid {
       page.htmlEnabled = row.cells[0].htmlEnabled;
       page.cssClass = row.cells[0].cssClass;
     }
+    if (row.getTable().compact) {
+      page.htmlEnabled = true;
+    }
     return page;
   }
 
   /**
-   * This function creates the text property of this page. The default implementation returns the texts of the summary columns of the table or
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/src/desktop/outline/pages/Page.ts
```diff
@@ -517,15 +517,23 @@ export class Page extends TreeNode implements PageModel, ObjectWithUuid {
...
-   * This function creates the text property of this page. The default implementation returns the texts of the summary columns of the table or
-   * from the first cell of the given row. It's allowed to ignore the given row entirely, when you override this function.
+   * This function creates the text property of this page. The default implementation returns the compactValue of the row if the table is compact,
+   * the texts of the summary columns of the table or from the first cell of the given row. It's allowed to ignore the given row entirely, when
+   * you override this function.
    */
   computeTextForRow(row: TableRow): string {
-    const summaryColumns = row.getTable().summaryColumns();
+    const table = row.getTable();
+    if (table.compact) {
```
cguglielmo, line 533: Why is this necessary? I don't think we should to that. It looks bad if the node is selected (content is not readable) and bloats the tree.
Could you implement the same logic as in Java (SummaryCellBuilder)? MobileSummaryCellBuilder can be ignored for now. Most of the logic is already implemented in SamplePageWithNodes.computeTextForRow. Please remove this code afterwards.

fschinkel:
replaced by #1565

PR #1539, opened 2025-05-23
Add MaintenanceMode for REST resources (feature/filter)

matthiaso:
321464

### DIFF
--- a/org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/maintenance/IMaintenanceModeService.java
+++ b/org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/maintenance/IMaintenanceModeService.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.scout.rt.rest.maintenance;
+
+import org.eclipse.scout.rt.platform.ApplicationScoped;
+
+/**
+ * <p>Interface for an application scoped bean to determine whether maintenance mode is currently active/current session may override it.</p>
+ * <p>An implementation must be provided if {@link MaintenanceMode} annotation is used.</p>
+ *
+ * @see MaintenanceMode
+ */
+@ApplicationScoped
+public interface IMaintenanceModeService {
+
+  /**
+   * @return true if maintenance mode is active
+   */
+  boolean isActive();
+
+  /**
+   * @return true if specific request/session may override maintenance mode
+   */
+  boolean isGrantOverride();
+}
--- a/org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/maintenance/MaintenanceMode.java
+++ b/org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/maintenance/MaintenanceMode.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.scout.rt.rest.maintenance;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.eclipse.scout.rt.rest.IRestResource;
+
+/**
+ * <p>Annotate a {@link IRestResource} to disable access if application is in maintenance mode ({@link MaintenanceModeFilter} is installed for annotated types/methods).</p>
+ * <p>If annotation is used an implementation of {@link IMaintenanceModeService} must be provided (could be as simple as a bean holding a boolean attribute to enable/disable maintenance mode)</p>
+ * <p>Annotation may be used on types and their methods, if it is provided on both of them the method's annotation is the relevant one (allowing to disable a resource type in general for maintenance mode and enable a specific method on the type).</p>
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD, ElementType.TYPE})
+public @interface MaintenanceMode {
+
+  boolean disabled() default true;
+}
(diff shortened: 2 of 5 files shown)


org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/maintenance/IMaintenanceModeService.java
```diff
@@ -0,0 +1,26 @@
...
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.scout.rt.rest.maintenance;
+
+import org.eclipse.scout.rt.platform.ApplicationScoped;
+
+@ApplicationScoped
+public interface IMaintenanceModeService {
```
paolobazzi, line 21: Add short javadoc indicating purpose of this interface
And maybe add default impl. in Scout using @Order(4000) and offering getter/setter?

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/maintenance/IMaintenanceModeService.java
```diff
@@ -0,0 +1,26 @@
...
+
+  /**
+   * @return true if maintenance mode is active
+   */
+  boolean isActive();
+
+  /**
+   * @return true if maintenance mode is active but specific request/session may override maintenance mode
```
paolobazzi: is this javadoc correct? I'll suggest just to return true if `specific request/session may override maintenance mode` without considerin isActive state of maintenance mode here

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/maintenance/MaintenanceModeFeature.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.eclipse.scout.rt.platform.BEANS;
+import org.eclipse.scout.rt.platform.util.Assertions;
+import org.eclipse.scout.rt.platform.util.ObjectUtility;
+import org.eclipse.scout.rt.rest.RestApplication.IRestApplicationClassesContributor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * <p>Installs the {@link MaintenanceModeFilter} for REST resource types and their methods which may be annotated with {@link MaintenanceMode} tp check for maintenance mode on access.</p>
```
paolobazzi: Suggestion:
```
 * <p>Installs the {@link MaintenanceModeFilter} for REST resource types and their methods which are annotated with {@link MaintenanceMode} to check for maintenance mode on access.</p>
 ```

File: org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/maintenance/MaintenanceMode.java [hunk 4]
```diff
@@ -0,0 +1,28 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.eclipse.scout.rt.rest.IRestResource;
+
+/**
+ * <p>Annotate a {@link IRestResource} to disable access if application is in maintenance mode ({@link MaintenanceModeFilter} is installed for annotated types/methods).</p>
+ * <p>If annotation is used an implementation of {@link IMaintenanceModeService} must be provided (could be as simple as a bean holding a boolean attribute to enable/disable maintenance mode)</p>>
```
paolobazzi: Add note about precedence: annotation on method over annotation on resource class

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/maintenance/MaintenanceModeFilter.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+@Priority(Priorities.AUTHORIZATION)
+public class MaintenanceModeFilter implements ContainerRequestFilter {
+
+  private static final Logger LOG = LoggerFactory.getLogger(MaintenanceModeFilter.class);
+
+  protected static final String DEFAULT_MAINTENANCE_MESSAGE = "Resources are temporarily unavailable due to maintenance.";
+
+  @Context
+  protected ResourceInfo m_resourceInfo;
```
paolobazzi: resource info is unused here

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/maintenance/MaintenanceMode.java
paolobazzi: Remove last `>` of javadoc

paolobazzi (commented):
Looks good, see minor inputs.

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/maintenance/MaintenanceMode.java
matthiaso: Thanks/done.

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/maintenance/MaintenanceModeFilter.java
matthiaso: Removed the member attribute, thanks for the hint.

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/maintenance/MaintenanceMode.java
matthiaso: Added another paragraph about it.

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/maintenance/MaintenanceModeFeature.java
matthiaso: Thanks for the suggestion/applied it.

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/maintenance/IMaintenanceModeService.java
matthiaso: That's right; applied the suggestion/thanks

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/maintenance/IMaintenanceModeService.java
matthiaso: Added a javadoc for the interface.

Do we require the default impl for the bean? Actually I did not create it on purpose; w/o using the annoation no impl would be necessary; if the annotation is used the feature would throw an error if no impl is available for the bean. On the other side: If we usually create a default impl I could of course also create one here.

PR #1542, opened 2025-05-26
Features/cgu/25.2/table improve api

cguglielmo:


eclipse-scout-core/src/form/fields/smartfield/TableProposalChooser.ts
```diff
@@ -8,22 +8,24 @@
  * SPDX-License-Identifier: EPL-2.0
  */
 import {
-  aria, arrays, Cell, Column, ColumnDescriptor, ListBoxTableAccessibilityRenderer, lookupField, LookupRow, objects, ProposalChooser, scout, SmartFieldLookupResult, Table, TableLayoutResetter, TableRow, TableRowClickEvent, TableRowModel,
+  aria, arrays, Cell, Column, ColumnDescriptor, ColumnModel, ListBoxTableAccessibilityRenderer, lookupField, LookupRow, ObjectOrChildModel, objects, ProposalChooser, scout, SmartFieldLookupResult, Table, TableLayoutResetter, TableRow,
+  TableRowClickEvent,
```
bschwarzent: Format code

eclipse-scout-core/src/table/MoveTableRowMenuHelper.ts
```diff
@@ -0,0 +1,79 @@
...
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {Action, arrays, Event, EventEmitter, EventHandler, Menu, scout, Table, TableRow, TableRowMoveDirection} from '../index';
+
+/**
+ * When installed, the helper implements the actions for the given row menus to move the selected rows up or down
```
bschwarzent:
```suggestion
 * When installed, the helper implements the actions for the given row menus to move the selected rows up or down.
```

eclipse-scout-core/src/table/MoveTableRowMenuHelper.ts
```diff
@@ -0,0 +1,79 @@
...
+      // Already installed
+      return;
+    }
+    this.table = scout.assertInstance(options.table, Table);
+    this.moveRowUpMenu = scout.assertInstance(options.moveRowUpMenu, Menu);
+    this.moveRowDownMenu = scout.assertInstance(options.moveRowDownMenu, Menu);
+    this.selectedRowsFilter = options.selectedRowsFilter;
+
+    this._actionHandler = this._onMoveRowMenuAction.bind(this);
+    this._tableHandler = event => this._updateMenus();
```
bschwarzent: It would be more consistent if either both handles are implemented here or implemented as instance methods. Since the class might (theoretically) be extended, I would opt for the instance method to give them access to the event object.

eclipse-scout-core/src/table/MoveTableRowMenuHelper.ts
```diff
@@ -0,0 +1,79 @@
...
+    this.moveRowDownMenu.off('action', this._actionHandler);
+    this.table.off('rowsSelected rowsInserted rowsUpdated rowsDeleted rowOrderChanged', this._tableHandler);
+  }
+
+  protected _updateMenus() {
+    let moveable = this.table.selectedRows.length && this.table.rows.length > 1;
+    let firstRow = arrays.first(this.table.visibleRows);
+    let lastRow = arrays.last(this.table.visibleRows);
+    // Using a new dimension makes it possible to still hide and disable the menus by the consumer of this helper, even if this helper would show or enable the menus
+    this.moveRowUpMenu.setPropertyDimension('visible', 'boundary', moveable);
```
bschwarzent: Why is this dimension called 'boundary'?

eclipse-scout-core/src/table/Table.ts
```diff
@@ -1884,125 +1881,165 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
   /**
    * Moves the row to the top.
    */
-  moveRowToTop(row: TableRow) {
+  moveRowToTop(row: TableRow): boolean {
```
bschwarzent, line 1886: Describe return value in jsdoc (same for `moveRowToBottom`)

eclipse-scout-core/src/table/Table.ts
```diff
@@ -1884,125 +1881,165 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
     let rowIndex = this.rows.indexOf(row);
-    this.moveRow(rowIndex, this.rows.length - 1);
+    if (rowIndex < 0) {
+      return false;
+    }
+    return this.moveRow(rowIndex, this.rows.length - 1);
   }
 
   /**
    * Moves the row one up, disregarding filtered rows.
```
bschwarzent: What does "disregarding filtered rows" mean? How is this different from moving only "visible" rows?

eclipse-scout-core/src/table/Table.ts
```diff
@@ -2928,10 +3005,43 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+   * table.insertRows(rows);
+   *
+   * @example Type-safe approach to insert rows using {@link Column}.
+   * // Create an array of actual table rows. The values or cell properties can be set using the appropriate column.
+   * // This approach is more type-safe and does not rely on the order of the columns.
+   * let row0 = scout.create(TableRow, {parent: table});
+   * table.columnById('FirstColumn').setCellValue(row0, 'value row 0 column 0');
+   * table.columnById('SecondColumn').setCellValue(row0, 'value row 0 column 1');
+   *
+   * let row1 = scout.create(TableRow, {parent: table});
```
bschwarzent, line 3041: Idea: Maybe we could make this more DSL-like by providing a static factory method to create table rows with an initializer function.

```
table.insertRows([
  TableRow.create(table, row => {
    table.columnById('FirstColumn').setCellValue(row, 'first value');
    ...
  }),
  TableRow.create(table, row => {
    table.columnById('FirstColumn').setCellValue(row, 'first value in second row');
    ...
  })
]);
```

eclipse-scout-core/src/table/Table.ts
```diff
@@ -3814,6 +3924,13 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
     return null;
   }
 
+  /**
+   * @returns all selected rows in the same order as {@link rows}.
+   */
+  get selectedRowsSorted(): TableRow[] {
```
bschwarzent: Can we make the use of getter and normal methods consistent? E.g. this is a getter, but `checkedRows()` is not.

eclipse-scout-core/src/table/Table.ts
```diff
@@ -4059,6 +4179,26 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
     this._renderEmptyData();
   }
 
+  /**
+   * Sorts the given rows _in place_ according the order of {@link rows}.
+   * Does nothing, if rows is null or undefined.
+   *
+   * @returns a reference to the given rows which are now in order.
+   */
+  orderRows(rows: TableRow[], direction: 'asc' | 'desc' = 'asc'): TableRow[] {
```
bschwarzent: The name of this method is a bit odd. It sounds like ordering a pizza. ;-)

`sortRows()` or `applyRowOrder`? 

Who else calls this method? Does it need to be public?

eclipse-scout-core/src/table/columns/Column.ts
```diff
@@ -202,16 +206,23 @@ export class Column<TValue = string> extends PropertyEventEmitter implements Col
...
-    this.table = table;
-    if (this.table) {
-      this.table.on('columnMoved columnStructureChanged', this._tableColumnsChangedHandler);
+    this.parent = parent;
+    if (this.parent) {
+      this.parent.on('columnMoved columnStructureChanged', this._tableColumnsChangedHandler);
     }
   }
 
+  get table() {
```
bschwarzent: Since this is (technically) a method, should we declare the return type?

eclipse-scout-core/src/table/MoveTableRowMenuHelper.ts
cguglielmo: it can be any name, boundary refers to the boundary of the table, e.g. a row cannot be moved further up if it is the first row. I changed it to helper, maybe this is less confusing

eclipse-scout-core/src/table/Table.ts
cguglielmo: you are right, I would prefer having getters for these kind of methods like selectedRow(), filteredRows(), filterCount() etc., but since these methods cannot be deprecated, migration would be mandatory -> I removed get from selectedRowsSorted.

PR #1543, opened 2025-05-26
Table: improve moveColumn

cguglielmo:
- moveColumn now does nothing if column is already at the correct position or is not allowed to be moved to a new position. This ensures header item animation does not break when calling moveColumn for such columns.
- Removed oldPos and dragging parameter to simplify api
- Moving back to original position if user tries to move passed a fixed position column now works the same way as releasing without changing to a new position.

### DIFF
--- a/eclipse-scout-core/src/table/TableHeader.ts
+++ b/eclipse-scout-core/src/table/TableHeader.ts
@@ -258,24 +258,23 @@ export class TableHeader extends Widget implements TableHeaderModel {
     let that = this;
     $headers.each(function() {
       // move to old position and then animate
-      $(this).css('left', $(this).data('old-pos') - $(this).offset().left)
-        .animate({
-          left: 0
-        }, {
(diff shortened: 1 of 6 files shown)


eclipse-scout-core/src/table/TableHeader.ts
```diff
@@ -773,8 +767,13 @@ export class TableHeader extends Widget implements TableHeaderModel {
       });
 
       // move column
-      if (newPos > -1 && oldPos !== newPos) {
-        that.table.moveColumn($header.data('column'), oldPos, newPos, true);
+      const oldLeft = $header.offset().left;
+      const moved = that.table.moveColumn($header.data('column'), newPos);
+      if (moved) {
+        const left = parseInt($header.css('left'), 0) + oldLeft - $header.offset().left;
```
bschwarzent:
```suggestion
        const left = $header.cssLeft() + oldLeft - $header.offset().left;
```

PR #1547, opened 2025-05-28
IdCodec: Improve extensibility of unqualified composite id parts

paolobazzi:
Allow to intercept encoding of unqualified composite id parts.

417577


paolobazzi:
Review see https://github.com/eclipse-scout/scout.rt/pull/1545

PR #1552, opened 2025-06-05
Tree: send correct events if root node changes

cguglielmo:
If root node is visible and changes dynamically, the ui needs to be informed about the new state. Currently, the UI server sends selection and page changed events referring the old root node which creates exceptions in the browser because that root node cannot be found.

422641

### DIFF
--- a/eclipse-scout-core/src/desktop/outline/overview/TileOutlineOverview.ts
+++ b/eclipse-scout-core/src/desktop/outline/overview/TileOutlineOverview.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -7,7 +7,9 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-import {aria, Desktop, EventHandler, HtmlComponent, InitModelOf, OutlineOverview, Page, PageTileGrid, PropertyChangeEvent, RowLayout, scout, TileOutlineOverviewEventMap, TileOutlineOverviewModel} from '../../../index';
+import {
+  aria, Desktop, EventHandler, HtmlComponent, InitModelOf, OutlineOverview, Page, PageTileGrid, PropertyChangeEvent, RowLayout, scout, TileOutlineOverviewEventMap, TileOutlineOverviewModel, TreeNodesDeletedEvent, TreeNodesInsertedEvent
+} from '../../../index';
 
 export class TileOutlineOverview extends OutlineOverview implements TileOutlineOverviewModel {
   declare model: TileOutlineOverviewModel;
@@ -20,6 +22,8 @@ export class TileOutlineOverview extends OutlineOverview implements TileOutlineO
   contentHtmlComp: HtmlComponent;
   $title: JQuery;
   protected _desktopNavigationVisibilityChangeHandler: EventHandler<PropertyChangeEvent<boolean, Desktop>>;
+  protected _outlineNodesInsertedHandler: EventHandler<TreeNodesInsertedEvent>;
+  protected _outlineNodesDeletedHandler: EventHandler<TreeNodesDeletedEvent>;
 
   constructor() {
     super();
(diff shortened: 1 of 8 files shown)


cguglielmo:
jenkins run test

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/tree/JsonTree.java
```diff
@@ -1101,8 +1118,8 @@ protected void handleUiPropertyChange(String propertyName, JSONObject data) {
   /**
    * Ignores invisible root node and deleted and filtered nodes because for the UI, they don't exist.
```
bschwarzent: Adjust documentation to reflect new argument `checkInvisibleRootNode`.
And maybe copy the original javadoc to the new method `isNodeAccepted(ITreeNode node)`.

org.eclipse.scout.rt.ui.html.test/src/test/java/org/eclipse/scout/rt/ui/html/json/tree/JsonTreeTest.java
```diff
@@ -739,10 +739,10 @@ public void testWrongEventOrder() throws JSONException {
     assertEquals("nodesInserted", events.get(0).getType());
```
bschwarzent, line 659: I noticed a commented-out test case "testNodeAndUserFilter" (ca. lines 490-530), maybe we could use this change to clean that up.

org.eclipse.scout.rt.ui.html.test/src/test/java/org/eclipse/scout/rt/ui/html/json/desktop/JsonOutlineTest.java
```diff
@@ -333,6 +335,34 @@ public void testNoDetailTableSentForInvisibleRootNode() {
...
 
+  @Test
+  public void testChangingVisibleRootNodeForMobile() {
+    IOutline outline = new OutlineWithOneNode();
+    outline.setRootNodeVisible(true);
+    BEANS.get(MobileDeviceTransformer.class).transformOutline(outline);
+
+    IPage<?> oldRootNode = outline.getRootPage();
+    JsonOutline<IOutline> jsonOutline = UiSessionTestUtility.newJsonAdapter(m_uiSession, outline);
+    String oldRootNodeId = jsonOutline.optNodeId(oldRootNode);
```
bschwarzent: Is the use of optNodeId deliberate? I think, getNodeId() does (and should) work here.

eclipse-scout-core/src/desktop/outline/overview/TileOutlineOverview.ts
```diff
@@ -28,17 +32,29 @@ export class TileOutlineOverview extends OutlineOverview implements TileOutlineO
...
+    this._outlineNodesDeletedHandler = this._onOutlineNodesDeleted.bind(this);
   }
 
   protected override _init(model: InitModelOf<this>) {
     super._init(model);
-    if (!this.pageTileGrid) {
-      this.pageTileGrid = this._createPageTileGrid();
+    this.pageTileGrid = this._createPageTileGrid();
+    if (this.outline.compact) {
+      this.outline.on('nodesDeleted', this._outlineNodesDeletedHandler.bind(this));
```
bschwarzent: Remove `.bind(this)`

The handlers are already bound functions. Calling bind() again, does not change "this" but still creates a new function instance. The listeners are therefore never removed again.

eclipse-scout-core/src/desktop/outline/overview/TileOutlineOverview.ts
```diff
@@ -113,4 +133,22 @@ export class TileOutlineOverview extends OutlineOverview implements TileOutlineO
...
+      // If there is already a page tile grid, it will handle the insertion itself
+      return;
+    }
+    this.setPageTileGrid(this._createPageTileGrid());
+  }
+
+  protected _onOutlineNodesDeleted(event: TreeNodesDeletedEvent) {
+    // If outline is compact, the root page is passed to the page tile grid.
+    // If the root page is deleted, the grid needs to be recreated
+    // In outline is not compact, the grid itself takes care of page removals / insertions
```
bschwarzent:
```suggestion
    // If outline is not compact, the grid itself takes care of page removals / insertions
```

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/tree/JsonTree.java
cguglielmo: I removed the param, it is not really necessary because when the method is used recursively for child pages, the invisible root node will never be part of the node list because the nodes are initially filtered using getChildrenIfInvisibleRootNode

eclipse-scout-core/src/desktop/outline/overview/TileOutlineOverview.ts
cguglielmo: oops 🙈

PR #1553, opened 2025-06-06
Migrate Maven publishing from OSSRH to Central Portal

cguglielmo:
OSSRH service will be shut down on June 30th, 2025.

416713


cguglielmo:
jenkins run test

PR #1559, opened 2025-06-16
Features/cgu/25.2/table organizer form 389574

cguglielmo:


eclipse-scout-core/src/menu/menus.ts
```diff
@@ -203,5 +203,32 @@ export const menus = {
...
+
+  /**
+   * Removes the given menus from the existing menus of the menu owner and calls {@link MenuOwner.setMenus} to set the new list of menus.
+   */
+  deleteMenus(menuOwner: MenuOwner, menusToDelete: Menu[]) {
+    menusToDelete = arrays.ensure(menusToDelete);
+    if (menusToDelete.length === 0) {
+      return;
+    }
+    let menus = menuOwner.menus.slice();
```
fschinkel: you can use `arrays.diff`

eclipse-scout-core/src/menu/menus.ts
```diff
@@ -203,5 +203,32 @@ export const menus = {
...
+   * Appends the given menus to the existing menus of the menu owner and calls {@link MenuOwner.setMenus} to set the new list of menus.
+   */
+  insertMenus(menuOwner: MenuOwner, menusToInsert: ObjectOrChildModel<Menu>[]) {
+    menusToInsert = arrays.ensure(menusToInsert);
+    if (menusToInsert.length === 0) {
+      return;
+    }
+    let menus = menuOwner.menus as ObjectOrChildModel<Menu>[];
+    menuOwner.setMenus(menus.concat(menusToInsert));
+  },
```
fschinkel, line 218: do we need to check if menus are already part of the `MenuOwner`

eclipse-scout-core/src/table/organizer/TableOrganizer.ts
```diff
@@ -121,28 +114,50 @@ export class TableOrganizer implements ObjectWithType {
...
-    if (column.grouped) {
-      this.table.groupColumn(column, true, null, true); // multiGroup=true is necessary to not affect potentially other grouped columns
+  /**
+   * Moves the columns to their new position according to the indices in the given `visibleColumns` array.
+   */
+  moveColumns(visibleColumns: Column<any>[]) {
+    // Add guiOnly columns to visibleColumns array if they are not already included
+    // moveColumn only works with indices based on all visibleColumns including guiOnly columns
+    let guiOnlyColumns = this.table.columns.filter(column => column.guiOnly);
+    for (const column of guiOnlyColumns.reverse()) {
```
fschinkel, line 135: `Table.filterColumns`?

eclipse-scout-core/src/table/organizer/TableOrganizerForm.ts
```diff
@@ -0,0 +1,199 @@
...
+ */
+import {Action, arrays, Column, Event, Form, InitModelOf, MoveTableRowMenuHelper, scout, ShowInvisibleColumnsForm, Table, TableCompleteCellEditEvent, TableRow, TableRowModel, TableRowsSelectedEvent, WidgetModel} from '../../index';
+import TableOrganizerFormModel, {ColumnsTable0, ProfilesTable, TableOrganizerFormWidgetMap} from './TableOrganizerFormModel';
+
+export class TableOrganizerForm extends Form {
+  declare widgetMap: TableOrganizerFormWidgetMap;
+  table: Table;
+  profilesTable: ProfilesTable;
+  columnsTable: ColumnsTable0;
+  keyColumn: Column<Column>;
```
fschinkel, line 21: I think we do not need to export the whole ShowInvisibleColumnsFormModel that leads to this strange name. Can you fix clean up that file (move interfaces to main widget and remove json model from index.ts)?

eclipse-scout-core/src/table/organizer/TableOrganizerForm.ts
```diff
@@ -0,0 +1,199 @@
...
+    this.profilesTable.columnById('DefaultConfigColumn').setCellValue(row, false);
+    this.profilesTable.insertRow(row);
+    this._renameConfig(row);
+  }
+
+  protected _newConfigName(): string {
+    let profileNo = 1;
+    const baseName = this.session.text('New');
+    this.profilesTable.rows.map(row => this.profilesTable.columnById('ConfigNameColumn').cellValue(row));
+    const profiles = this.profilesTable.columnById('ConfigNameColumn').cellValues();
```
fschinkel, line 94: remove

eclipse-scout-core/src/table/organizer/TableOrganizerForm.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  protected _updateColumnMenus() {
+    let selectedColumns = this.keyColumn.selectedCellValues();
+    let columnAddable = this.table.organizer.isColumnAddable(arrays.last(selectedColumns));
+    let columnRemovable = false;
+    let columnModifiable = false;
+    let columnMovableToLeft = false;
+    let columnMovableToRight = false;
+    for (const column of selectedColumns) {
+      if (this.table.organizer.isColumnModifiable(column)) {
```
fschinkel, line 133: This does not seem correct, if one column is movable, ... then the action is enabled but only half of the expected action is executed.

eclipse-scout-core/src/table/organizer/TableOrganizerMenu.ts
```diff
@@ -0,0 +1,43 @@
...
+    this.on('propertyChange:selected', event => this._updateForm());
+  }
+
+  protected _updateForm() {
+    if (!this.selected) {
+      return;
+    }
+    if (!this.form) {
+      this.setForm(scout.create(TableOrganizerForm, {
+        parent: this
```
fschinkel, line 35: move to model

eclipse-scout-core/test/table/organizer/TableOrganizerFormSpec.ts
```diff
@@ -0,0 +1,218 @@
...
+      await form.load();
+      expect(columnsTable.rows.length).toBe(0);
+      expect(addColumnMenu.visible).toBe(true);
+    });
+  });
+
+  describe('remove menu', () => {
+
+    it('is visible if a selected row can be made invisible', async () => {
+      let table = createTable(2);
```
fschinkel, line 195: ... selected column ...

eclipse-scout-core/src/table/organizer/TableOrganizer.ts
```diff
@@ -121,28 +114,50 @@ export class TableOrganizer implements ObjectWithType {
   }
 
   /**
-   * Hides the given column. If it was grouped, the grouping is removed. If it was part of a filter, the filter is removed.
+   * Hides the given columns. If it was grouped, the grouping is removed. If it was part of a filter, the filter is removed.
    */
-  hideColumn(column: Column) {
+  hideColumns(columns: Column<any>[]) {
     if (!this.table) {
```
fschinkel, line 109: add API change to migration guide

eclipse-scout-core/src/table/TableHeaderMenu.ts
```diff
@@ -365,24 +371,15 @@ export class TableHeaderMenu extends Popup implements TableHeaderMenuModel {
     });
 
     this.moveGroup.render(this.$columnActions);
-    // link buttons with the group header, the header is updated with the text of the action
-    aria.linkElementWithLabel(this.toBeginButton.$container, this.moveGroup.$text);
```
fschinkel, line 369: how do these aria labels work now?

fschinkel (changes requested):

eclipse-scout-core/src/menu/menus.ts
cguglielmo: The menu owner already does it, if it is required. I don't think we should do it here

eclipse-scout-core/src/table/organizer/TableOrganizerForm.ts
cguglielmo: You are right, this is the current pattern. But unfortunately if we did this, it would make extending from that form more complicated because the widget map is not exported but needed for the widget map of the customized form. We need to discuss how to solve this

eclipse-scout-core/src/table/organizer/TableOrganizerForm.ts
cguglielmo: what do you mean?

eclipse-scout-core/src/table/organizer/TableOrganizerForm.ts
cguglielmo: It was intended to make it work the same way as RemoveColumn does: move only those columns that are movable. But after revisiting it I found some issues that are hard to fix. And the move helper actually wants to keep the distances between the columns when multiple columns are moved at once, this would not be guaranteed anymore. So, I simplified it.

eclipse-scout-core/src/table/organizer/TableOrganizerMenu.ts
cguglielmo: I want to create it only when the menu is selected. If it is specified in the model it, it will be created together with the menu, see also 364072. I added a comment.

eclipse-scout-core/src/table/TableHeaderMenu.ts
cguglielmo: TableHeaderMenuGroup.ts now sets the aria attributes

PR #1560, opened 2025-06-17
Migrate Maven publishing from OSSRH to Central Portal

cguglielmo:
OSSRH service will be shut down on June 30th, 2025.

The name in pom.xml is required, otherwise the publishing will fail.

416713


cguglielmo:
jenkins run test

cguglielmo:
jenkins run test

PR #1565, opened 2025-06-18
Mobile: several improvements

fschinkel:
Mobile: several improvements

AutoLeafPageWithNodes
Extend from PageWithNodes, but remove detailTable and detailForm created by super class. Use computeTextForRow to compute the text for this page instead of the text of the first cell as computeTextForRow correctly handles e.g. compact rows or summary columns.

Page
Improve updatePageFromTable and computeTextForRow. First determine the summary columns in the following order
- the compact column if the outline is compact
- the summary columns of the table
- the first visible column
With these column, build the summary text and use the first of these columns to determine the properties htmlEnabled, cssClass, iconId of the current page.

SearchFormTableControl
Add a table control designed for search forms. It has a default icon and keyStroke and enables/disables itself whether a form is set or not. It re-triggers the 'search' and 'reset' events of its form.

PageWithTable
Use the SearchFormTableControl to build the searchFilter and listen for 'search' and 'reset' of the SearchFormTableControl to reload the table. If the outline is compact, set alwaysCreateChildPage to true and leaf to false. In addition, hide all table controls that are not the SearchFormTableControl and close the SearchFormTableControl after a search is executed.

ParentTablePageMenuContributor
The ParentTablePageMenuContributor does not need to pass menus from a parent page to its children if the outline is compact. This is done by the Outline.

JsPage
A JsPage must not always send its text property to the UI. This was done because the text might come from a summary column from the parent page. The only case the text is from a summary of the parent page is when the parent page is a IPageWithTable. In this case send not just the text but also htmlEnabled, cssClass and iconId as these properties are also set by the SummaryCellBuilder.

TableRowDetail
Improve column selection used to display row detail. Earlier only visible columns where used to build the row detail. But as the TableRowDetail needs to display the same values as the TableCompactHandler use the columns of the compact handler if set on the table. As a fallback use visible columns including the compacted columns because otherwise no column will be used if the table is compacted.

422641

### DIFF
--- a/eclipse-scout-core/src/desktop/outline/pages/Page.ts
+++ b/eclipse-scout-core/src/desktop/outline/pages/Page.ts
@@ -8,9 +8,9 @@
  * SPDX-License-Identifier: EPL-2.0
  */
 import {
-  arrays, BaseDoEntity, BookmarkSupport, BookmarkTableRowIdentifierDo, ButtonTile, ChildModelOf, Constructor, dataObjects, DoTypeResolver, EnumObject, Event, EventHandler, EventListener, EventMapOf, EventModel, EventSupport, Form,
-  HtmlComponent, InitModelOf, inspector, Menu, MenuBar, MenuOwner, ObjectIdProvider, ObjectOrType, ObjectWithUuid, Outline, PageDetailMenuContributor, PageEventMap, PageModel, ParentTablePageMenuContributor, PropertyChangeEvent,
-  RequiredUnlessNotSubclass, scout, SomeRequired, strings, Table, TableRow, TableRowClickEvent, TileOutlineOverview, TileOverviewForm, TreeNode, typeName, UuidPathOptions, Widget
+  arrays, BaseDoEntity, BookmarkSupport, BookmarkTableRowIdentifierDo, ButtonTile, ChildModelOf, Column, comparators, Constructor, dataObjects, DoTypeResolver, EnumObject, Event, EventHandler, EventListener, EventMapOf, EventModel,
+  EventSupport, Form, HtmlComponent, InitModelOf, inspector, Menu, MenuBar, MenuOwner, ObjectIdProvider, ObjectOrType, ObjectWithUuid, Outline, PageDetailMenuContributor, PageEventMap, PageModel, ParentTablePageMenuContributor,
+  PropertyChangeEvent, RequiredUnlessNotSubclass, scout, SomeRequired, strings, Table, TableRow, TableRowClickEvent, TileOutlineOverview, TileOverviewForm, TreeNode, typeName, UuidPathOptions, Widget
 } from '../../../index';
 import $ from 'jquery';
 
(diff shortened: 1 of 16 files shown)


eclipse-scout-core/src/desktop/outline/pages/Page.ts
```diff
@@ -512,27 +512,61 @@ export class Page extends TreeNode implements PageModel, ObjectWithUuid {
...
+    if (!summaryColumns) {
+      summaryColumns = this._computeSummaryColumns(row);
     }
-    if (row.cells.length >= 1) {
-      return row.cells[0].text;
+    return strings.join(' ', ...summaryColumns.map(summaryColumn => summaryColumn.cellText(row)));
+  }
+
+  /**
+   * Compute the summary columns of the given {@link TableRow}. The summary columns are
```
cguglielmo:
```suggestion
   * Computes the summary columns of the given {@link TableRow}. The summary columns are
```

eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
```diff
@@ -59,6 +97,12 @@ export class PageWithTable extends Page implements PageWithTableModel {
     table.off('rowAction', this._tableRowActionHandler);
     table.off('rowOrderChanged', this._tableRowOrderChangeHandler);
     table.off('reload', this._tableDataLoadHandler);
+
+    table.on('propertyChange:tableControls', this._tableControlsChangeHandler);
```
cguglielmo: on -> off

eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
```diff
@@ -103,6 +147,34 @@ export class PageWithTable extends Page implements PageWithTableModel {
     this.outline.mediator.onTableRowOrderChanged(event, this);
   }
 
+  protected _onTableControlsChange(e: PropertyChangeEvent<TableControl[]>) {
+    // disable search/reset listeners on old search form controls
+    arrays.ensure(e.oldValue)
+      .filter<SearchFormTableControl>((tableControl: TableControl): tableControl is SearchFormTableControl => tableControl instanceof SearchFormTableControl)
```
cguglielmo:
```suggestion
      .filter((tableControl: TableControl) => tableControl instanceof SearchFormTableControl)
```

eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
```diff
@@ -103,6 +147,34 @@ export class PageWithTable extends Page implements PageWithTableModel {
...
+      .forEach(searchFormTableControl => {
```
cguglielmo:
```suggestion
      .forEach((searchFormTableControl: SearchFormTableControl) => {
```

eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
```diff
@@ -103,6 +147,34 @@ export class PageWithTable extends Page implements PageWithTableModel {
...
+        searchFormTableControl.off('search', this._searchFormTableControlSearchHandler);
+        searchFormTableControl.off('reset', this._searchFormTableControlResetHandler);
+      });
+
+    // enable search/reset listeners on search form control
```
cguglielmo, line 157: extract to _add/removeSearchFormTableControlHandlers so it can be reused in init/destroyDetailTable

eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
```diff
@@ -41,15 +45,49 @@ export class PageWithTable extends Page implements PageWithTableModel {
...
+  }
+
   protected override _initDetailTable(table: Table) {
     super._initDetailTable(table);
+
+    if (this.outline.compact) {
+      // set table compact if outline is compact
+      table.setCompact(true);
+
+      // hide all table controls except
```
cguglielmo:
```suggestion
      // hide all table controls except search form table control
```

eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
```diff
@@ -161,13 +233,25 @@ export class PageWithTable extends Page implements PageWithTableModel {
     return this.getSearchFilter();
   }
 
+  /**
+   * Returns the {@link SearchFormTableControl} for the given {@link Table}, or `null` if no {@link SearchFormTableControl} is present.
+   */
+  protected _findSearchFormTableControl(table: Table): SearchFormTableControl {
```
cguglielmo, line 251: Is this method needed? I don't think it brings much benefit.

eclipse-scout-core/src/table/controls/SearchFormTableControl.ts
```diff
@@ -0,0 +1,55 @@
...
+  declare eventMap: SearchFormTableControlEventMap;
+  declare self: SearchFormTableControl;
+
+  protected _formSearchHandler: EventHandler<Event<Form>> = this._onFormSearch.bind(this);
+  protected _formResetHandler: EventHandler<Event<Form>> = this._onFormReset.bind(this);
+
+  constructor() {
+    super();
+
+    this.iconId = icons.SEARCH;
```
cguglielmo, line 23: TooltipText should be set as well

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/desktop/JsonOutline.java
```diff
@@ -245,6 +242,18 @@ protected void putDetailFormAndTable(JSONObject json, IPage<?> page) {
   protected void putJsPageObjectTypeAndModel(JSONObject json, IJsPage jsPage) {
     putProperty(json, IJsPage.PROP_JS_PAGE_OBJECT_TYPE, jsPage.getJsPageObjectType());
     putProperty(json, IJsPage.PROP_JS_PAGE_MODEL, jsonDoHelper().dataObjectToJson(jsPage.getJsPageModel()));
+
+    if (jsPage.getParentNode() instanceof IPageWithTable) {
+      JSONObject jsPageModel = ObjectUtility.nvl(json.optJSONObject(IJsPage.PROP_JS_PAGE_MODEL), new JSONObject());
+
+      // Send text because it might come from a summary column from a parent PageWithTable
```
cguglielmo:
```suggestion
      // Send the properties which might come from a summary column from a parent PageWithTable
```

eclipse-scout-core/src/desktop/outline/Outline.ts
```diff
@@ -951,18 +951,6 @@ export class Outline extends Tree implements DisplayParent, OutlineModel {
         detailTable.setMenuBarVisible(false);
         this._attachDetailMenusListener(detailTable);
       }
-      // Get single selection menus from parent detail table
```
cguglielmo: This cannot be removed, it is needed for Scout Classic. Maybe we should check for compact mode of outline in ParentTablePageMenuContributor

eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
```diff
@@ -41,15 +45,49 @@ export class PageWithTable extends Page implements PageWithTableModel {
...
+  setAlwaysCreateChildPage(alwaysCreateChildPage: boolean) {
+    this.alwaysCreateChildPage = alwaysCreateChildPage;
+  }
+
   protected override _initDetailTable(table: Table) {
     super._initDetailTable(table);
+
+    if (this.outline.compact) {
+      // set table compact if outline is compact
+      table.setCompact(true);
```
cguglielmo, line 67: The 'moreLink' should be removed because it cannot be clicked anyway, see Table.java#createSummaryCompactHandler. Maybe htmlToPlainText is also necessary

eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
fschinkel: It is as `_initDetailTable` initializes the table passed as argument and we need to find the search form of this table.

eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
```diff
@@ -42,8 +45,41 @@ export class PageWithTable extends Page implements PageWithTableModel {
...
+
+      // disable more-link and enable html to plain text on compact handler
+      table.compactHandler?.setMoreLinkAvailable(false);
```
cguglielmo: CompactHandler is only set to null for Scout Classic tables. JavaScript tables always have a compact handler -> question marks can be omitted.

eclipse-scout-core/src/table/TableCompactHandler.ts
```diff
@@ -198,11 +210,15 @@ export class TableCompactHandler implements TableCompactHandlerModel, ObjectWith
         text = '?';
       }
       line.textBlock.setText(text);
+      this.lineCustomizer?.(line);
```
cguglielmo: LineCustomizer should always be applied, not only for boolean columns.

PR #1569, opened 2025-06-26
Removed TypedId

work-lto:
The typedId is no longer necessary and is cleaned-up with this change

351084


matthiaso:
Go jenkins

org.eclipse.scout.rt.dataobject.test/src/test/java/org/eclipse/scout/rt/dataobject/fixture/DataObjectWithTypedIdDo.java
```diff
@@ -18,16 +18,15 @@
 import org.eclipse.scout.rt.dataobject.DoList;
 import org.eclipse.scout.rt.dataobject.DoValue;
 import org.eclipse.scout.rt.dataobject.TypeName;
-import org.eclipse.scout.rt.dataobject.id.TypedId;
 
 @TypeName("DataObjectWithTypedId")
 public class DataObjectWithTypedIdDo extends DoEntity {
```
matthiaso: Do we keep this class? This object does not contain a `TypedId` anymore (probably because `TypedId` does not exist anymore).

matthiaso (approved):

PR #1573, opened 2025-07-03
DataObjectInventory: prevent replacements

fschinkel:
DOs must not be loaded and registered twice. Otherwise, the inventory will always know the last loaded prototype and will use it everywhere during serialization and deserialization. If there are instances of the DOs already created these DOs may no longer be serializable, as e.g. all attributes present are checked against the current DO definition (i.e. the last loaded prototype). Prevent re-registering by throwing an error if the type name is already present.

410638

### DIFF
--- a/eclipse-scout-core/src/dataobject/DataObjectInventory.ts
+++ b/eclipse-scout-core/src/dataobject/DataObjectInventory.ts
@@ -47,6 +47,9 @@ export class DataObjectInventory {
       return false;
     }
 
+    if (this._constructorByTypeName.has(typeName)) {
+      throw new Error(`There is already a constructor registered for type name '${typeName}'.`);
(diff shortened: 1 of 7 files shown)


eclipse-scout-core/src/dataobject/DataObjectInventory.ts
```diff
@@ -47,6 +47,9 @@ export class DataObjectInventory {
       return false;
     }
 
+    if (this._constructorByTypeName.has(typeName)) {
+      throw new Error(`There is a constructor registered already for type name '${typeName}'.`);
```
cguglielmo: It think it sounds better if 'already' comes after 'is'.
There is already a constructor registered for type name '${typeName}'.

PR #1575, opened 2025-07-04
Tree: delegate unloadNode to TreeNode.unload

cguglielmo:
This makes it easier to implement page specific unload logic.

424779


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/tree/AbstractTree.java
```diff
@@ -2651,10 +2651,7 @@ public void setMultiCheck(boolean b) {
   public void unloadNode(ITreeNode node) {
     try {
       setTreeChanging(true);
-      //
-      setNodeExpanded(node, false);
-      removeAllChildNodes(node);
-      node.setChildrenLoaded(false);
+      node.unload();
```
fschinkel, line 2654: do we need a null check here?

PR #1576, opened 2025-07-04
Features/bsh/25.2/js preferences

bschwarzent:


eclipse-scout-core/src/prefs/UiPreferences.ts
```diff
@@ -0,0 +1,494 @@
...
+
+  /**
+   * Replaces the {@link UiPreferencesStore} for this singleton object. This is intended to be used in tests only.
+   * In a normal application, the store should not be changed dynamically. Instead, register the desired store
+   * implementation via {@link ObjectFactory}.
+   *
+   * **Important:** This method clears internal data structures, but does *not* automatically reload preferences
+   * from the new store. To do so, {@link load} has to be called manually.
+   *
+   * @return the old store
```
cguglielmo: @return -> @returns

eclipse-scout-core/src/prefs/UiPreferences.ts
```diff
@@ -0,0 +1,494 @@
...
+  /**
+   * Schedules a task to call {@link store}. This method is to be called whenever a preference has been changed.
+   * By scheduling a task rather than storing immediately, we can coalesce multiple store requests into a single one.
+   */
+  scheduleStore() {
+    clearTimeout(this._storeTimeoutId);
+    this._storeTimeoutId = setTimeout(() => {
+      this.store()
+        .catch(error => {
+          console.warn('Unable to store UI preferences', error);
```
cguglielmo: don't use console logs, errorHandler can log it using logError

eclipse-scout-core/src/prefs/UiPreferences.ts
```diff
@@ -0,0 +1,494 @@
...
+          }
+        }
+        scout.create(ErrorHandler, {displayError: false, logError: true}).handle(`Unable to map filter to data object [table=${table.id}, filterType=${filter?.filterType}, filterLabel=${filter?.createLabel()}`);
+        return null;
+      })
+      .filter(Boolean);
+  }
+
+  /**
+   * If the table is customizable, return sthe customizer data. Otherwise, `null` is returned.
```
cguglielmo: return sthe -> returns the

eclipse-scout-core/src/prefs/UiPreferences.ts
```diff
@@ -0,0 +1,494 @@
...
+   */
+  storeTablePreferenceProfile(table: Table, profileId: string, profile: TableClientUiPreferenceProfileDo) {
+    if (!profileId || !profile) {
+      return;
+    }
+
+    let prefs = this.getOrCreateTablePreferences(table);
+    let existingProfile = this.getTablePreferencesProfile(prefs, profileId);
+    if (existingProfile) {
+      if (objects.equals(profile, existingProfile)) {
```
cguglielmo: I think profile.equals should be preferred

eclipse-scout-core/src/table/organizer/TableOrganizer.ts
```diff
@@ -208,18 +210,18 @@ export class TableOrganizer implements ObjectWithType {
...
-  modifyColumn(column: Column<any>) {
-    if (this.table.isCustomizable()) {
-      // TODO bsh [js-table] Delegate to this.table.tableCustomizer
-    } else {
-      // NOP (currently not supported)
+  modifyColumn(column: Column<any>): JQuery.Promise<void> {
+    if (this.table.isCustomizable() && this.table.customizer.isCustomizable(column)) {
+      return this.table.customizer.modifyColumn(column);
     }
+    // NOP (currently not supported)
```
cguglielmo: Is this comment necessary? It will never be supported I assume

eclipse-scout-core/src/table/Table.ts
```diff
@@ -6156,8 +6466,90 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
-    // TODO bsh [js-table] Delegate to this.customizer
-    return false;
+    return !!this.customizer;
+  }
+
+  setUiPreferencesEnabled(uiPreferencesEnabled: boolean) {
+    this.setProperty('uiPreferencesEnabled', !!uiPreferencesEnabled);
+  }
+
+  _setUiPreferencesEnabled(uiPreferencesEnabled: boolean) {
```
cguglielmo: missing protected modifier

eclipse-scout-core/src/table/Table.ts
```diff
@@ -6156,8 +6466,90 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+    uiPreferences.applyTablePreferencesProfile(this, this._initialUiPreferences, {
+      applyUserFilters: true
+    });
+    uiPreferences.clearGlobalTablePreferenceProfile(this);
+  }
+
+  applyColumnPreferences(columnPreferences: TableColumnClientUiPreferenceDo[]) {
+    let columnPreferencesMap = new Map(arrays.ensure(columnPreferences).map(pref => [pref.columnId, pref]));
+    let newColumns = this.columns
+      .filter(c => !c.guiOnly)
```
cguglielmo: maybe map to \[column, prefs] first to simplify the sort and map functions? 
.map(column => \[column, columnPreferencesMap.get(column.buildUuid())] as \[Column<any>, TableColumnClientUiPreferenceDo])

eclipse-scout-core/src/data-objects.ts
```diff
@@ -188,3 +188,21 @@ export class TextColumnUserFilterStateDo extends BaseDoEntity implements IUserFi
 
 export interface ITableCustomizerDo extends BaseDoEntity {
 }
+
+// --------------------------------------------------
+
+@typeName('scout.UiPreferences')
+export class UiPreferencesDo extends BaseDoEntity {
+  tablePreferences?: TableClientUiPreferencesDo[];
+  // --- Add when necessary: ---
```
cguglielmo: I would remove this uncommented code

eclipse-scout-core/src/scout.ts
```diff
@@ -73,6 +73,13 @@ export interface ObjectWithUuid {
    */
   uuid: string;
 
+  /**
+   * Widget to be used instead of the default parent when computing the uuid path for the object.
```
cguglielmo: Widget -> Object

eclipse-scout-core/src/scout.ts
```diff
@@ -118,6 +125,12 @@ export interface ObjectModelWithUuid<TObject = object> extends ObjectModel<TObje
    * A unique identifier for the object. Typically, a new random UUID can be used.
    */
   uuid?: string;
+  /**
+   * Widget to be used instead of the default parent when computing the uuid path for the object.
```
cguglielmo: Widget -> Object

eclipse-scout-core/src/table/userfilter/UserFilterStateMapper.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  tryFromDo(table: Table, filterState: IUserFilterStateDo): TableUserFilter {
+    if (this._acceptFilterState(filterState)) {
+      return this._fromDo(table, filterState);
+    }
+    return null;
+  }
+
+  protected _createColumnFilter(table: Table, columnId: string, model: any): TFilter {
```
cguglielmo: This method is not used, column filter state mappers should probably use it?

eclipse-scout-core/src/table/Table.ts
```diff
@@ -4247,6 +4291,27 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+          newFilters.push(filter);
+          return;
+        }
+      }
+      scout.create(ErrorHandler, {displayError: false, logError: true}).handle(`No mapper found for user filter state: ${dataObjects.stringify(filterState)}`);
+    });
+
+    // Update filters
+    this.setFilters(newFilters);
+    this.filter();
```
cguglielmo: Calling this.filter() is not necessary, setFilters already does it

eclipse-scout-core/src/table/Table.ts
```diff
@@ -6156,8 +6466,90 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+      return; // can happen if preferences are applied before the customizer is installed or if preferences contains obsolete data
+    }
+
+    // Use setter for 'visible' property because it is a multidimensional property
+    column.setVisible(columnPreferences.visible, false); // parameter 'false' skips call of onColumnVisibilityChanged()
+    // FIXME bsh [js-preferences] this triggers an unnecessary change. As a matter of fact, _renderViewport is called too many times anyway!
+    column.setWidth(columnPreferences.width);
+
+    // Properties without setter (changes will be applied later by _setColumns)
+    column.sortIndex = columnPreferences.sortOrder;
```
cguglielmo, line 6556: sortActive needs to be set as well, otherwise sorting won't be applied correctly

eclipse-scout-core/src/table/Table.ts
```diff
@@ -6156,8 +6466,90 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+
+  protected _applyColumnPreferencesToColumn(columnPreferences: TableColumnClientUiPreferenceDo, column: Column<any>) {
+    if (!columnPreferences || !column) {
+      return; // can happen if preferences are applied before the customizer is installed or if preferences contains obsolete data
+    }
+
+    // Use setter for 'visible' property because it is a multidimensional property
+    column.setVisible(columnPreferences.visible, false); // parameter 'false' skips call of onColumnVisibilityChanged()
+    // FIXME bsh [js-preferences] this triggers an unnecessary change. As a matter of fact, _renderViewport is called too many times anyway!
+    column.setWidth(columnPreferences.width);
```
cguglielmo: I think it is ok to just assign the width without using the setter, I don't think we need to trigger a resize event

eclipse-scout-core/src/table/Table.ts
```diff
@@ -6156,8 +6466,90 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+
+    // Properties without setter (changes will be applied later by _setColumns)
+    column.sortIndex = columnPreferences.sortOrder;
+    column.sortAscending = columnPreferences.sortAscending;
+    column.grouped = columnPreferences.groupingActive;
+
+    if (column instanceof NumberColumn) {
+      column.setAggregationFunction(columnPreferences.aggregationFunctionId as NumberColumnAggregationFunction);
```
cguglielmo, line 6563: Maybe add comment that setter is required to update the aggregation and background effect functions (e..g column.aggrStart etc.)

eclipse-scout-core/src/prefs/UiPreferences.ts
bschwarzent: The idea was to print a meaningful message instead of only the original error (much might just be a network error without any context). But I agree that outputting to the browser console is often not that useful. I removed it.

eclipse-scout-core/src/table/organizer/TableOrganizer.ts
bschwarzent: Comment adjusted

eclipse-scout-core/src/table/Table.ts
bschwarzent: I'm not sure if all those brackets make it easier to read. I added some comments instead. The column properties are now applied in a separate forEach loop. I think it's pretty readable now.

eclipse-scout-core/src/table/userfilter/UserFilterStateMapper.ts
bschwarzent: You are correct, that method is left over from a previous version. It didn't seem to make the individual mappers much simpler, so I reverted it but forget to delete this method. It is now removed.

eclipse-scout-core/src/table/Table.ts
bschwarzent: Hm, you seem to be correct. I one had a situation where just calling setFilters() would not apply them, but I can't reproduce it anymore. Therefore, I removed the filter() call, as suggested.

eclipse-scout-core/src/table/Table.ts
bschwarzent: Ok, I expected this to be a "computed" property which is automatically set. It is now also set here, derived from `sortIndex >= 0`.

PR #1580, opened 2025-07-07
PageWithNodes: automatically detect reloadable flag

bschwarzent:
Pages with static child pages (from the model) cannot be reloaded, because the initial model is not known anymore and thus, the child page cannot be recreated. However, when child pages are created dynamically by implementing the method _createChildPages(), reloading _is_ possible. To avoid requiring every implementation to manually set the 'reloadable' flag to true (which can easily be forgotten), we check if the method _createChildPages was overwritten. If yes, we assume that the page is reloadable. In case that detection is incorrect, it still can be manually be set to false again.


cguglielmo (approved):

bschwarzent:
Run tests

PR #1582, opened 2025-07-07
Increase default max header size

work-lto:
423941


org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ApplicationProperties.java
```diff
@@ -218,6 +218,11 @@ public String getKey() {
     public String description() {
       return "Specifies the maximum allowed size in bytes for a HTTP request header. The default value is 8 KB (see " + HttpConfiguration.class.getName() + ").";
```
rogersteblerbsi: The description should also be updated

work-lto: Done

rogersteblerbsi (approved):

PR #1584, opened 2025-07-09
Add title to ITreeNode

work-lto:
Title gives you the possibility to differentiate between Text (especially when it contains html content) and what the cell text represents

319634

### DIFF
--- a/eclipse-scout-core/src/tree/TreeNode.ts
+++ b/eclipse-scout-core/src/tree/TreeNode.ts
@@ -32,6 +32,7 @@ export class TreeNode implements TreeNodeModel, ObjectWithType {
   parentNode: TreeNode;
   session: Session;
   text: string;
+  plainTextTitle: string;
   tooltipText: string;
   foregroundColor: string;
   backgroundColor: string;
@@ -87,6 +88,7 @@ export class TreeNode implements TreeNodeModel, ObjectWithType {
     this.rendered = false;
     this.session = null;
     this.text = null;
+    this.plainTextTitle = null;
 
     this._loadChildrenPromise = null;
   }
@@ -225,6 +227,10 @@ export class TreeNode implements TreeNodeModel, ObjectWithType {
 
   setText(text: string) {
     this.text = text;
+  }
+
+  setPlainTextTitle(plainTextTitle: string) {
+    this.plainTextTitle = plainTextTitle;
   }
 
   /**
--- a/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/tree/AbstractTreeNode.java
+++ b/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/tree/AbstractTreeNode.java
@@ -77,6 +77,7 @@ public abstract class AbstractTreeNode implements ITreeNode, ICellObserver, ICon
   private final OptimisticLock m_childrenLoadedLock;
   private final Cell m_cell;
 
+  private String m_plainTextTitle;
   private List<ITreeNode> m_childNodeList;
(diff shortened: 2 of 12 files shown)


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/tree/AbstractTreeNode.java
```diff
@@ -76,6 +76,7 @@ public abstract class AbstractTreeNode implements ITreeNode, ICellObserver, ICon
   private final Object m_filteredChildNodesLock;
   private final OptimisticLock m_childrenLoadedLock;
   private final Cell m_cell;
+  private String m_plainTextTitle;
```
rogersteblerbsi: Probably better to move it down to the other non-final fields. Maybe you can just move it one line down, so that it's still directly after `m_cell` but in the non-final section :)

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/tree/JsonTree.java
```diff
@@ -637,10 +637,11 @@ protected void handleModelNodesUpdated(TreeEvent event) {
       JSONObject jsonNode = new JSONObject();
       putProperty(jsonNode, "id", nodeId);
       // Only send _some_ of the properties. Everything else (e.g. "checked", "expanded") will be handled with separate events.
-      // --> See also: Tree.js/_onNodesUpdated()
+      // --> See also: Tree.js/_onNodesUpdate
```
rogersteblerbsi: This line was changed by Matthias Villiger on 25.2 short time ago to:

`// --> See also: Tree.ts/_applyUpdatedNodeProperties()`

Should it also be the same here?

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/bookmark/view/BookmarkViewForm.java
```diff
@@ -172,18 +172,14 @@ protected void execChangedDisplayText() {
...
             return text == null || m_lowercaseFilterPattern == null || m_lowercaseFilterPattern.matcher(text.toLowerCase()).matches();
           }
         }
 
         @Order(20)
         @ClassId("abaae7ed-f148-4368-94ac-e53710b25b24")
         public class UserBookmarkTreeField extends AbstractBookmarkTreeField {
-          @Override
-          protected int getConfiguredGridW() {
-            return 1;
```
rogersteblerbsi, line 185: This change does not appear to be related to the `plainTextTitle` change. Is it correct, or an accidental change? :)

eclipse-scout-core/src/tree/TreeNode.ts
```diff
@@ -87,6 +88,7 @@ export class TreeNode implements TreeNodeModel, ObjectWithType {
     this.rendered = false;
     this.session = null;
     this.text = null;
+    this.plainTextTitle = null;
```
rogersteblerbsi, line 91: This should probably also have a setter, so that it matches the Java class.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/bookmark/view/BookmarkViewForm.java
work-lto: I wanted to get rid of the Warning because the super method looks identical

rogersteblerbsi (approved):

PR #1592, opened 2025-07-14
Tree: ensure node is visible when selected if parent is lazily expanded

cguglielmo:
Use case:
- set lazyExpandingEnabled on parent node
- expand parent nodes
- select child node -> child node is not visible

Reason: when the lazy parent is expanded, there is no selected node yet. Because of the laziness, no child nodes are shown which is correct. When the child node is selected and not visible, it tries to expand all parent nodes, but they are already expanded, so nothing happens. Instead, it needs to trigger the LazyNodeFilter because this filter will now return true for the child node because it is now selected.

424792

### DIFF
--- a/eclipse-scout-core/src/tree/Tree.ts
+++ b/eclipse-scout-core/src/tree/Tree.ts
@@ -2130,20 +2130,25 @@ export class Tree extends Widget implements TreeModel, Filterable<TreeNode> {
     this._nodesSelectedInternal(nodes);
     this._triggerNodesSelected(debounceSend);
 
-    if (this.selectedNodes.length > 0 && !this.visibleNodesMap[this.selectedNodes[0].id]) {
-      this._expandAllParentNodes(this.selectedNodes[0]);
-    }
-
-    if (this.selectedNodes.length === 1) {
-      aria.linkElementWithActiveDescendant(this.$container, this.selectedNodes[0].$node);
+    let selectedNode = this.selectedNode();
+    if (selectedNode) {
+      aria.linkElementWithActiveDescendant(this.$container, selectedNode.$node);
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/src/tree/Tree.ts
```diff
@@ -2130,20 +2130,25 @@ export class Tree extends Widget implements TreeModel, Filterable<TreeNode> {
     this._nodesSelectedInternal(nodes);
     this._triggerNodesSelected(debounceSend);
 
-    if (this.selectedNodes.length > 0 && !this.visibleNodesMap[this.selectedNodes[0].id]) {
-      this._expandAllParentNodes(this.selectedNodes[0]);
-    }
-
-    if (this.selectedNodes.length === 1) {
-      aria.linkElementWithActiveDescendant(this.$container, this.selectedNodes[0].$node);
+    let selectedNode = this.selectedNodes[0];
```
bschwarzent:
```suggestion
    let selectedNode = this.selectedNode();
```

eclipse-scout-core/src/tree/Tree.ts
```diff
@@ -2130,20 +2130,25 @@ export class Tree extends Widget implements TreeModel, Filterable<TreeNode> {
...
+    if (selectedNode) {
+      aria.linkElementWithActiveDescendant(this.$container, selectedNode.$node);
```
bschwarzent, line 2135: We should probably un-link the previous node when selectedNode is null (e.g. when `tree.deselectAll()` is called). But it seems the aria utility is not capable of that?!

cguglielmo: probably, can you create a separate ticket for this?

PR #1603, opened 2025-07-21
Improve thread safety of SharedVariableMap

rogersteblerbsi:
- Synchronize all methods accessing m_variables to prevent race conditions
- Add Javadoc clarifying synchronization requirements on m_variables
- Ensure property change events fire consistently after modifications

413519

### DIFF
--- a/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/services/common/context/SharedVariableMap.java
+++ b/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/services/common/context/SharedVariableMap.java
@@ -28,6 +28,18 @@ public class SharedVariableMap implements Serializable, Map<String, Object> {
   private static final long serialVersionUID = 1L;
   public static final String PROP_VALUES = "values";
 
+  /**
+   * Internal variable map.
+   * <p>
+   * Access to this map is synchronized on the instance of {@link SharedVariableMap}
+   * to ensure thread safety for all operations that modify or read the map.
+   * Since {@link HashMap} is not thread-safe, this synchronization protects
+   * against concurrent modifications and ensures visibility of changes across threads.
+   * <p>
+   * <b>Important:</b> Any new method that accesses or modifies {@code m_variables}
+   * <b>must</b> be properly synchronized on the {@link SharedVariableMap} instance.
+   * Omitting synchronization may cause race conditions or data inconsistency.
+   */
   private final Map<String, Object> m_variables;
   private transient BasicPropertySupport m_propertySupport;
 
@@ -52,7 +64,7 @@ public class SharedVariableMap implements Serializable, Map<String, Object> {
   /**
    * Update values of this variable map with the new one.
    */
-  public void updateInternal(Map<String, Object> newMap) {
+  public synchronized void updateInternal(Map<String, Object> newMap) {
     if (m_variables.equals(newMap)) {
       return; // nothing changed
     }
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/services/common/context/SharedVariableMap.java
```diff
@@ -53,25 +54,29 @@ public void removePropertyChangeListener(PropertyChangeListener listener) {
    * Update values of this variable map with the new one.
    */
   public void updateInternal(Map<String, Object> newMap) {
-    if (m_variables.equals(newMap)) {
-      return; // nothing changed
+    synchronized (m_variables) {
+      if (m_variables.equals(newMap)) {
+        return; // nothing changed
+      }
+      m_variables.clear();
```
matthiaso: After we talked about it, a second thought:

- If we really want a thread-safe implementation all _m_variables_ accessing methods probably need to be synchronized on _m_variables_
- Otherwise there is a short time-frame here where _updateInternal_ is called, the _m_variables_ Map is already cleared - and now another thread could call _containsKey_, _containsValue_, _get_ or _isEmpty_, _size_, _toString_ (which are currently not synchronized on _m_variables_) and could return a wrong result.
- However if all methods accessing m_variables are synchronized themselves, maybe _m_variables_ does not even need to be a _synchronizedMap_ anymore?
- And why do we synchronize especially on _m_variables_? We could as well just a _synchronized_ modifier to the method declarations (and sychronize on the instance of the _SharedVariableMap_ itself) as all synchronized blocks seem to synchronize the full method anyways?

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/services/common/context/SharedVariableMap.java
```diff
@@ -53,25 +54,29 @@ public void removePropertyChangeListener(PropertyChangeListener listener) {
...
     }
-
-    m_variables.clear();
-    putAll(newMap); // fires a value changed event
   }
 
   private void fireValuesChanged() {
-    m_propertySupport.firePropertyChange(PROP_VALUES, null, CollectionUtility.copyMap(m_variables));
+    // Already synchronized by the caller
+    m_propertySupport.firePropertyChange(PROP_VALUES, null, Collections.synchronizedMap(CollectionUtility.copyMap(m_variables)));
```
matthiaso: Maybe synchronize here as well? Double synchronization should not harm, usually caller should be responsible for synchronization - but if it is called w/o we would ensure synchronization here.

Also: Do we really need to return a synchronizedMap here?

matthiaso (changes requested):

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/services/common/context/SharedVariableMap.java
rogersteblerbsi: Now it just uses a simple `HashMap` and all methods accessing the `m_variables` are synchronized.

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/services/common/context/SharedVariableMap.java
rogersteblerbsi: See comment above.

matthiaso (approved):

PR #1604, opened 2025-07-21
HtmlHelper.toPlainText: correctly decode character references

bschwarzent:
Both named and decimal character references should be decoded to their original character when converting HTML to plain text. While the numeric references can be decoded automatically, the list of named references is fixed and specified in an RFC. The previously supported list of characters was not complete. This is now fixed.

Additionally, special handling of non-breaking spaces was removed. If they are present in HTML, they should still be present in the plain text as their unicode equivalent \u00A0, because they are not collapsed when the HTML is rendered.

413114


eclipse-scout-core/test/encoder/PlainTextEncoderSpec.ts
```diff
@@ -259,4 +269,23 @@ describe('PlainTextEncoder', () => {
     htmlText = '<\tabc attr=\'someText\'>';
     expect(encoder.removeAttributeValues(htmlText)).toBe('<\tabc attr=\'someText\'>');
   });
+
+  it('trims lines, but preserves other white-space', () => {
+    expect(encoder.encode('hello')).toBe('hello');
+    expect(encoder.encode('one\ntwo')).toBe('one\ntwo');
+    expect(encoder.encode('one\r\ntwo')).toBe('one\ntwo');
+    expect(encoder.encode('one\rtwo')).toBe('one\ntwo');
+    expect(encoder.encode('one   two')).toBe('one   two');
```
nsteger, line 278: are these special spaces? in this case may add a comment? otherwise I would expect that duplicated spaces are deleted..

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/html/HtmlEntities.java
```diff
@@ -0,0 +1,121 @@
...
+        break;
+      }
+
+      String encoded = sb.substring(start, end + 1);
+      String decoded = decodeNamedCharacterReference(encoded); // Named character reference
+      if (decoded == null) {
+        decoded = decodeNumericCharacterReference(encoded); // Numeric character reference
+      }
+      if (decoded == null) {
+        start = end + 1;
```
nsteger, line 81: what if string is for example `"&&auml;"` ? I would expect the result "&ä", but I think the algorithm will return `"&&auml;"` ?

nsteger: => yes, I added `assertEquals("&ä", entities.unescapeAll("&&auml;"));` in HtmlEntitiesTest, which fails

org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/html/HtmlEntitiesTest.java
```diff
@@ -0,0 +1,132 @@
...
+import static org.junit.Assert.*;
+
+import org.eclipse.scout.rt.platform.BEANS;
+import org.eclipse.scout.rt.testing.platform.runner.PlatformTestRunner;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@SuppressWarnings({"ConcatenationWithEmptyString", "SpellCheckingInspection", "TextBlockMigration"})
+@RunWith(PlatformTestRunner.class)
+public class HtmlEntitiesTest {
```
nsteger, line 21: may  add some of these tests to ts spec? (plain text encoder should do same transformations)

bschwarzent:
Replaced by https://github.com/eclipse-scout/scout.rt/pull/1683

PR #1608, opened 2025-07-22
Page: collapse lazy expanded nodes on reload

bschwarzent:
When a page is expanded lazily (i.e. the node is expanded, but only some of the child nodes are visible, marked by "+" instead of "v" icon) and the table is reloaded, all child pages vanish because they are replaced by new instances. Because the parent node stays expanded, this leaves the "+" icon without any visible child nodes, which is confusing.

To prevent this unwanted state, a lazily expanded page is now automatically collapsed when the data is reloaded. Nothing changes if the page is already fully expanded (it stays expanded).

221481


cguglielmo (commented):
Looks good, but is the ticket reference in the commit comment correct?

bschwarzent:
> Looks good, but is the ticket reference in the commit comment correct?

No, thanks for pointing that out.

The change now also contains some unit tests (Java and JS).

PR #1609, opened 2025-07-24
Table: improve tile mode when column structure changes

fschinkel:
When columns are modified using setColumns do not call _renderColumns if data is not rendered.
Update groupBy and sortBy field in TileTableHeaderBox when column structure changes.

378974

### DIFF
--- a/eclipse-scout-core/src/table/TileTableHeaderBox.ts
+++ b/eclipse-scout-core/src/table/TileTableHeaderBox.ts
@@ -8,8 +8,8 @@
  * SPDX-License-Identifier: EPL-2.0
  */
 import {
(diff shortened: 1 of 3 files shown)


eclipse-scout-core/src/table/TileTableHeaderBox.ts
```diff
@@ -103,6 +103,22 @@ export class TileTableHeaderBox extends GroupBox implements TileTableHeaderBoxMo
 
     // it's okay to sync the fields here, _onGroupingChange/_onSortingChange will return early since the tileMode property is not set yet at this point
     this._syncSortingGroupingFields();
+
```
cguglielmo: Add in _installListeners and remove in _uninstallListeners due to consistency reasons? Even though I don't think it is necessary to remove the listeners at all, because a destroyed table should not trigger any events anymore. But I don't know why it was added in the first place.

cguglielmo (approved):

PR #1610, opened 2025-07-24
DataObjectDeserializer: add retainTypeVersion-option

fschinkel:
Typically, the _typeVersion can be removed when deserializing data objects. But there may be existing logic that relies on the _typeVersion being present on pojo DOs (e.g. for equals checks). Therefore, introduce the new option retainTypeVersion to enable/disable the removal of the _typeVersion during deserialization. The default value of this option is a predicate that ignores the _typeVersion if the deserialized DataObject is a BaseDoEntity.

410638

### DIFF
--- a/eclipse-scout-core/src/dataobject/serialize/DataObjectDeserializer.ts
+++ b/eclipse-scout-core/src/dataobject/serialize/DataObjectDeserializer.ts
@@ -7,7 +7,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-import {BaseDoEntity, Constructor, dataObjects, DoValueMetaData, doValueMetaData, InitModelOf, ObjectModel, objects, ObjectWithType, scout} from '../../index';
+import {BaseDoEntity, Constructor, dataObjects, DoValueMetaData, doValueMetaData, InitModelOf, ObjectModel, objects, ObjectWithType, Predicate, scout} from '../../index';
 
 export class DataObjectDeserializer implements DataObjectDeserializerModel, ObjectWithType {
 
@@ -16,9 +16,15 @@ export class DataObjectDeserializer implements DataObjectDeserializerModel, Obje
   id: string;
   objectType: string;
   createPojoIfDoIsUnknown: boolean;
+  retainTypeVersion: Predicate<object>;
 
   constructor(model?: InitModelOf<DataObjectDeserializer>) {
     this.createPojoIfDoIsUnknown = !!model?.createPojoIfDoIsUnknown;
+    this.retainTypeVersion = scout.nvl(model?.retainTypeVersion, (obj: object) => !(obj instanceof BaseDoEntity));
+    if (!objects.isFunction(this.retainTypeVersion)) {
+      const retain = !!this.retainTypeVersion;
+      this.retainTypeVersion = () => retain;
+    }
   }
 
   deserialize<T extends object>(value: any, valueMetaData?: DoValueMetaData<T>): T {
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/src/dataobject/serialize/DataObjectDeserializer.ts
```diff
@@ -16,9 +16,11 @@ export class DataObjectDeserializer implements DataObjectDeserializerModel, Obje
   id: string;
   objectType: string;
   createPojoIfDoIsUnknown: boolean;
+  retainTypeVersion: boolean;
 
   constructor(model?: InitModelOf<DataObjectDeserializer>) {
     this.createPojoIfDoIsUnknown = !!model?.createPojoIfDoIsUnknown;
+    this.retainTypeVersion = !!scout.nvl(model?.retainTypeVersion, this.createPojoIfDoIsUnknown);
```
mvilliger: To use 'createPojoIfDoIsUnknown' as fallback also means that if I only change this property, the deserialized content may change as well. Not only BaseDoEntity -> Pojo. It also has the side effect, on the _typeVersion. I'm not sure if this is expected from an API point of view: just changing createPojoIfDoIsUnknown=true may also lead to new attributes.
Is this default required?

eclipse-scout-core/src/dataobject/serialize/DataObjectDeserializer.ts
```diff
@@ -76,14 +79,20 @@ export class DataObjectDeserializer implements DataObjectDeserializerModel, Obje
 export interface DataObjectDeserializerModel extends ObjectModel<DataObjectDeserializer> {
   /**
    * Controls the kind of object that will be created when deserializing unknown DataObjects.
-   * If true, a pojo will be created for unknown DOs. If false, instances of {@link BaseDoEntity} will be created.
-   * Default is false.
+   * If <code>true</code>, a pojo will be created for unknown DOs. If <code>false</code>, instances of {@link BaseDoEntity} will be created.
+   * Default is <code>false</code>.
```
mvilliger: I think we use no longer the HTML syntax but the markdown syntax using backticks now?
If we really keep the default fallback for retainTypeVersion to createPojoIfDoIsUnknown (see comment above) then I would clearly indicate this here: Enabling the property createPojoIfDoIsUnknown may also have an effect on the _typeVersion attribute and therefore on the attributes of the deserialized objects. Maybe this side-effect is not expected and it would be better if the default is just 'false'?

eclipse-scout-core/src/dataobject/serialize/DataObjectDeserializer.ts
```diff
@@ -76,14 +79,20 @@ export class DataObjectDeserializer implements DataObjectDeserializerModel, Obje
...
    *   <li>The _type attribute of the object to deserialize cannot be found in TS.</li>
    *   <li>There is no _type attribute and the TypeScript attribute declaration (metadata) does not provide a class type which could be used as fallback.</li>
    * </ol>
    */
-  createPojoIfDoIsUnknown: boolean;
+  createPojoIfDoIsUnknown?: boolean;
+  /**
+   * Controls whether the '_typeVersion' property is removed or not when deserializing DataObjects.
+   * If <code>true</code>, the '_typeVersion' will be set on the deserialized DataObject. If <code>false</code>, it will be ignored.
+   * Default is <code>false</code> if {@link createPojoIfDoIsUnknown} is <code>false</code> and <code>true</code> otherwise.
```
mvilliger: Also use backticks here instead of &lt;code>?
And you can remove the "or not" part: wheter .. is removed when ...

eclipse-scout-core/src/dataobject/serialize/DataObjectDeserializer.ts
```diff
@@ -76,14 +88,22 @@ export class DataObjectDeserializer implements DataObjectDeserializerModel, Obje
...
    * A DataObject can be unknown e.g. if:
    * <ol>
    *   <li>The _type attribute of the object to deserialize cannot be found in TS.</li>
    *   <li>There is no _type attribute and the TypeScript attribute declaration (metadata) does not provide a class type which could be used as fallback.</li>
    * </ol>
    */
-  createPojoIfDoIsUnknown: boolean;
+  createPojoIfDoIsUnknown?: boolean;
+  /**
+   * Controls whether the '_typeVersion' property is removed when deserializing DataObjects.
```
mvilliger: whether the property .. is **retained** when...

mvilliger (approved):

PR #1614, opened 2025-07-28
Features/cgu/26.1/readonly input

cguglielmo:


eclipse-scout-core/src/form/fields/fields.ts
```diff
@@ -175,5 +176,27 @@ export const fields = {
...
+      return true;
+    }
+    if (!$elem.is('input')) {
+      return false;
+    }
+    if (!$elem.attr('type')) { // input without type defaults to text
+      return true;
+    }
+    // According to https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/readonly
+    let types = ['text', 'search', 'tel', 'url', 'email', 'password', 'date', 'month', 'week', 'time', 'datetime-local', 'datetime-number'];
```
bschwarzent: Shorter and more efficient with `scout.isOneOf` or `Array.prototype.includes`

eclipse-scout-core/src/form/fields/fields.ts
```diff
@@ -175,5 +176,27 @@ export const fields = {
...
+   */
+  supportsReadOnlyAttribute($element: JQuery<HTMLInputElement | HTMLTextAreaElement>) {
+    let $elem = $.ensure($element);
+    if (!$elem.length) {
+      return false;
+    }
+    if ($elem.is('textarea')) {
+      return true;
+    }
+    if (!$elem.is('input')) {
```
bschwarzent: Can this happen? The parameter type indicates that only input and text area elements can be passed. If this check should stay, maybe allow all jquery elements in the parameter?

eclipse-scout-core/src/form/fields/fields.ts
```diff
@@ -175,5 +176,27 @@ export const fields = {
   focusAndInsertText($input: JQuery<HTMLInputElement | HTMLTextAreaElement>, text: string) {
     $input[0].focus();
     $input.document(true).execCommand('insertText', false, text);
+  },
+
+  /**
+   * @returns true if the element supports the readonly attribute
+   */
+  supportsReadOnlyAttribute($element: JQuery<HTMLInputElement | HTMLTextAreaElement>) {
```
bschwarzent: The attribute is called "readonly" (one word) - maybe reflect that in the method name?

eclipse-scout-core/src/focus/focusUtils.ts
```diff
@@ -89,9 +89,10 @@ export const focusUtils = {
       return false;
     }
 
-    if ($element.is('input[disabled][type=text], textarea[disabled]')) {
+    if ($element.is('input[disabled][type=text], textarea[disabled], input[readonly], textarea[readonly]')) {
```
bschwarzent: * For inputs that are disabled, this code requires type=text, but for readonly it does not. Is this intentional?
* What about all the other text-like types, e.g. `tel`, `email`, `search` etc.? The `type` attribute can even be entirely absent, shouldn't this still be considered to be selectable?
* Why are we even considering the enabled-ness of the field? Don't we _always_ want to have input fields selectable?

eclipse-scout-core/src/jquery/jquery-scout.js
```diff
@@ -696,12 +696,29 @@ $.fn.setEnabled = function(enabled) {
...
+      this._updateTabIndexForReadOnly(!enabled);
+    } else {
+      this.toggleAttr('disabled', !enabled);
+    }
   }
   this.trigger(enabled ? 'enable' : 'disable');
   return this;
 };
 
+$.fn._updateTabIndexForReadOnly = function(readonly) {
```
bschwarzent: Consider removing the readonly parameter. Instead, the value could determined automatically:
```
const readonly = this.attr('readonly');
return this.attr('tabindex', readonly ? -1 : null);
```

This would make the two calls simpler, especially in setTabbable():

```
$.fn.setTabbable = function(tabbable) {
  this.attr('tabindex', tabbable ? 0 : null);
  // Re-evaluate tabindex considering readonly state
  this._updateTabIndexForReadOnly();
  return this;
};
```

eclipse-scout-core/src/focus/focusUtils.ts
cguglielmo: As already discussed:
Yes, it is intentional, because I wanted to support other types for readonly inputs as well.
I don't know why they are not supported for disabled inputs, probably because we don't use them.
Actually, the code for disabled inputs is not necessary at all because the browser does not trigger the mouse down handler if the input is disabled. Maybe the browsers changed their behavior in the last years -> I removed it completely. 
Moved the code to FocusManager so we only have to handle the case specific for the mouse down handler.

PR #1619, opened 2025-07-29
Improve preserveAndSetProperty

cguglielmo:
The function only works if the preserved value is initialized to null. This is not intuitive and does not work if the value should be preserved without adjusting the preserver.
Also, it is not possible to preserve null values.

428304

### DIFF
--- a/eclipse-scout-core/src/form/fields/groupbox/GroupBox.ts
+++ b/eclipse-scout-core/src/form/fields/groupbox/GroupBox.ts
@@ -84,7 +84,6 @@ export class GroupBox extends CompositeField implements GroupBoxModel {
     this.$body = null;
     this.$title = null;
     this.$subLabel = null;
-    this._statusPositionOrig = null;
   }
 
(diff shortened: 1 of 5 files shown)


fschinkel (commented):
The commit message describes the problem in present tense. Can you please change this to past tense and describe what has changed in order to solve this issue.

In addition, please add a migration guide.
Do we want to release this fix in a minor build as it requires a code migration?

cguglielmo:
> The commit message describes the problem in present tense. Can you please change this to past tense and describe what has changed in order to solve this issue.

A problem is typically described in present tense, this is correct. The message now contains an explanation what has changed.

> In addition, please add a migration guide. Do we want to release this fix in a minor build as it requires a code migration?

As discussed: we will do it only for 26.1.
26.1 PR: https://github.com/eclipse-scout/scout.rt/pull/1717
Migration Guide PR: https://github.com/eclipse-scout/scout.docs/pull/333

PR #1627, opened 2025-08-04
Features/cgu/25.2/428727 table editor glasspane

cguglielmo:


eclipse-scout-core/src/util/arrays.ts
```diff
@@ -608,16 +608,19 @@ export const arrays = {
...
       }
     }
+    return -1;
   },
 
   /**
    * Replacement for remove() that works for arrays of jQuery objects (compares DOM nodes).
    */
-  $remove(arr: JQuery[], $element: JQuery) {
+  $remove(arr: JQuery[], $element: JQuery): boolean {
```
bschwarzent, line 621: Document (or remove) return value.

(Same for $indexOf)

PR #1630, opened 2025-08-05
Fix row filter for tables in tile mode

nsteger:
TableTileGridMediator now also handles
TableEvent.TYPE_ROW_FILTER_CHANGED such that tiles are indeed filtered by a programmed row filter.

427908

### DIFF
--- a/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/TableTileGridMediator.java
+++ b/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/TableTileGridMediator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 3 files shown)


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/TableTileGridMediator.java
```diff
@@ -89,6 +89,22 @@ public void tableChanged(TableEvent e) {
...
+          // delete tiles which are not accepted by filter
+          Predicate<ITableRowTileMapping> filteredOutPredicate = m -> !filteredRows.contains(m.getTableRow());
+          tileMappings.stream().filter(filteredOutPredicate).forEach(m -> m.getTile().dispose());
+          tileMappings.removeIf(filteredOutPredicate);
+
+          // insert non-existing tiles
+          filteredRows.removeAll(tileMappings.stream().map(ITableRowTileMapping::getTableRow).collect(Collectors.toList()));
+          tileMappings.addAll(m_table.createTiles(filterTopLevelTableRows(filteredRows)));
+
+          setTileMappings(tileMappings);
```
bschwarzent, line 107: Although this is the last case in the switch, add an explicit `break` at the end of this block. It does not matter now, but if in the future someone adds another case, they might easily miss adding the necessary break, causing unexpected behavior.

PR #1631, opened 2025-08-05
Revert adding property to ITreeNode

work-lto:
The property plainTextTitle was reverted. If you need to give a plain text title you can override the method toPlainText.

319634


cguglielmo (changes requested):
Please also revert the changes in TreeNode.ts

cguglielmo (approved):

PR #1638, opened 2025-08-11
Fix row filter for tables in tile mode

nsteger:
TableTileGridMediator now also handles
TableEvent.TYPE_ROW_FILTER_CHANGED such that tiles are indeed filtered by a programmed row filter.

427908


nsteger:
Backport of https://github.com/eclipse-scout/scout.rt/pull/1630

PR #1657, opened 2025-08-14
CellEditor: complete cell edit if another popup opens

cguglielmo:
Popup._onPopupOpen calls close() but that should not be called for cell editors because it just closes the editor without correctly cancelling or completing the edit.
Instead of extracting the close() call of _onPopupOpen in a method with a name to be defined, close() now just delegates to completeCellEdit().

417214


cguglielmo:
Merged with a separate PR for 26.1 only.

PR #1666, opened 2025-08-18
FocusManager: prevent scrolling when activating focus context

bschwarzent:
When a focus context is activated when a dialog is clicked, the current scroll position should not be changed.

Consider this case:
- A dialog is open with a text area with a long text.
- The user places the cursor at the beginning of the text area.
- The user presses Ctrl-F to invoke the browser-native search bar.
- The user types a text that is in the text area but outside the current view port. The browser highlights the found word and scrolls it into view automatically.
- The user now clicks the text area.
- In some browsers (e.g. Chrome), this causes the text area to jump back to the previous cursor position at the start of the field.

This happens because clicking the text area "activates" the dialog and the focus context:
o Form#_onDialogMouseDown
o FormController#_activateDialog
o Desktop#moveOverlaysBehindAndFocus
o FocusManager#activateFocusContext

It is ok to validate an activate the focus context, but the scroll position should not change in this case (the user can only click elements that are currently within the view port).

395445

### DIFF
--- a/eclipse-scout-core/src/focus/FocusContext.ts
+++ b/eclipse-scout-core/src/focus/FocusContext.ts
@@ -246,8 +246,8 @@ export class FocusContext {
   /**
    * Calls {@link #validateAndSetFocus} with {@link #lastValidFocusedElement}.
    */
-  validateFocus(filter?: () => boolean) {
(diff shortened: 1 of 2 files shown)


bschwarzent:
Run tests

PR #1668, opened 2025-08-19
StringColumn: truncate html content correctly

cguglielmo:
Currently, if a column has a max length and htmlEnabled set to true, the html content may be truncated incorrectly resulting in UI errors. This happens because the Scout JS table just concatenates the table rows together and if the cells contain html tags especially <div> that are not closed, the rows won't be built correctly.

Now, only the text inside the html content will be truncated preserving the html tags.

426350

### DIFF
--- a/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/columns/AbstractStringColumn.java
+++ b/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/columns/AbstractStringColumn.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -12,6 +12,7 @@ package org.eclipse.scout.rt.client.ui.basic.table.columns;
 import org.eclipse.scout.rt.client.extension.ui.basic.table.columns.IStringColumnExtension;
 import org.eclipse.scout.rt.client.ui.basic.table.ITable;
 import org.eclipse.scout.rt.client.ui.basic.table.ITableRow;
+import org.eclipse.scout.rt.client.ui.basic.table.TableUtility;
 import org.eclipse.scout.rt.client.ui.form.fields.IFormField;
 import org.eclipse.scout.rt.client.ui.form.fields.IValueField;
 import org.eclipse.scout.rt.client.ui.form.fields.stringfield.AbstractStringField;
@@ -42,12 +43,11 @@ public abstract class AbstractStringColumn extends AbstractColumn<String> implem
    */
 
   /**
-   * Configures the maximum length of text in this column. This configuration only limits the text length in case of
-   * editable cells.
+   * Configures the maximum length of text in this column. This value is also used by the cell editor if the column is editable.
    * <p>
    * Subclasses can override this method. Default is {@code 4000}.
    *
-   * @return Maximum length of text in an editable cell.
+   * @return Maximum length of text
    */
   @ConfigProperty(ConfigProperty.INTEGER)
   @Order(130)
@@ -185,7 +185,12 @@ public abstract class AbstractStringColumn extends AbstractColumn<String> implem
   protected String/* validValue */ validateValueInternal(ITableRow row, String rawValue) {
     String value = super.validateValueInternal(row, rawValue);
     if (value != null && value.length() > getMaxLength()) {
-      value = value.substring(0, getMaxLength());
+      if (isHtmlEnabled()) {
+        value = TableUtility.getHtmlHelper().truncate(value, getMaxLength(), true);
+      }
+      else {
+        value = value.substring(0, getMaxLength()) + "…";
+      }
     }
     return StringUtility.nullIfEmpty(value);
   }
--- a/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/html/HtmlHelper.java
+++ b/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/html/HtmlHelper.java
@@ -13,6 +13,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.eclipse.scout.rt.platform.ApplicationScoped;
+import org.eclipse.scout.rt.platform.util.ObjectUtility;
 import org.eclipse.scout.rt.platform.util.StringUtility;
 
 /**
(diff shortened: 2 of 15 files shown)


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/columns/AbstractStringColumn.java
```diff
@@ -19,13 +19,17 @@
...
+import org.eclipse.scout.rt.platform.html.HtmlHelper;
+import org.eclipse.scout.rt.platform.util.LazyValue;
 import org.eclipse.scout.rt.platform.util.StringUtility;
 
 /**
  * Column holding Strings
  */
 @ClassId("e564abbc-5f57-4ccc-a50c-003c408df519")
 public abstract class AbstractStringColumn extends AbstractColumn<String> implements IStringColumn {
+  private LazyValue<HtmlHelper> m_htmlHelper = new LazyValue<>(HtmlHelper.class);
```
bschwarzent: Is this really necessary? Needs memory for all columns, and most of them not even have htmlEnabled=true. `BEANS.get()` should be fast enough.

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/html/HtmlHelper.java
```diff
@@ -272,4 +273,147 @@ else if (c == '>') {
...
+      }
+      else {
+        addEllipsis(result, ellipsisPos);
+        ellipsisPos = -1;
+      }
+    }
+    return result.toString();
+  }
+
+  private int updateEllipsisPos(StringBuilder result, int maxLength, boolean addEllipsis, int ellipsisPos, int numChars) {
```
bschwarzent: `protected` instead of `private`?

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/html/HtmlHelper.java
```diff
@@ -272,4 +273,147 @@ else if (c == '>') {
...
+      }
+      char prevChar = i > 1 ? html.charAt(i - 1) : '\0';
+      char nextChar = i < size - 1 ? html.charAt(i + 1) : '\0';
+      if (!inComment && character == '<' && i + 4 < size && "<!--".equals(html.substring(i, i + 4))) {
+        // Comment found <!--
+        inComment = true;
+        result.append(character);
+        continue;
+      }
+      if (inComment && "-->".equals(html.substring(i - 2, i + 1))) {
```
bschwarzent: This will generate an OutOfBoundsException when a comment is opened but never closed. Like in the previous if, we need to check the substring indexes: `i -2 >= 0 && i + 1`

Or create a new utility function substringSafe() that ignores out-of-bounds indexes.

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/html/HtmlHelper.java
```diff
@@ -272,4 +273,147 @@ else if (c == '>') {
...
+      else {
+        addEllipsis(result, ellipsisPos);
+        ellipsisPos = -1;
+      }
+    }
+    return result.toString();
+  }
+
+  private int updateEllipsisPos(StringBuilder result, int maxLength, boolean addEllipsis, int ellipsisPos, int numChars) {
+    if (addEllipsis && ellipsisPos == -1 && numChars == maxLength) {
```
bschwarzent: Why is "ellipsisPos == -1" necessary?

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/html/HtmlHelper.java
```diff
@@ -272,4 +273,147 @@ else if (c == '>') {
...
+        inEntity = true;
+        result.append(character);
+        continue;
+      }
+      if (inEntity && character == ';') {
+        // Closing HTML entity found
+        inEntity = false;
+        numChars++;
+        result.append(character);
+        // Update ellipsis position because an entity counts as character
```
bschwarzent: Comment necessary? This should be obvious, since numChars was just increased.

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/html/HtmlHelper.java
```diff
@@ -272,4 +273,147 @@ else if (c == '>') {
...
+        continue;
+      }
+      if (inTag || inEntity || inComment) {
+        result.append(character);
+        continue;
+      }
+      numChars++;
+      if (numChars <= maxLength) {
+        result.append(character);
+        if (!inTag && !inEntity) {
```
bschwarzent: Check necessary? If inTag or inEntity were true, we would have continued the loop on line 390.

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/html/HtmlHelper.java
```diff
@@ -272,4 +273,147 @@ else if (c == '>') {
...
+   */
+  public String truncate(String html, Integer maxLength, boolean addEllipsis) {
+    if (StringUtility.isNullOrEmpty(html)) {
+      return html;
+    }
+    if (maxLength <= 0) {
+      return addEllipsis ? "…" : "";
+    }
+    int size = html.length();
+    if (size < maxLength) {
```
bschwarzent:
```suggestion
    if (size <= maxLength) {
```

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/html/HtmlHelper.java
```diff
@@ -272,4 +273,147 @@ else if (c == '>') {
...
+      return html;
+    }
+
+    boolean inTag = false;
+    boolean inEntity = false;
+    boolean inAttr = false;
+    boolean inComment = false;
+    char attrChar = '"';
+    int numOpenTags = 0;
+    int numChars = 0;
```
bschwarzent:
```suggestion
    int numTextChars = 0;
```
Avoid confusion with the "chars" we loop over.

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/html/HtmlHelper.java
```diff
@@ -272,4 +273,147 @@ else if (c == '>') {
...
+  }
+
+  private int updateEllipsisPos(StringBuilder result, int maxLength, boolean addEllipsis, int ellipsisPos, int numChars) {
+    if (addEllipsis && ellipsisPos == -1 && numChars == maxLength) {
+      ellipsisPos = result.length();
+    }
+    return ellipsisPos;
+  }
+
+  private void addEllipsis(StringBuilder text, int position) {
```
bschwarzent: Since the name "addEllipsis" is already used for the parameter (which is not involved in this method), maybe "insertEllipsis" would be more appropriate?

```suggestion
  private void insertEllipsis(StringBuilder text, int position) {
```

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/columns/AbstractStringColumn.java
cguglielmo: For 100'000 rows:
beans.get took 28.215500ms
LazyValue took 2.294900ms

I think it is worth it but agree with the memory -> moved Helper to TableUtility.

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/html/HtmlHelper.java
cguglielmo: Agree, but a check for i + 1 is not necessary

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/html/HtmlHelper.java
cguglielmo: I think it can help

bschwarzent: Maybe we could just move the suffix in the comments above:

```
      if (!inComment && !inEntity && !inAttr && character == '&') {
        // HTML entity (character reference) found, e.g. &gt;                <-- suffix removed
        ...
      }
      if (inEntity && character == ';') {
        // Closing HTML entity found -> counts as one character              <-- suffix added
```

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/columns/AbstractStringColumn.java
bschwarzent: The relative difference is big, but the absolute time is still quite small. The overall processing needed for 100k rows is probably much larger. Any how many such tables are there really?

But okay, it's probably better to optimize what we can now. The only problem with cached bean instances is that they cannot be replaced anymore. This is sometimes relevant in tests. In that regard, I'd prefer it if we cached it per table or column instance. Previously, each CompactLineBlock instance called BEANS.get() once, allow at least new instances of the object to obtain the newly registered bean.

cguglielmo: 30ms does not seem big but if there are several methods that take that long it accumulates quickly.
You are right the bean cannot be replaced anymore during runtime, but I don't think this should ever be done, except for maybe profiling or testing purposes. If someone really needs this we can add it later.

PR #1670, opened 2025-08-20
BinaryRefResourceLoader: do not create HttpSession unnecessarily

fschinkel:
If the BinaryRefResourceLoader creates a new HttpSession, there is no possibility the requested object is part of an existing cache. Therefore, the handler tries to load the requested object for which a ClientSession is resolved. As the HttpSession is newly created there will never be a ClientSession associated to it and the handler will return null.
Therefore, do not create a new HttpSession if the request does not already have a session and exit early.

430777


paolobazzi (approved):
Looks good!

PR #1676, opened 2025-08-21
jquery-scout: remove select and selectOne

cguglielmo:
The naming is confusing and inconsistent.

376663


eclipse-scout-core/src/jquery/jquery-scout-types.ts
```diff
@@ -535,15 +535,10 @@ declare global {
      */
     addDeviceClass(): this;
 
-    /**
-     * Adds the class 'selected' to the current element and removes class selected from siblings.
-     */
-    selectOne(): this;
-
     /**
      * Toggles the class 'selected'.
```
bschwarzent:
```suggestion
     * Adds or removes the class 'selected'.
```
It only toggles the class if the given value is different from the current state.

PR #1678, opened 2025-08-22
Add continuous integration GitHub action

cguglielmo:
The action builds and tests the code base by using mvn verify on every pull request and git push.


cguglielmo:
Does not work because it would require an older version of the checkout action, see https://github.com/actions/checkout/issues/1590.

cguglielmo:
run test

cguglielmo:
jenkins run test

PR #1682, opened 2025-08-22
Move user preferences from shared to client

rogersteblerbsi:
As part of the move toward a stateless server session, the
IPreferences, IUserPreferencesService, IUserPreferencesStorageService,
IPreferenceChangeListener and their corresponding implementations
have been moved from shared to client

423389

### DIFF
--- a/org.eclipse.scout.rt.client.test/src/test/java/org/eclipse/scout/rt/client/services/common/prefs/PreferencesTest.java
+++ b/org.eclipse.scout.rt.client.test/src/test/java/org/eclipse/scout/rt/client/services/common/prefs/PreferencesTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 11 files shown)


org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/services/common/prefs/Preferences.java
```diff
@@ -67,7 +67,7 @@ public class Preferences implements IPreferences {
   private final transient FastListenerList<IPreferenceChangeListener> m_eventListeners;
   private boolean m_dirty;
 
-  protected Preferences(String name, ISession userScope) {
+  public Preferences(String name, ISession userScope) {
```
paolobazzi: this still contains reference to ISession which will be removed later on

paolobazzi (commented):
See input about ISession usage in shared

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/services/common/prefs/Preferences.java
rogersteblerbsi: The complete preferences have now been moved from shared to client

paolobazzi (approved):
Looks good. 
Check Scout doc and add section in migration guide.

PR #1683, opened 2025-08-22
HtmlHelper.toPlainText: correctly decode character references

bschwarzent:
Both named and decimal character references should be decoded to their original character when converting HTML to plain text. While the numeric references can be decoded automatically, the list of named references is finite and specified in an RFC. The previously supported list of characters was not complete. This is now fixed.

Additionally, special handling of non-breaking spaces was removed. If they are present in HTML, they should still be present in the plain text as their unicode-equivalent \u00A0, because they are not collapsed when the HTML is rendered.

Similar logic exists in JS in the form of the PlainTextEncoder class. Its behavior was aligned with its Java counterpart. As a consequence, the default value of the 'trim' option was changed from false to true. Test cases were adjusted accordingly. The same test cases from HtmlHelperTest were added to PlainTextEncoderSpec.

413114


eclipse-scout-core/src/encoder/PlainTextEncoder.ts
```diff
@@ -59,22 +41,24 @@ export class PlainTextEncoder {
...
+    text = text.replace(/\r/g, '');
+    text = text.replace(/\n/g, ' ');
+    // Create newlines for certain end tags
+    text = text.replace(/<br\/?><\/div>|<\/div>|<br\/?>|<\/p>|<p\/>|<\/tr>|<\/h[1-6]>|<\/dt>|<\/dd>|<\/dl>|<\/li>|<\/head>/gi, '\n');
 
     // Remove script and style contents
     text = text.replace(/<script\b[^<]*(?:(?!<\/script>)<[^<]*)*<\/script>/gi, '');
     text = text.replace(/<style\b[^<]*(?:(?!<\/style>)<[^<]*)*<\/style>/gi, '');
 
+    // Separate td with ' '
```
nsteger: move this block up right after "// Convert native newlines to whitespace" block analogous to HtmlHelper?

org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/html/HtmlHelperTest.java
```diff
@@ -71,26 +73,43 @@ public void testToPlainText() {
     assertEquals("one & two", helper.toPlainText("<html><head>one & two</html>"));
     assertEquals("one & two", helper.toPlainText("<html><head>one &amp; two</html>"));
     assertEquals("one & two\nthree", helper.toPlainText("<html><head>one &amp; two</head><body>three</html>")); // [?] invalid <body>, has no end tag
+    assertEquals("&amp;", helper.toPlainText("&amp;amp;"));
```
nsteger, line 76: I found another corner case:
`helper.toPlainText("&<span>amp;</span>")`
I would expect
`&amp;`
but the result is
`&`

BUT: the result was the same before this change. Should we ignore this?

bschwarzent: Good find! Unfortunately, we cannot handle this case correctly with the current implementation (tags are removed before entities are converted). Because this is not something we expect to happen in the real world, I suggest we ignore it for now. I added corresponding test cases with a `[?]` marker.

PR #1689, opened 2025-09-01
Always use CustomHttpRequestRetryStrategy

matthiaso:
- Previously not used at all for ScoutApacheConnector, honor retry config properties also here

- Always use strategy regardless how retry config properties are defined

- Add header X-Scout-Final-Response to disapprove automatic retry by status code

429204

### DIFF
--- a/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/ApacheHttpTransportFactory.java
+++ b/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/ApacheHttpTransportFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -15,7 +15,6 @@ import javax.net.ssl.SSLSocketFactory;
 
 import org.apache.hc.client5.http.config.ConnectionConfig;
 import org.apache.hc.client5.http.impl.DefaultClientConnectionReuseStrategy;
-import org.apache.hc.client5.http.impl.DefaultHttpRequestRetryStrategy;
 import org.apache.hc.client5.http.impl.DefaultRedirectStrategy;
 import org.apache.hc.client5.http.impl.classic.HttpClientBuilder;
 import org.apache.hc.client5.http.impl.classic.HttpClients;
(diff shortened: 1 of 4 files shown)


org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/ApacheHttpTransportFactory.java
```diff
@@ -109,17 +108,12 @@ protected void addConnectionKeepAliveSettings(HttpClientBuilder builder) {
...
     final boolean retryOnNoHttpResponseException = CONFIG.getPropertyValue(ApacheHttpTransportRetryOnNoHttpResponseExceptionProperty.class);
     final boolean retryOnSocketExceptionByConnectionReset = CONFIG.getPropertyValue(ApacheHttpTransportRetryOnSocketExceptionByConnectionResetProperty.class);
-    if (retryOnNoHttpResponseException || retryOnSocketExceptionByConnectionReset) {
-      CustomHttpRequestRetryStrategy retryStrategy = new CustomHttpRequestRetryStrategy(1, retryOnNoHttpResponseException, retryOnSocketExceptionByConnectionReset);
-      retryStrategy.enable(builder);
-    }
-    else {
-      builder.setRetryStrategy(new DefaultHttpRequestRetryStrategy());
-    }
+    CustomHttpRequestRetryStrategy retryStrategy = new CustomHttpRequestRetryStrategy(1, retryOnNoHttpResponseException, retryOnSocketExceptionByConnectionReset);
```
rogersteblerbsi, line 115: Just noticed that the `DefaultHttpRequestRetryStrategy` also disables a retry for `NoRouteToHostException` since httpcomponents-client 5.2 (https://github.com/apache/httpcomponents-client/commit/755b69ea3cc44557ed0517b7348ab16870a39975).
Probably we should also add it to our code.

Also there was a small code cleanup in the `DefaultHttpRequestRetryStrategy` in the `retryRequest` method (not the one that you added, but the already existing one). The nesting `else` clause was removed.
(https://github.com/apache/httpcomponents-client/commit/25f080f3079a45939b845af31acb8e7fdd48c360#diff-4992ae8000df8fcafde368e44f9560676e94c0153dd2bc5ef2d2bd1a6a40edd8L181)

matthiaso: Thanks for pointing these changes out; I applied both to out custom strategy (more/less untested). 

At least the second change should not be a change of behavior anyways.

See https://github.com/eclipse-scout/scout.rt/compare/ebef5ffd6cd4ba53220d9a273328de3d232a62d5..b1473c0195aac511631f0037dadd90db031beb82 for changes since your last review.

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/http/retry/CustomHttpRequestRetryStrategy.java
```diff
@@ -136,6 +143,24 @@ public boolean retryRequest(
...
+    }
+
+    return super.retryRequest(response, execCount, context);
+  }
+
+  /**
+   * @return true if {@link #FINAL_RESPONSE_HEADER} is set for response (and thus retry should be skipped regardless of status code)
+   */
+  protected boolean isFinalResponse(HttpResponse response) {
+    return response.getLastHeader(FINAL_RESPONSE_HEADER) != null;
```
rogersteblerbsi: I think it's ok to just check if the header exists, so `FINAL_RESPONSE_HEADER=false` would also result in `true`.
In this case you can also just write:
```suggestion
    return response.containsHeader(FINAL_RESPONSE_HEADER);
```
Maybe you can add a comment about this, that we ignore the value of the header and just check for its existence.

matthiaso: Changed: https://github.com/eclipse-scout/scout.rt/pull/1689#event-19459976491

rogersteblerbsi (approved):

PR #1694, opened 2025-09-03
Remove SequenceNumberDuplicateDetector

matthiaso:
Currently only used for service tunnel to handle certain network issues; if we really consider this an issue we should have implemented a similar technique also for other incoming channels (e.g. REST requests).

423389
425245


matthiaso:
@paolobazzi , @rogersteblerbsi 

What do you think about this pull request? Might be nice for simplification of service-tunnel (move it to REST) and server session removal.

Obviously this is the cheaper option - the other option would be to keep duplicate detection in some way, e.g. by adding yet another header for our "service tunnel" requests in REST client/server filters and store the duplicate detection in some sort of a cache by client session identifier (by user would not be sufficient as users might have multiple sessions for one server). However if multiple backend servers are in use this duplicate detection would be local anyways e.g. we would not be able to recognize if a request is first send to server A and then as a duplicate again to server B.

If we keep the duplicate detection we should consider adding it to all other incoming HTTP interfaces as well (server and client side), we might even consider if it is necessary to have such a detection for external applications accessing interfaces.

On a production environment the last "Duplicate Request" log message I found was logged almost half a year ago.

rogersteblerbsi:
> @paolobazzi , @rogersteblerbsi
> 
> What do you think about this pull request? Might be nice for simplification of service-tunnel (move it to REST) and server session removal.
> 
> Obviously this is the cheaper option - the other option would be to keep duplicate detection in some way, e.g. by adding yet another header for our "service tunnel" requests in REST client/server filters and store the duplicate detection in some sort of a cache by client session identifier (by user would not be sufficient as users might have multiple sessions for one server). However if multiple backend servers are in use this duplicate detection would be local anyways e.g. we would not be able to recognize if a request is first send to server A and then as a duplicate again to server B.
> 
> If we keep the duplicate detection we should consider adding it to all other incoming HTTP interfaces as well (server and client side), we might even consider if it is necessary to have such a detection for external applications accessing interfaces.
> 
> On a production environment the last "Duplicate Request" log message I found was logged almost half a year ago.

Since we have not had any of these messages in almost half a year, and afaik we also don't use this feature in our other internal Scout-based applications, I would say we can safely remove it.

paolobazzi:
+1 for removal of this feature, since it is not available for our REST-based services since years without causing any troubles. Futhermore in a typical kubernetes-based deployment the network infrastructure beween UI and backend pod is completely under the control of the kubernetes cluster and does not contains any unwanted proxies or firewalls which could cause POST requests to be retried.

paolobazzi (approved):

PR #1697, opened 2025-09-03
RestRequestCancellationRegistry: Do not keep threads interrupted

matthiaso:
We interrupt these threads to cancel an action however the threads may be reused by other actions (e.g. Jetty thread pooling); interrupted state should be reset on disposal.

372827
425245


paolobazzi (approved):
looks good

PR #1701, opened 2025-09-05
Reduce RunContext(s) for service-tunnel requests

matthiaso:
Hand through ServiceTunnelOptions to decide about id signature

425245


matthiaso:
Unfortunately change is not as easy as expected; RunContext is also expected in https://github.com/eclipse-scout/scout.rt/blob/976d2b32487008016d2dab969608a8ba35dcab3f/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/servicetunnel/DataObjectWrapperUtility.java#L35-L43 (and content-handler are currently not initalized per request/response) - keep RunContext as is for now.

PR #1711, opened 2025-09-09
Features/cgu/26.1/cell editor 417214

cguglielmo:


bschwarzent (approved):
Small typo in the commit message:

> Unfortunately, if acceptInput of the SmartField is called a second time,
> , a new deferred instance will be created without resolving the previous
> one. This has the effect that the cell editor never closes.

(unnecessary comma at the start of the second line)

PR #1700, opened 2025-09-04
ErrorResponseBuilder: Add headers

matthiaso:
429204

PR #1719, opened 2025-09-11
ErrorResponseBuilder: Add headers

matthiaso:
429204


matthiaso:
https://github.com/eclipse-scout/scout.rt/pull/1700

PR #1732, opened 2025-09-23
IRestResource: Content type text/html with charset UTF-8

lbosshart:
- Added constant in IRestResource for html endpoints to include the default charset UTF-8.
- Security: Including a charset is best practice. Guarantees that clients interpret the payload as UTF-8. Reduces the risk of XSS.

#434062

### DIFF
--- a/org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/IRestResource.java
+++ b/org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/IRestResource.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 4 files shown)


org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/IRestResource.java
```diff
@@ -9,11 +9,20 @@
...
 /**
  * Marks a class as REST resource so that it can be found by {@link RestApplication} using jandex.
  */
 @Bean
 public interface IRestResource {
+
+  /**
+   * When returning responses with content type text/html, it is best practice to include a charset.
+   * Use this constant in annotation {@link Produces} instead of {@link MediaType#TEXT_HTML}.
+   */
```
paolobazzi: Since only a few resources use this constant I'll suggest to move it to a separate interface, e.g.

```
/**
 * Commonly used media types for Scout REST resources.
 *
 * @see MediaType MediaType for a set of builtin default media types.
 */
public interface IRestMediaType {

  /**
   * When returning responses with content type text/html, it is best practice to include a charset.
   * Use this constant in annotation {@link Produces} instead of {@link MediaType#TEXT_HTML}.
   */
  String TEXT_HTML_UTF8 = MediaType.TEXT_HTML + ";charset=UTF-8";
}
````

paolobazzi (approved):
See input for minor improvement.

paolobazzi (approved):

PR #1736, opened 2025-09-24
Use OutputStream provided by StreamingOutput

matthiaso:
Do not use stream of response directly.
Stream must not be closed and jersey does wrap the stream as an UnCloseableOutputStream.

425245


paolobazzi (approved):
looks good, tested locally successful!

PR #1740, opened 2025-09-25
Support for chunked data for REST resources and REST clients

paolobazzi:


org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/chunked/IChunkedDataWriter.java
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * @return IChunkedDataWriter for chunks of type {@code type}, using given {@code delimiter}.
+   */
+  static <T> IChunkedDataWriter<T> create(Class<T> type, String delimiter) {
+    //noinspection unchecked
+    return BEANS.get(IChunkedDataWriter.class).init(type, delimiter);
+  }
+
+  /**
+   * @return Initialized IChunkedDataWriter for chunks of type {@code type}, using given {@code delimiter}.
```
bschwarzent:
```suggestion
   * @return initialized IChunkedDataWriter for chunks of type {@code type}, using given {@code delimiter}.
```

Does this return itself or a different instance?

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/chunked/IChunkedDataWriter.java
```diff
@@ -0,0 +1,60 @@
...
+   */
+  GenericType<T> getResponse();
+
+  /**
+   * Writes given {@code chunk} into the wrapped {@link Response}.
+   */
+  void write(T chunk) throws IOException;
+
+  /**
+   * CLoses the wrapped JAX-RS {@link Response}.
```
bschwarzent:
```suggestion
   * Closes the wrapped JAX-RS {@link Response}.
```

Can this be "auto-closable"?

What happens if the writer is not closed?

The reader has an "isClosed" method - is this not necessary here?

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/client/chunked/IChunkedDataReader.java
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * @return IChunkedDataReader for given {@code response} providing chunks of type {@code type}, using given {@code delimiter}.
+   */
+  static <T> IChunkedDataReader<T> create(Response response, Class<T> type, String delimiter) {
+    //noinspection unchecked
+    return BEANS.get(IChunkedDataReader.class).init(response, type, delimiter);
+  }
+
+  /**
+   * @return Initialized IChunkedDataReader for given {@code response} providing chunks of type {@code type}, using given {@code delimiter}.
```
bschwarzent:
```suggestion
   * @return initialized IChunkedDataReader for given {@code response} providing chunks of type {@code type}, using given {@code delimiter}.
```

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/client/chunked/IChunkedDataReader.java
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * @return next chunk of data or {@code null} if no more data available.
+   */
+  T read();
+
+  /**
+   * @return {@code true} if stream was closed, otherwise {@code false}.
+   */
+  boolean isClosed();
```
bschwarzent, line 80: What if the reader wants to cancel reading, i.e. close the stream early?

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/chunked/IChunkedDataWriter.java
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * @return Initialized IChunkedDataWriter for chunks of type {@code type}, using given {@code delimiter}.
+   */
+  IChunkedDataWriter<T> init(Class<T> type, String delimiter);
+
+  /**
+   * @return wrapped JAX-RS {@link Response}.
+   */
+  GenericType<T> getResponse();
```
bschwarzent: What is the use of this?

org.eclipse.scout.rt.rest.jersey.server/src/main/java/org/eclipse/scout/rt/rest/jersey/server/chunked/ChunkedDataWriter.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Scout REST API wrapper around Jersey {@link ChunkedOutput} instance.
+ * <p>
+ * See documentation
+ * <a href="https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest/user-guide.html#chunked-output">Jersey chunked output</a>
+ */
+public class ChunkedDataWriter<T> implements IChunkedDataWriter<T> {
+
+  protected static final String DEFAULT_DELIMITER = "\n\n";
```
bschwarzent: Is this constant helpful? Why not use the Jersey default? It appears that the delimiter has to be adjusted for each content type anyway. And the actual producer/consumer has to specify the delimiter again (see ChunkedDataTest).

paolobazzi: Removed constant, use Jersey default: org.glassfish.jersey.server.ChunkedOutput#ZERO_LENGTH_DELIMITER

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/chunked/IChunkedDataWriter.java
paolobazzi: Comment updated. Method is void and does not return instance.

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/chunked/IChunkedDataWriter.java
paolobazzi: - Changed to implement Closable interface
- HTTP connection remains open if writer is not closed (until http connection keepalive closes the connection or client read timeout closes the connection), and client is waiting for next chunks
- added isClosed() method

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/client/chunked/IChunkedDataReader.java
paolobazzi: added close method

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/client/chunked/IChunkedDataReader.java
paolobazzi: Changed to void, updated javadoc

org.eclipse.scout.rt.rest.jersey.client/src/main/java/org/eclipse/scout/rt/rest/jersey/client/ChunkedDataReader.java
```diff
@@ -0,0 +1,79 @@
...
+      public Type getRawType() {
+        return ChunkedInput.class;
+      }
+
+      @Override
+      public Type getOwnerType() {
+        return null;
+      }
+    }) {
+    });
```
matthiaso:
```suggestion
    }));
```

org.eclipse.scout.rt.rest.jersey.test/src/main/java/org/eclipse/scout/rt/rest/jersey/fixture/ChunkedDataResource.java
```diff
@@ -0,0 +1,212 @@
...
+  public GenericType<String> getStringsScout() {
+    @SuppressWarnings("resource")
+    IChunkedDataWriter<String> writer = IChunkedDataWriter.create(String.class, "\n\n");
+    Jobs.schedule(() -> {
+      try {
+        for (int i = 1; i <= 3; i++) {
+          writer.write(Integer.toString(i));
+          LOG.info("Wrote chunk '{}'", i);
+          SleepUtil.sleepSafe(10, TimeUnit.MILLISECONDS);
+        }
```
matthiaso:
```suggestion
          assertFalse(writer.isClosed());
        }
```

org.eclipse.scout.rt.rest.jersey.test/src/main/java/org/eclipse/scout/rt/rest/jersey/fixture/ChunkedDataResource.java
```diff
@@ -0,0 +1,212 @@
...
+      }
+      finally {
+        writer.close();
+      }
+    }, Jobs.newInput());
```
matthiaso:
```suggestion
      assertTrue(writer.isClosed);
    }, Jobs.newInput());
```

All methods for this resource seem to be quite similar, actually it would also be possible to generalize these methods (save some lines of code, however probably makes it more difficult to read..), e.g. something like:

```java
  @GET
  @Path("string-scout")
  @Produces(MediaType.TEXT_PLAIN)
  public GenericType<String> getStringsScout() {
    return getScoutInternal(IChunkedDataWriter.create(String.class, "\n\n"), Integer::toString, IChunkedDataWriter::toEntity);
  }

  protected <W extends IChunkedDataWriter<T>, T, R> R getScoutInternal(W writer, IntFunction<T> writeSupplier, Function<W, R> returnFunction) {
    Jobs.schedule(() -> {
      try {
        for (int i = 1; i <= 3; i++) {
          writer.write(writeSupplier.apply(i));
          LOG.info("Wrote chunk '{}'", i);
          SleepUtil.sleepSafe(10, TimeUnit.MILLISECONDS);
          assertFalse(writer.isClosed());
        }
      }
      finally {
        writer.close();
      }
      assertTrue(writer.isClosed());
    }, Jobs.newInput());
    return returnFunction.apply(writer);
  }
```

org.eclipse.scout.rt.rest.jersey.test/src/test/java/org/eclipse/scout/rt/rest/jersey/client/ChunkedDataTest.java
```diff
@@ -0,0 +1,198 @@
...
+    Response response = m_target
+        .path("dataobject-scout")
+        .request().get();
+
+    IChunkedDataReader<FixtureDo> reader = IChunkedDataReader.create(response, FixtureDo.class, "\n\n");
+    assertFalse(reader.isClosed());
+
+    FixtureDo chunk = reader.read();
+    reader.close();
+    assertEquals(BEANS.get(ChunkedDataResource.class).createFixtureDo(1), chunk);
```
matthiaso, line 189: It is probably difficult/impossible to test here that also the server-side (the ChunkedDataResource) stops sending?

matthiaso (approved):

org.eclipse.scout.rt.rest.jersey.test/src/test/java/org/eclipse/scout/rt/rest/jersey/client/ChunkedDataTest.java
paolobazzi: I dont see how to build a testcase for this behavior. Server stops sending means, no more data is written to writer.
- if the writer is not closed by the server, the client will keep on waiting for more chunks until timeout
- if the writer is closed by the server, the client will just receive the already written chunks without any other error

org.eclipse.scout.rt.rest.jersey.server/src/main/java/org/eclipse/scout/rt/rest/jersey/server/chunked/ChunkedDataWriter.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @Override
+  public void init(Class<T> clazz, String delimiter) {
+    assertNull(m_output, "already initialized");
+    assertNotNull(delimiter, "missing delimiter");
+
+    // Use delimiter specified by org.glassfish.jersey.client.ChunkedInput.parser (e.g. "\r\n") if no custom delimiter was specified.
+    // Providing a null or empty delimiter to ChunkedOutput leads to no delimiter at all which cannot be parsed by ChunkedInput later on.
+    // See https://stackoverflow.com/questions/40429196/chunkedoutput-response-from-jersey-rest-services
+    m_output = new ChunkedOutput<>(clazz, StringUtility.isNullOrEmpty(delimiter) ? "\r\n" : delimiter);
```
bschwarzent: Again, the use of StringUtility.isNullOrEmpty() seems superfluous with the previous assertion.

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/client/chunked/IChunkedDataReader.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.scout.rt.rest.client.chunked;
```
bschwarzent, line 10: How important is it that this interface is in the "client" package? If not so much, maybe it would be nicer to put both interfaces in the same "chunked" package?

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/client/chunked/IChunkedDataReader.java
```diff
@@ -0,0 +1,56 @@
...
+   */
+  static <T> IChunkedDataReader<T> create(Response response, Class<T> type, String delimiter) {
+    //noinspection unchecked
+    IChunkedDataReader<T> reader = BEANS.get(IChunkedDataReader.class);
+    reader.init(response, type, delimiter);
+    return reader;
+  }
+
+  /**
+   * Initialize reader for given {@code response} providing chunks of type {@code type}, using given {@code delimiter}.
```
bschwarzent, line 64: Maybe document AssertionError? (also in IChunkedDataWriter)

org.eclipse.scout.rt.rest.jersey.test/src/main/java/org/eclipse/scout/rt/rest/jersey/fixture/ChunkedDataResource.java
paolobazzi: Done. Added writeAsync method, handling writing job. 
Leaved definition of writer and return of response in separate test methods due to readability considerations.

org.eclipse.scout.rt.rest.jersey.client/src/main/java/org/eclipse/scout/rt/rest/jersey/client/ChunkedDataReader.java
```diff
@@ -0,0 +1,78 @@
...
+ * <a href="https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest/user-guide.html#d0e10936">Jersey chunked input</a>
+ */
+public class ChunkedDataReader<T> implements IChunkedDataReader<T> {
+
+  protected ChunkedInput<T> m_input;
+
+  @Override
+  public void init(Response response, Class<T> type, String delimiter) {
+    assertNull(m_input, "already initialized");
+    assertNotNull(delimiter, "missing delimiter");
```
bschwarzent: Are you sure that this assertion is correct? In IChunkedDataReader.create(Response, Class<T>), the delimiter is set to null, which would then trigger the assertion. Also if null was forbidden, an empty string would have to be used to activate the default delimiter, which does not feel right. And the use of StringUtility.isNullOrEmpty() would be unnecessary.

org.eclipse.scout.rt.rest.jersey.server/src/main/java/org/eclipse/scout/rt/rest/jersey/server/chunked/ChunkedDataWriter.java
```diff
@@ -0,0 +1,62 @@
...
+ * <a href="https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest/user-guide.html#chunked-output">Jersey chunked output</a>
+ */
+public class ChunkedDataWriter<T> implements IChunkedDataWriter<T> {
+
+  protected ChunkedOutput<T> m_output;
+
+  @Override
+  public void init(Class<T> clazz, String delimiter) {
+    assertNull(m_output, "already initialized");
+    assertNotNull(delimiter, "missing delimiter");
```
bschwarzent: See comment in ChunkedDataReader

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/client/chunked/IChunkedDataReader.java
paolobazzi: Done

org.eclipse.scout.rt.rest.jersey.client/src/main/java/org/eclipse/scout/rt/rest/jersey/client/ChunkedDataReader.java
paolobazzi: Removed assertion. Null may be used to switch to the default delimiter.
Added test case.

org.eclipse.scout.rt.rest.jersey.server/src/main/java/org/eclipse/scout/rt/rest/jersey/server/chunked/ChunkedDataWriter.java
paolobazzi: Removed assertion, added test case

bschwarzent (approved):

matthiaso (approved):

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/client/chunked/IChunkedDataReader.java
paolobazzi: After discussion with Beat we leave the two packages.

PR #1741, opened 2025-09-26
Allow to check if a file matches a MimeType

mvilliger:
Also adds some new MimeTypes and Magic bytes.

427883


org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/config/AbstractConfigProperty.java
```diff
@@ -119,7 +119,7 @@ protected void fireConfigChangedEvent(ConfigPropertyChangeEvent e) {
     configChangedListeners().list().forEach(listener -> listener.configPropertyChanged(e));
   }
 
-  private static final class P_ParsedPropertyValueEntry<TYPE> {
+  protected static final class P_ParsedPropertyValueEntry<TYPE> {
```
fschinkel, line 122: Why did you change this access modifier?

fschinkel (approved):

PR #1745, opened 2025-09-30
Tree: collapse lazy expanded nodes if all child nodes are removed

fschinkel:
When a node is expanded lazily (i.e. the node is expanded, but only some of the child nodes are visible, marked by "+" instead of "v" icon) and the child nodes are deleted, the parent node must not stay expanded, if there are no child nodes left that can be expanded using the "+" icon.

This was implemented explicitly for pages and was now moved to the tree in general. If the node is fully expanded it stays expanded.

427023

### DIFF
--- a/eclipse-scout-core/src/tree/Tree.ts
+++ b/eclipse-scout-core/src/tree/Tree.ts
@@ -2501,6 +2501,14 @@ export class Tree extends Widget implements TreeModel, Filterable<TreeNode> {
       this._removeNodes(deletedNodes, parentNode || parentNodesToReindex);
     }
 
+    arrays.ensure(parentNode || parentNodesToReindex).forEach(p => {
+      if (p.expandedLazy && !p.childNodes.length) {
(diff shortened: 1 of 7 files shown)


eclipse-scout-core/src/tree/Tree.ts
```diff
@@ -2501,6 +2501,14 @@ export class Tree extends Widget implements TreeModel, Filterable<TreeNode> {
       this._removeNodes(deletedNodes, parentNode || parentNodesToReindex);
     }
 
+    arrays.ensure(parentNode || parentNodesToReindex).forEach(p => {
+      if (p?.expandedLazy && !p.childNodes?.length) {
```
bschwarzent: Why are `?` necessary? `parentNodesToReindex` is always an array and only contains valid nodes. `parentNode` might be null or undefined, but this is removed automatically by arrays.ensure(). `TreeNode#childNodes` is also never null.

PR #1746, opened 2025-10-01
TreeAdapter: mark remote nodes

fschinkel:
Tree node events sent to the UI server for nodes that do not exist on the UI server create a lot of noise in the log. This may be acceptable if Scout JS expects the UI server to know a tree node but should be avoided when Scout JS knows for sure that a tree node does not exist on the UI server.
Therefore, all tree nodes created by the TreeAdapter are now marked as remote. This flag is used to filter events that are sent back to the UI server. In addition, the filter method introduces hybrid nodes which are nodes created as a wrapper in Java while all their logic is implemented in Scout JS. The only known variant of these hybrid nodes are JsPages and these are marked accordingly by the OutlineAdapter.

427023

### DIFF
--- a/eclipse-scout-core/src/desktop/outline/OutlineAdapter.ts
+++ b/eclipse-scout-core/src/desktop/outline/OutlineAdapter.ts
@@ -306,7 +306,8 @@ export class OutlineAdapter extends TreeAdapter {
       objectType: pageModel.jsPageObjectType,
       pageParam: pageParam,
       classId: pageModel.classId,
-      modelClass: pageModel.modelClass
+      modelClass: pageModel.modelClass,
+      __hybrid: true
     };
   }
 
@@ -323,9 +324,6 @@ export class OutlineAdapter extends TreeAdapter {
 
   protected override _initNodeModel(nodeModel?: TreeNodeModel): ChildModelOf<Page> {
     const model = super._initNodeModel(nodeModel) as ChildModelOf<Page>;
-    // This marker is only set for pages that represent a remote page on the UI server. It prevents menus from being inherited
-    // from the parent table page, because in the case of Java pages that is already done on the server.
-    model.remote = true;
     model.pageParam = dataObjects.deserialize(model.pageParam);
     return model;
   }
--- a/eclipse-scout-core/src/tree/TreeAdapter.ts
+++ b/eclipse-scout-core/src/tree/TreeAdapter.ts
@@ -56,23 +56,34 @@ export class TreeAdapter extends ModelAdapter {
   }
 
   protected _onWidgetNodeClick(event: TreeNodeClickEvent) {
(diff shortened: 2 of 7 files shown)


eclipse-scout-core/src/tree/TreeAdapter.ts
```diff
@@ -319,6 +335,18 @@ export class TreeAdapter extends ModelAdapter {
...
+      return true;
+    }
+    return includeHybrid ? !!node.__hybrid : false;
+  }
 }
 
 App.addListener('bootstrap', TreeAdapter.modifyTreePrototype);
+
+export interface AdapterTreeNode extends TreeNode {
+  remote?: boolean;
```
bschwarzent, line 352: I think we already discussed this, but I forgot why this property does not use the double underscore prefix.

eclipse-scout-core/src/tree/TreeAdapter.ts
```diff
@@ -319,6 +335,18 @@ export class TreeAdapter extends ModelAdapter {
     objects.replacePrototypeFunction(Tree, '_createTreeNode', TreeAdapter._createTreeNodeRemote, true);
     objects.replacePrototypeFunction(Tree, '_updateMarkChildrenChecked', TreeAdapter._updateMarkChildrenCheckedRemote, true);
   }
+
+  static isRemote(node: AdapterTreeNode, includeHybrid?: boolean): boolean {
```
bschwarzent: All usages seem to pass `true` as second argument, maybe set that as the default value? If the intention was to force the caller to think about whether they want hybrid nodes to be treated as remote nodes, they argument should not be optional.

eclipse-scout-core/src/desktop/outline/OutlineAdapter.ts
```diff
@@ -323,9 +324,6 @@ export class OutlineAdapter extends TreeAdapter {
 
   protected override _initNodeModel(nodeModel?: TreeNodeModel): ChildModelOf<Page> {
     const model = super._initNodeModel(nodeModel) as ChildModelOf<Page>;
-    // This marker is only set for pages that represent a remote page on the UI server. It prevents menus from being inherited
```
bschwarzent, line 326: Preserve this comment as jsdoc on AdapterTreeNode?

bschwarzent (commented):
Commit message:

> This may be acceptable if Scout JS can not know if the UI server knows a tree node or not but it is not ok if Scout JS can determine if a tree node does not exist on the UI server.

I don't quite understand this sentence. What is it trying to say?

eclipse-scout-core/src/tree/TreeAdapter.ts
fschinkel: The property exists already on the `Page` and was moved to the `TreeNode`. Therefore, it was not renamed to minimize code migration.

PR #1766, opened 2025-10-22
Reset Interrupt Status

work-lto:
Ensures that any interrupt flags set during request handling don’t leak back into Jetty’s thread pool.

372827

### DIFF
--- a/org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/filter/ExceptionFilter.java
+++ b/org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/filter/ExceptionFilter.java
@@ -62,6 +62,18 @@ public class ExceptionFilter implements Filter {
       }
       throw new JettyQuietExceptionWrapper(e);
     }
+    finally {
+      if (Thread.interrupted()) {
+        HttpServletRequest req = (HttpServletRequest) request;
+        MDC.put(CorrelationIdContextValueProvider.KEY, req.getHeader(CorrelationId.HTTP_HEADER_NAME));
+        try {
+          LOG.debug("Reset interrupted state");
+        }
+        finally {
+          MDC.remove(CorrelationIdContextValueProvider.KEY);
+        }
+      }
+    }
   }
 
   /**
@@ -69,10 +81,6 @@ public class ExceptionFilter implements Filter {
    */
   protected boolean isCausedByQuietException(Throwable t) {
     return BEANS.get(DefaultExceptionTranslator.class).throwableCausesAccept(t, e -> e instanceof QuietException);
-  }
-
-  @Override
-  public void destroy() {
   }
 
   /**
--- a/org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/filter/ExceptionFilter.java
+++ b/org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/filter/ExceptionFilter.java
@@ -53,26 +53,25 @@ public class ExceptionFilter implements Filter {
       }
 
       HttpServletRequest req = (HttpServletRequest) request;
-      MDC.put(CorrelationIdContextValueProvider.KEY, req.getHeader(CorrelationId.HTTP_HEADER_NAME));
-      try {
-        LOG.warn(req.getRequestURI(), e);
-      }
-      finally {
-        MDC.remove(CorrelationIdContextValueProvider.KEY);
-      }
+      runWithCorrelationId(req.getHeader(CorrelationId.HTTP_HEADER_NAME), () -> LOG.warn(req.getRequestURI(), e));
+
       throw new JettyQuietExceptionWrapper(e);
     }
     finally {
       if (Thread.interrupted()) {
         HttpServletRequest req = (HttpServletRequest) request;
-        MDC.put(CorrelationIdContextValueProvider.KEY, req.getHeader(CorrelationId.HTTP_HEADER_NAME));
-        try {
-          LOG.debug("Reset interrupted state");
-        }
-        finally {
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/filter/ExceptionFilter.java
```diff
@@ -62,6 +62,18 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
+    finally {
+      if (Thread.interrupted()) {
+        HttpServletRequest req = (HttpServletRequest) request;
+        MDC.put(CorrelationIdContextValueProvider.KEY, req.getHeader(CorrelationId.HTTP_HEADER_NAME));
+        try {
+          LOG.debug("Reset interrupted state");
+        }
+        finally {
+          MDC.remove(CorrelationIdContextValueProvider.KEY);
+        }
```
matthiaso: Same pattern seems to be used also in catch block; if we use this pattern more often we could think of extracting it somewhere? Maybe:


```
Index: org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/context/CorrelationIdContextValueProvider.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/context/CorrelationIdContextValueProvider.java b/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/context/CorrelationIdContextValueProvider.java
--- a/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/context/CorrelationIdContextValueProvider.java	(revision 20e595a030b6f512d5223cf493f9550699fe1088)
+++ b/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/context/CorrelationIdContextValueProvider.java	(date 1761212821310)
@@ -12,6 +12,7 @@
 import org.eclipse.scout.rt.platform.ApplicationScoped;
 import org.eclipse.scout.rt.platform.logger.DiagnosticContextValueProcessor;
 import org.eclipse.scout.rt.platform.logger.DiagnosticContextValueProcessor.IDiagnosticContextValueProvider;
+import org.eclipse.scout.rt.platform.util.concurrent.IRunnable;
 import org.slf4j.MDC;
 
 /**
@@ -36,4 +37,14 @@
   public String value() {
     return CorrelationId.CURRENT.get();
   }
+
+  public static void run(String correlationId, Runnable runnable) {
+    MDC.put(CorrelationIdContextValueProvider.KEY, correlationId);
+    try {
+      runnable.run();
+    }
+    finally {
+      MDC.remove(CorrelationIdContextValueProvider.KEY);
+    }
+  }
 }
```

Then this method would be shorter and we could just write:

```
CorrelationIdContextValueProvider.run(req.getHeader(CorrelationId.HTTP_HEADER_NAME), () -> LOG.warn(req.getRequestURI(), e));
```

Also this ensures that these calls are always wrapped with a try-finally (which they are already in this method correctly).

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/filter/ExceptionFilter.java
```diff
@@ -62,6 +62,18 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
       }
       throw new JettyQuietExceptionWrapper(e);
     }
+    finally {
+      if (Thread.interrupted()) {
+        HttpServletRequest req = (HttpServletRequest) request;
+        MDC.put(CorrelationIdContextValueProvider.KEY, req.getHeader(CorrelationId.HTTP_HEADER_NAME));
+        try {
+          LOG.debug("Reset interrupted state");
```
matthiaso: Should we maybe add more information here from the `HttpServletRequest` - maybe the URI? Do we have a username here?

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/context/RunMonitorCancelRegistry.java
```diff
@@ -97,6 +100,10 @@ public IRegistrationHandle register(final RunMonitor runMonitor, final String se
     return () -> {
       synchronized (m_registryLock) {
         m_registry.remove(entry);
+        if (entry.getRunMonitor().isCancelled() && Thread.interrupted()) {
+          // as thread may be used by other operations as well; interrupted state must be reset after previous interruption
+          LOG.trace("Reset interrupted state for cancelled and interrupted run monitor - requestId:{}", requestId);
```
matthiaso: Maybe log `sessionId` here as well (which is available as a variable)?

matthiaso (approved):

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/filter/ExceptionFilter.java
work-lto: Makes sense, i'll check the request object and see what other useful information i could log

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/filter/ExceptionFilter.java
work-lto: Even though it is much nicer and more concise i don't think it's a good idea to put it into a global method like this. In general, the correlation id should already be in the MDC when logging and i don't want to encourage users to use this new API. However, i can put this method in the ExceptionFilter class

matthiaso (approved):

PR #1768, opened 2025-10-23
Add child pages to IJsPage

fschinkel:
An IJsPage is a page created in Java as a wrapper while all its logic is
implemented in Scout JS. As this page lives in a Scout Classic Outline
all its child pages need to be known to the Outline in Java as well.
Therefore, the implementation in Scout JS can use the new JsPageHelper
and call the server to create its child pages for a list of IIds or
IPageParamDos. The creation of the child pages is then delegated to the
IJsPage wrapper in Java. Afterward, the JsPageHelper holds references to
the created child pages that can be used by the Scout JS page again.
The created child pages will be identified by an IPageParamDo and they
are inserted into the Outline directly on the server. The JsPageHelper
ensures that necessary information is synchronized, like e.g. summary
information from a parent table page.
The TreeAdapter may now send nodesChanged events back to the server in
order to update text, htmlEnabled, cssClass and iconId if it was changed
e.g. by a parent PageWithTable.
If there are childNodes present on the server for an IJsPage these
childNodes and the properties expanded and expandedLazy are added to the
jsPageModel. This ensures the outline state being correct even after a
browser reload.

427023

### DIFF
--- a/eclipse-scout-core/src/Extension.ts
+++ b/eclipse-scout-core/src/Extension.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -40,12 +40,21 @@ export class Extension<E> {
   next: (...args: any[]) => any;
 
   extend(extended: E, funcName: string) {
-    let origFunc = extended[funcName];
-    let extension = this;
+    const funcOrig = extended[funcName];
+    const extension = this;
     extended[funcName] = function(...args) {
+      const extendedOrig = extension.extended;
+      const nextOrig = extension.next;
+
       extension.extended = this;
-      extension.next = origFunc.bind(this);
-      return extension[funcName](...args);
+      extension.next = funcOrig.bind(this);
+
+      try {
+        return extension[funcName](...args);
+      } finally {
+        extension.extended = extendedOrig;
+        extension.next = nextOrig;
+      }
     };
   }
 
--- a/eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.less
+++ b/eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.less
@@ -1,0 +1,17 @@
+/*
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+// hide icon and text if child page of a js page is loading to prevent flickering (e.g. the child pages text will change from the text given in its model to the summary of its parent table page)
+.tree-node.js-page-child-page-loading {
+  & > .icon,
+  & > .text {
+    .invisible();
+  }
+}
--- a/eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
+++ b/eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
@@ -1,0 +1,468 @@
+/*
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+import {
+  arrays, dataObjects, HybridActionContextElement, HybridActionContextElements, HybridManager, InitModelOf, LoadChildPagesHybridActionDo, objects, ObjectWithType, Outline, OutlineAdapter, Page, PageParamDo, PageWithTable, scout,
+  SomeRequired, Table, TableRowsInsertedEvent, TreeAllChildNodesDeletedEvent, TreeNodesDeletedEvent, TreeNodesInsertedEvent, TreeNodesUpdatedEvent, typeName
+} from '../../../../index';
+
+/**
+ * This helper can be used to create child pages for a JsPage on the UI server.
+ * The child pages are created using an id or a {@link PageParamDo}.
+ * The child pages are created by the Java equivalent of the JsPage (see AbstractJsPage.createChildPage) when {@link #callLoadChildPages} is called.
+ * After the returned {@link Promise} is resolved the child pages are available by {@link #findChildPage}.
+ *
+ * To create child pages for a {@link PageWithTable}:
+ * - Call and await {@link #callLoadChildPages} in {@link PageWithTable#_loadTableData} using ids or {@link PageParamDo}s created from the loaded data.
+ * - Implement {@link PageWithTable#_createChildPage} and return the result of {@link #findChildPage} using an id or {@link PageParamDo} created for the given {@link TableRow}.
+ *
+ * To create child pages for a {@link PageWithNodes}:
+ * - Implement {@link PageWithNodes#_createChildPages} and call {@link #callLoadChildPages} using ids or {@link PageParamDo}s.
+ * - After the {@link Promise} is resolved use the same ids or {@link PageParamDo}s to call {@link #findChildPages} and return the resulting child pages.
+ */
+export class JsPageHelper implements JsPageHelperModel, ObjectWithType {
+
+  declare model: JsPageHelperModel;
+  declare initModel: SomeRequired<this['model'], 'page'>;
+
+  objectType: string;
+
+  page: Page;
+
+  protected _replaceChildPages = true;
+  /**
+   * Child pages by id (see {@link #_createChildPageId}).
+   */
+  protected _childPagesById = new Map<string, JsPageChildPage>();
+
+  // general listeners
+
+  protected _nodesUpdatedHandler = this._onNodesUpdated.bind(this);
+  protected _nodesDeletedHandler = this._onNodesDeleted.bind(this);
+  protected _allChildNodesDeletedHandler = this._onAllChildNodesDeleted.bind(this);
+
+  // listeners for PageWithTable
+
+  protected _nodesInsertedHandler = this._onNodesInserted.bind(this);
+  protected _tableRowsInsertHandler = this._onTableRowsInserted.bind(this);
+
+  init(model: InitModelOf<this>) {
+    $.extend(this, model);
+
+    scout.assertInstance(this.page, Page, 'Page not set or has wrong type');
+
+    // do not replace child pages initially
+    // if e.g. the browser is reloaded the previously created child pages already exist on the UI server and therefore childNodes are set in the pages initModel
+    // if the page is a PageWithTable its table will be loaded when the page is selected and this load must not lead to all child pages being replaced
+    this._replaceChildPages = !this.page.childNodes?.length;
+
+    this._installOutlineListeners();
+    this._installPageWithTableListeners();
+  }
+
+  public destroy() {
+    this._uninstallOutlineListeners();
+    this._uninstallPageWithTableListeners();
+    this.page = null;
+  }
+
+  protected _installOutlineListeners() {
+    this.outline.on('nodesUpdated', this._nodesUpdatedHandler);
+    this.outline.on('nodesDeleted', this._nodesDeletedHandler);
+    this.outline.on('allChildNodesDeleted', this._allChildNodesDeletedHandler);
+  }
+
+  protected _uninstallOutlineListeners() {
+    this.outline.off('nodesUpdated', this._nodesUpdatedHandler);
+    this.outline.off('nodesDeleted', this._nodesDeletedHandler);
+    this.outline.off('allChildNodesDeleted', this._allChildNodesDeletedHandler);
+  }
+
+  protected _onNodesUpdated(event: TreeNodesUpdatedEvent) {
+    if (!event.nodes?.length) {
+      return;
+    }
+
+    // filter child pages of page
+    const childPages = event.nodes.filter(node => node.parentNode === this.page) as Page[];
+    if (!childPages.length) {
+      return;
+    }
+
+    // Child pages may have been updated with e.g. a new text. Send a nodesChanged event in order to send this information to the UI server.
+    // This information is needed when the browser is reloaded as the child pages will still be available.
+    (this.outline.modelAdapter as OutlineAdapter).sendNodesChanged(childPages);
+  }
+
+  protected _onNodesDeleted(event: TreeNodesDeletedEvent) {
+    if (!event.nodes?.length) {
+      return;
+    }
+
+    // mark page as childrenLoaded=false if all child pages are deleted
+    if (!this.page.childNodes.length) {
+      this.page.childrenLoaded = false;
+    }
+
+    // clean up _childPagesById map
+    const pages = event.nodes as Page[];
+    this._removeChildPagesFromIdMap(pages);
+
+    if (!this.page.detailTable) {
+      return;
+    }
+
+    // remove corresponding rows, otherwise e.g. double-clicking a row will lead to errors as the page was deleted
+    const rows = pages.map(page => page.row).filter(Boolean);
+    this.page.detailTable.deleteRows(rows);
+  }
+
+  protected _onAllChildNodesDeleted(event: TreeAllChildNodesDeletedEvent) {
+    if (event.parentNode !== this.page) {
+      return;
+    }
+
+    // always mark page as childrenLoaded=false as all child pages are deleted
+    this.page.childrenLoaded = false;
+
+    // clean up _childPagesById map
+    this._childPagesById.clear();
+
+    if (!this.page.detailTable) {
+      return;
+    }
+
+    // remove all rows, otherwise e.g. double-clicking a row will lead to errors as the page was deleted
+    this.page.detailTable.deleteAllRows();
+  }
+
+  protected _installPageWithTableListeners() {
+    if (!(this.page instanceof PageWithTable)) {
+      return;
+    }
+
+    this.outline.on('nodesInserted', this._nodesInsertedHandler);
+
+    // detailTable may change -> install for current and listen on propertyChange
+    this._installPageWithTableDetailTableListeners(this.page?.detailTable);
+    this.page.on('propertyChange:detailTable', e => {
+      this._uninstallPageWithTableDetailTableListeners(e.oldValue);
+      this._installPageWithTableDetailTableListeners(e.newValue);
+    });
+  }
+
+  protected _uninstallPageWithTableListeners() {
+    if (!(this.page instanceof PageWithTable)) {
+      return;
+    }
+
+    this.outline.off('nodesInserted', this._nodesInsertedHandler);
+    this._uninstallPageWithTableDetailTableListeners(this.page?.detailTable);
+  }
+
+  protected _onNodesInserted(event: TreeNodesInsertedEvent) {
+    if (event.parentNode !== this.page) {
+      return;
+    }
+
+    // The event is triggered twice during the load/reload of a PageWithTable that creates its child pages using the JsPageHelper.
+    // 1. When the child pages are inserted on the UI server it is called for the first time.
+    //    In this case e.g. the text may not be correct as the page was not updated from the table row yet.
+    //    If this happens, the PageWithTable is marked with childrenLoaded=false.
+    // 2. The second time this event is triggered is when the PageWithTable replaces its rows.
+    //    Now the child page is ready to be inserted as it was linked to the row and updated from it.
+    //    As the table load/reload is now completed, the PageWithTable is marked with childrenLoaded=true.
+
+    // Mark all child pages loading iff the PageWithTable has not loaded its children.
+    for (const childPage of event.nodes) {
+      childPage.toggleCssClass('js-page-child-page-loading', !this.page.childrenLoaded);
+      childPage._decorate();
+    }
+  }
+
+  protected _installPageWithTableDetailTableListeners(detailTable: Table) {
+    if (!detailTable) {
+      return;
+    }
+    detailTable.on('rowsInserted', this._tableRowsInsertHandler);
+  }
+
+  protected _uninstallPageWithTableDetailTableListeners(detailTable: Table) {
+    if (!detailTable) {
+      return;
+    }
+    detailTable.off('rowsInserted', this._tableRowsInsertHandler);
+  }
+
+  protected _onTableRowsInserted(event: TableRowsInsertedEvent) {
+    // When rows are inserted in a PageWithTable, all child pages are ready, i.e. they are linked to their row and updated from it.
+    // Send a nodesChanged event in order to send summary information from the PageWithTable to the UI server.
+    // This information is needed when the browser is reloaded as the child pages will still be available but the table data won't.
+    const pages = event.rows.map(row => row.page).filter(Boolean);
+    (this.outline.modelAdapter as OutlineAdapter).sendNodesChanged(pages);
+  }
+
+  get outline(): Outline {
+    return this.page.outline;
+  }
+
+  /**
+   * Calls loadChildPages on the UI server for the given ids or {@link PageParamDo}s. The pages are created by the Java equivalent of the JsPage (see AbstractJsPage.createChildPage).
+   * The returned {@link Promise} is resolved when all child pages are created on the UI server.
+   *
+   * @param idsOrPageParams ids or {@link PageParamDo}s that are used to create child pages on the UI server.
+   * @param replace Whether existing child pages shall be replaced or reused. If not specified, the child pages will always be replaced except for the first call in order to keep existing child pages after a browser reload.
+   */
+  async callLoadChildPages(idsOrPageParams: (string | PageParamDo)[], replace?: boolean): Promise<void> {
+    // Ensure pageParams and collect them distinctly.
+    // The given list may contain duplicate pageParams.
+    // Sending these duplicates to the server would result in pages added multiple times to the outline which leads to unexpected behaviour.
+    const childPageIds = new Set<string>();
+    const pageParams: PageParamDo[] = [];
+    arrays.ensure(idsOrPageParams)
+      .map(idOrPageParam => this._createChildPageParam(idOrPageParam))
+      .filter(Boolean)
+      .forEach(pageParam => {
+        // The pageParam is a PageParamDo and a Set only checks "===" and not ".equals()".
+        // Therefore, create the child page id to ensure that duplicate page params are only used once.
+        const childPageId = this._createChildPageId(pageParam);
+        if (childPageIds.has(childPageId)) {
+          return;
+        }
+        childPageIds.add(childPageId);
+
+        pageParams.push(pageParam);
+      });
+    // ensure replace flag
+    replace = scout.nvl(replace, this._replaceChildPages);
+
+    // call load child pages
+    const hybridManager = await HybridManager.get(this.page.session, true);
+    await hybridManager.callActionAndWaitWithContext('scout.LoadChildPages',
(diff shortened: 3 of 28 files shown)


org.eclipse.scout.rt.ui.html.selenium/src/main/java/org/eclipse/scout/rt/ui/html/selenium/util/SeleniumUtil.java
```diff
@@ -193,6 +194,14 @@ public static By byModelClassAttributeContainsNot(Class<?> modelClass, String at
         modelClass.getName(), attribute, value));
   }
 
+  public static By byDataUuid(String uuid) {
```
bschwarzent: I would remove the word "data" from the method name (the fact that this uses a data attribute to identify the element is irrelevant for the caller). Similar to `byModelClass`.

eclipse-scout-core/src/Extension.ts
```diff
@@ -43,9 +43,18 @@ export class Extension<E> {
     let origFunc = extended[funcName];
     let extension = this;
```
bschwarzent: Proposal for an improved commit message:

```
Extensions are created only once but are executed for multiple instances
of the extended class. To access the extended instance or function, the
properties 'extended' and 'next' are set before the extension is called.
Since the extended function may be called recursively for multiple
targets, these properties must be reset after each execution of the
extension.

Consider an extension for TreeNode._init that calls itself for
all child nodes.

protected _init(model: InitModelOf<TreeNode>) {
  this.next(model);
  this.extended.foo = 42;
}

If the properties are not reset to their previous values after the
extension call, they will retain the values of the last call. If the
tree node for which this extension was initially called has child nodes,
the 'extended' property will not point to the expected original tree
node but to its last child node, as the extension has been called for
all child nodes during "this.next(model)".
```

eclipse-scout-core/src/Extension.ts
```diff
@@ -43,9 +43,18 @@ export class Extension<E> {
...
     extended[funcName] = function(...args) {
+      const extendedOrig = extension.extended;
```
bschwarzent, line 46: Consistency "let" vs. "const" (either use const as well for the two existing variables above or use let everywhere)

eclipse-scout-core/src/Extension.ts
```diff
@@ -43,9 +43,18 @@ export class Extension<E> {
...
+      const nextOrig = extension.next;
```
bschwarzent, line 47: Potential confusion with the word "orig" ("origFunc" vs. "nextOrig"), also inconsistent naming

eclipse-scout-core/src/Extension.ts
```diff
@@ -43,9 +43,18 @@ export class Extension<E> {
...
+
       extension.extended = this;
       extension.next = origFunc.bind(this);
-      return extension[funcName](...args);
+
+      const result = extension[funcName](...args);
```
bschwarzent: Variable "result" could be made obsolete with the use of a try-finally block

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.less
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
```
bschwarzent, line 2: Proposal for an improved commit message:

```
An IJsPage is a page created in Java whose logic is fully implemented in
Scout JS. Since this page resides within a Scout Classic outline,
all its child pages must also be known to the Java outline.

To do so, the implementation in Scout JS can use the new JsPageHelper
to create child pages for a list of IIds or IPageParamDos. The actual
creation of the child pages is delegated to the IJsPage wrapper in Java.
Afterward, the JsPageHelper holds references to the created child pages,
allowing the Scout JS page to interact with them.

The created child pages are identified by an IPageParamDo and are inserted
directly into the server-side outline. The JsPageHelper ensures that
necessary information (such as summary information from a parent table
page) is properly synchronized. The TreeAdapter can send a new
'nodesChanged' event back to the server in order to update text,
htmlEnabled, cssClass and iconId if it was changed e.g. by a parent
PageWithTable.

If there are childNodes present on the server for an IJsPage, these
childNodes and the properties 'expanded' and 'expandedLazy' are added to
the jsPageModel. This ensures that the outline state remains consistent
even after a browser reload.
```

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.less
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+.outline.tree > .tree-data {
```
bschwarzent: Remove unnecessary outer rules (`invisible` mixin specifies !important, so it will be applied even if the remaining rule was too unspecific)

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -0,0 +1,470 @@
...
+
+import {
+  arrays, dataObjects, HybridActionContextElement, HybridActionContextElements, HybridManager, InitModelOf, LoadChildPagesHybridActionDo, objects, ObjectWithType, OutlineAdapter, Page, PageParamDo, PageWithTable, scout, SomeRequired,
+  styles, Table, TableRowsInsertedEvent, TreeAllChildNodesDeletedEvent, TreeNodesDeletedEvent, TreeNodesInsertedEvent, TreeNodesUpdatedEvent, typeName
+} from '../../../../index';
+
+/**
+ * This helper can be used to create child pages for a JsPage on the UI server.
+ * The child pages are created using an id or a {@link PageParamDo}.
+ * The child pages are created by the Java pendant of the JsPage (see org.eclipse.scout.rt.client.ui.desktop.outline.pages.js.AbstractJsPage.createChildPage) when {@link #callLoadChildPages} is called.
```
bschwarzent: pendant -> "counterpart" or "equivalent"
(a pendant is a necklace)

Also: it is unnecessary to specify the entire FQCN, I think `AbstractJsPage.createChildPage` is enough.

Adjust the JsDoc of callLoadChildPages similarly.

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -0,0 +1,470 @@
...
+ * After the returned {@link Promise} is resolved the child pages are available by {@link #findChildPage}.
+ *
+ * To create child pages for a {@link PageWithTable}:
+ * <ul>
```
bschwarzent: Don't use HTML to format jsdoc

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -0,0 +1,470 @@
...
+    $.extend(this, model);
+
+    scout.assertInstance(this.page, Page, 'Page not set or has wrong type');
+
+    // do not replace child pages initially
+    // if e.g. the browser is reloaded the previously created child pages already exist on the UI server and therefore childNodes are set in the pages initModel
+    // if the page is a PageWithTable its table will be loaded when the page is selected and this load must not lead to all child pages being replaced
+    this._replaceChildPages = !this.page.childNodes?.length;
+
+    this._installListeners();
```
bschwarzent: Maybe call this "_installOutlineListeners"?

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -0,0 +1,470 @@
...
+  }
+
+  public destroy() {
+    this._uninstallListeners();
+    this._uninstallPageWithTableListeners();
+    this.page = null;
+  }
+
+  protected _installListeners() {
+    this.page.outline.on('nodesUpdated', this._nodesUpdatedHandler);
```
bschwarzent: Consider a "this.outline" getter as an alias for "this.page.outline"

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -0,0 +1,470 @@
...
+      return;
+    }
+
+    // mark page as childrenLoaded=false if all child pages are deleted
+    if (!this.page.childNodes.length) {
+      this.page.childrenLoaded = false;
+    }
+
+    if (!this._childPagesById.size) {
+      return;
```
bschwarzent: Is it guaranteed that the detailTable does not have any rows if this map is empty?

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -0,0 +1,470 @@
...
+
+  protected _onNodesInserted(event: TreeNodesInsertedEvent) {
+    if (event.parentNode !== this.page) {
+      return;
+    }
+
+    // The event is triggered twice during the load/reload of a PageWithTable that creates its child pages using the JsPageHelper.
+    // 1. When the child pages are inserted on the UI server it is called for the first time.
+    //    In this case e.g. the text may not be correct as the page was not updated from the table row yet.
+    //    If this happens The PageWithTable is marked with childrenLoaded=false.
```
bschwarzent:
```suggestion
    //    If this happens, the PageWithTable is marked with childrenLoaded=false.
```

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -0,0 +1,470 @@
...
+    // 2. The second time this event is triggered is when the PageWithTable replaces its rows.
+    //    Now the child page is ready to be inserted as it was linked to the row and updated from it.
+    //    As the table load/reload is now completed the PageWithTable is marked with childrenLoaded=true.
```
bschwarzent:
```suggestion
    //    As the table load/reload is now completed, the PageWithTable is marked with childrenLoaded=true.
```

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -0,0 +1,470 @@
...
+
+    // Mark all child pages loading iff the PageWithTable has not loaded its children.
+    for (const childPage of event.nodes) {
+      childPage.setCssClass(styles.toggleCssClass(childPage.cssClass, 'js-page-child-page-loading', !this.page.childrenLoaded));
+      childPage._decorate();
+    }
+  }
+
+  protected _installPageWithTableTableListeners(table: Table) {
```
bschwarzent: Consider using the name "detailTable" instead of just "table" to make it easier to spot (because PageWithTable also includes the word "table").

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -0,0 +1,470 @@
...
+    // 1. When the child pages are inserted on the UI server it is called for the first time.
+    //    In this case e.g. the text may not be correct as the page was not updated from the table row yet.
+    //    If this happens The PageWithTable is marked with childrenLoaded=false.
+    // 2. The second time this event is triggered is when the PageWithTable replaces its rows.
+    //    Now the child page is ready to be inserted as it was linked to the row and updated from it.
+    //    As the table load/reload is now completed the PageWithTable is marked with childrenLoaded=true.
+
+    // Mark all child pages loading iff the PageWithTable has not loaded its children.
+    for (const childPage of event.nodes) {
+      childPage.setCssClass(styles.toggleCssClass(childPage.cssClass, 'js-page-child-page-loading', !this.page.childrenLoaded));
```
bschwarzent: Consider adding convenience methods addCssClass/removeCssClass/toggleCssClass to the Page class. This was already done on Cell, which is also not a Widget.

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -0,0 +1,470 @@
...
+  protected _uninstallPageWithTableTableListeners(table: Table) {
+    if (!table) {
+      return;
+    }
+    table.off('rowsInserted', this._tableRowsInsertHandler);
+  }
+
+  protected _onTableRowsInserted(event: TableRowsInsertedEvent) {
+    // When rows are inserted in a PageWithTable, all child pages are ready, i.e. they are linked to their row and updated from it.
+    // Send a nodesChanged event in order to send summary information from parent PageWithTable to the UI server.
```
bschwarzent:
```suggestion
    // Send a nodesChanged event in order to send summary information from the PageWithTable to the UI server.
```
I think the word "parent" might be confusing here (there is only one PageWithTable involved).

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -0,0 +1,470 @@
...
+    const pages = event.rows.map(row => row.page).filter(Boolean);
+    (this.page.outline.modelAdapter as OutlineAdapter).sendNodesChanged(pages);
+  }
+
+  /**
+   * Calls loadChildPages on the UI server for the given ids or {@link PageParamDo}s. The pages are created by the Java pendant of the JsPage (see org.eclipse.scout.rt.client.ui.desktop.outline.pages.js.AbstractJsPage.createChildPage).
+   * The returned {@link Promise} is resolved when all child pages are created on the UI server.
+   *
+   * @param idsOrPageParams ids or {@link PageParamDo}s that are used to create child pages on the UI server.
+   * @param replace Whether all child pages shall be replaced. If not given the child pages will always be replaced except for the first call in order to keep existing child pages after a browser reload.
```
bschwarzent: "If not given" -> "If not specified," or "By default,"

Also, should we make clear what happens if child pages are *not* replaced? Looking at the code, when the same id is loaded twice, the existing page is still replaced. Maybe rename the parameter to make it explicit, e.g. "replaceAllChildPages"?

Is it even necessary to have this as parameter? The default behavior seems to be correct in all cases. Is this just for tests? In that case, they might just change the value of this._replaceChildPages before calling this method.

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -0,0 +1,470 @@
...
+   */
+  async callLoadChildPages(idsOrPageParams: string | PageParamDo | (string | PageParamDo)[], replace?: boolean): Promise<void> {
+    // ensure pageParams and collect them distinctly
+    const serializedPageParams = new Set<string>();
```
bschwarzent: Explain why. (Is it because the ids or page params in the table are not necessarily unique?)

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -0,0 +1,470 @@
...
+  }
+
+  /**
+   * Calls loadChildPages on the UI server for the given ids or {@link PageParamDo}s. The pages are created by the Java pendant of the JsPage (see org.eclipse.scout.rt.client.ui.desktop.outline.pages.js.AbstractJsPage.createChildPage).
+   * The returned {@link Promise} is resolved when all child pages are created on the UI server.
+   *
+   * @param idsOrPageParams ids or {@link PageParamDo}s that are used to create child pages on the UI server.
+   * @param replace Whether all child pages shall be replaced. If not given the child pages will always be replaced except for the first call in order to keep existing child pages after a browser reload.
+   */
+  async callLoadChildPages(idsOrPageParams: string | PageParamDo | (string | PageParamDo)[], replace?: boolean): Promise<void> {
```
bschwarzent: Since the method name is already in plural, we can remove the singular argument types (the caller can easily create an array themselves).


```suggestion
  async callLoadChildPages(idsOrPageParams: (string | PageParamDo)[], replace?: boolean): Promise<void> {
```

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -0,0 +1,470 @@
...
+    // ensure pageParams and collect them distinctly
+    const serializedPageParams = new Set<string>();
+    const pageParams: PageParamDo[] = [];
+    arrays.ensure(idsOrPageParams)
+      .map(idOrPageParam => this._createChildPageParam(idOrPageParam))
+      .filter(Boolean)
+      .forEach(pageParam => {
+        // The pageParam is a PageParamDo and a Set only checks "===" and not ".equals()".
+        // Therefore, stringify the PageParamDo to ensure that duplicate page params are only used once.
+        const serializedPageParam = dataObjects.stringify(pageParam);
```
bschwarzent: Could we also use the "_createChildPageId" method here instead? Then it would be same same as with the _childPagesById map.

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -0,0 +1,470 @@
...
+  /**
+   * Finds the child pages for the given ids or {@link PageParamDo}s and returns them in the order of the given ids or {@link PageParamDo}s.
+   */
+  findChildPages(idsOrPageParams: string | PageParamDo | (string | PageParamDo)[]): Page[] {
+    if (!idsOrPageParams) {
+      return [];
+    }
+
+    const idsOrPageParamsArray = arrays.ensure(idsOrPageParams);
+    if (!idsOrPageParamsArray.length) {
```
bschwarzent: Unnecessary shortcut

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -0,0 +1,470 @@
...
+   * Adds all given child pages to the {@link #_childPagesById}-map. If no child pages are given the child pages of {@link #page} are taken.
+   */
+  protected _addChildPagesToIdMap(childPages?: JsPageChildPage[]) {
+    // use child pages of this.page as default
+    childPages = childPages || this.page.childNodes;
+
+    for (const childPage of childPages) {
+      // extract and assert id
+      const id = this._extractChildPageId(childPage);
+      scout.assertValue(id, `${childPage} does not have a child page param.`);
```
bschwarzent: This will not be a useful error message, because pages don't have a toString() method. Can we simply skip the page instead?

Same in _removeChildPagesFromIdMap.

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -0,0 +1,470 @@
...
+    }
+
+    // stringify PageParamDo
+    return dataObjects.stringify(idOrPageParam);
+  }
+
+  /**
+   * Extracts the child page id from the given {@link JsPageChildPage}.
+   */
+  protected _extractChildPageId(childPage?: JsPageChildPage): string {
```
bschwarzent: Maybe swap _extractChildPageId and _extractChildPageParam, so the order is the same as with _createChildPageParam and _createChildPageId.
 
Why is this called "extract"? Extracting an object suggests that is removed from its place. This seems to be more like a "get" or "find".

org.eclipse.scout.rt.client.test/src/test/java/org/eclipse/scout/rt/client/ScoutClientDataObjectSignatureTest.java
```diff
@@ -0,0 +1,25 @@
...
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.scout.rt.client;
+
+import org.eclipse.scout.rt.dataobject.testing.signature.AbstractDataObjectSignatureTest;
+
+public class ScoutClientDataObjectSignatureTest extends AbstractDataObjectSignatureTest {
```
bschwarzent: Consider moving the DOs to org.eclipse.scout.rt.api.data, so we don't need this additional test. This is already done so for a lot of other client related DOs (bookmarks, code, config properties, permissions, preferences)

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/tree/ITreeUIFacade.java
```diff
@@ -57,4 +59,13 @@ public interface ITreeUIFacade {
   void fireAppLinkActionFromUI(String ref);
 
   void setDisplayStyleFromUI(String style);
+
+  void changeNodeFromUI(ITreeNode node, ICell cell);
```
bschwarzent: If we have it, why not use NodeCellTuple in this method as well?

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/tree/ITreeUIFacade.java
```diff
@@ -57,4 +59,13 @@ public interface ITreeUIFacade {
...
+
+  default void changeNodesFromUI(Collection<NodeCellTuple> nodeCellTuples) {
```
bschwarzent: Other UI facade methods processing multiple nodes (e.g. setNodesSelectedFromUI) to the following:
- pushUIProcessor
- setTreeChanging(true)

While the first seems not to be relevant anymore, the tree changing state might influence the number of events emitted. Is this skipped here deliberately? Maybe it would be more consistent if there was only one of the two methods on the UI facade.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/outline/pages/js/IJsPage.java
```diff
@@ -34,4 +38,8 @@ public interface IJsPage extends IPage<ITable> {
   IDoEntity getJsPageModel();
 
   void setJsPageModel(IDoEntity jsPageModel);
+
+  void loadChildPagesFromUI(List<IPageParamDo> pageParams, boolean replace);
```
bschwarzent: The "fromUI" suffix is usually reserved for UIFacades. Unless this is important, I would remove it here.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/outline/pages/js/AbstractJsPage.java
```diff
@@ -87,4 +101,106 @@ public String classId() {
...
+      setChildrenDirty(false);
+    }
+    finally {
+      getOutline().setTreeChanging(false);
+    }
+  }
+
+  protected List<IPage<?>> createChildPages(List<IPageParamDo> pageParams) {
+    if (!CollectionUtility.hasElements(pageParams)) {
+      return emptyArrayList();
```
bschwarzent: Why not return Collections.emptyList()? The method does not specify which type of list is returned. And the caller even creates a new writable array list before changing its content (line 123).

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/outline/pages/js/AbstractJsPage.java
```diff
@@ -87,4 +101,106 @@ public String classId() {
...
+
+  protected IPage<?> findExistingPage(IPageParamDo pageParam) {
+    return getChildPages().stream()
+        .filter(p -> ObjectUtility.equals(pageParam, p.getPrimaryKey()))
+        .findAny()
+        .orElse(null);
+  }
+
+  /**
+   * Override this method to create a child page using an {@link IPageParamDo} given by the Scout JS pendant of this page.
```
bschwarzent: See remark about "pendant" in JsPageHelper

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/tree/AbstractTree.java
```diff
@@ -3031,6 +3032,11 @@ public void fireAppLinkActionFromUI(String ref) {
     public void setDisplayStyleFromUI(String style) {
       setDisplayStyle(style);
     }
+
+    @Override
+    public void changeNodeFromUI(ITreeNode node, ICell cell) {
+      node.getCellForUpdate().updateFrom(cell);
```
bschwarzent: The usual pattern would be to call an instance method here instead of directly changing the model state. That way, AbstractOutline could overwrite that method instead of having to subclass the P_UIFacade.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/outline/AbstractOutline.java
```diff
@@ -1103,4 +1106,19 @@ protected void disposeTreeInternal() {
     }
     super.disposeTreeInternal();
   }
+
+  @Override
+  protected ITreeUIFacade createUIFacade() {
```
bschwarzent: Can be simplified, see comment in AbstractTree

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/outline/AbstractSearchOutline.java
```diff
@@ -123,7 +122,7 @@ public ISearchOutlineUiFacade getUIFacade() {
     return (ISearchOutlineUiFacade) super.getUIFacade();
   }
 
-  protected class P_UIFacade extends AbstractTree.P_UIFacade implements ISearchOutlineUiFacade {
+  protected class P_UIFacade extends AbstractOutline.P_UIFacade implements ISearchOutlineUiFacade {
```
bschwarzent: Not necessary, see comment in AbstractTree

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/outline/pages/js/AbstractJsPage.java
```diff
@@ -87,4 +101,106 @@ public String classId() {
...
+      return;
+    }
+
+    IDoEntity jsPageModel = getJsPageModel();
+    if (jsPageModel == null) {
+      jsPageModel = BEANS.get(DoEntity.class);
+    }
+
+    // properties come from a summary column from a parent PageWithTable, transfer them to jsPageModel so that they are still available after a browser reload
+    jsPageModel.put("text", cell.getText());
```
bschwarzent, line 201: This is the same list of properties as in JsonOutline#putJsPageObjectTypeAndModel and in JsonTree#handleUiNodesChanged. Should we link all these places with a comment, in case we add another property in the future?

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/desktop/JsPageChildPageToJsonContributor.java
```diff
@@ -0,0 +1,62 @@
...
+import org.eclipse.scout.rt.client.ui.desktop.outline.pages.IPage;
+import org.eclipse.scout.rt.client.ui.desktop.outline.pages.js.IJsPage;
+import org.eclipse.scout.rt.platform.BEANS;
+import org.eclipse.scout.rt.platform.util.ObjectUtility;
+import org.eclipse.scout.rt.ui.html.json.JsonDataObjectHelper;
+import org.json.JSONObject;
+
+/**
+ * The creation of child pages of an {@link IJsPage} is triggered by the JsPageHelper.
+ * This helper sends an {@link IPageParamDo} that is used to create the page.
```
bschwarzent: 3x the word "This", but 2x refers to the previous sentence, 1x to this class.

Suggestion: Replace the first two by "The".

PR #1770, opened 2025-10-27
Features/pbz/26.1/config file credential verifier

paolobazzi:


mvilliger (approved):
May also be a candidate for backport to 25.2

PR #1775, opened 2025-10-29
Fix flip/flop BeanCreationWithExceptionsTest

matthiaso:
There is no guarantee that the code within both futures is actually called concurrently, it may as well be called after each other or in any order. Even the second future may be the future to instantiate the bean while the first future may just return the created bean/throw an exception.

### DIFF
--- a/org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/BeanCreationWithExceptionsTest.java
+++ b/org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/BeanCreationWithExceptionsTest.java
@@ -11,8 +11,11 @@ package org.eclipse.scout.rt.platform;
 
 import static org.junit.Assert.*;
 
+import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import jakarta.annotation.PostConstruct;
 
@@ -21,6 +24,7 @@ import org.eclipse.scout.rt.platform.exception.ExceptionHandler;
 import org.eclipse.scout.rt.platform.exception.PlatformException;
 import org.eclipse.scout.rt.platform.job.IFuture;
 import org.eclipse.scout.rt.platform.job.Jobs;
+import org.eclipse.scout.rt.platform.util.Assertions;
 import org.eclipse.scout.rt.platform.util.SleepUtil;
 import org.eclipse.scout.rt.platform.util.concurrent.ThreadInterruptedError;
 import org.eclipse.scout.rt.testing.platform.runner.PlatformTestRunner;
@@ -168,8 +172,7 @@ public class BeanCreationWithExceptionsTest {
     IFuture<TestBean> f1 = scheduleGetBean();
     IFuture<TestBean> f2 = scheduleGetBean();
     state.constructorLatch.countDown();
-    f1.awaitDoneAndGet();
-    f2.awaitDoneAndGet();
+    assertSame(f1.awaitDoneAndGet(), f2.awaitDoneAndGet());
     state.assertInvocations(1, 1);
   }
 
@@ -181,20 +184,9 @@ public class BeanCreationWithExceptionsTest {
     IFuture<TestBean> f1 = scheduleGetBean();
     IFuture<TestBean> f2 = scheduleGetBean();
     state.constructorLatch.countDown();
-    try {
-      f1.awaitDoneAndGet();
-      fail("expecting exception");
-    }
-    catch (BeanCreationException e) {
-      assertTestBeanException(true, e);
-    }
-    try {
-      f2.awaitDoneAndGet();
-      fail("expecting exception");
-    }
-    catch (BeanCreationException e) {
-      // expected
-    }
+    assertTestBeanException(true,
+        assertThrows(BeanCreationException.class, () -> f1.awaitDoneAndGet()),
+        assertThrows(BeanCreationException.class, () -> f2.awaitDoneAndGet()));
     state.assertInvocations(0, 0);
   }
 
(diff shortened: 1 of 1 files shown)


org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/BeanCreationWithExceptionsTest.java
```diff
@@ -168,8 +169,9 @@ public void testConcurrentSuccessfulConstruction() {
     IFuture<TestBean> f1 = scheduleGetBean();
     IFuture<TestBean> f2 = scheduleGetBean();
     state.constructorLatch.countDown();
-    f1.awaitDoneAndGet();
-    f2.awaitDoneAndGet();
+    assertEquals(f1.awaitDoneAndGet(), f2.awaitDoneAndGet());
```
bschwarzent: Better use `assertSame`?

org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/BeanCreationWithExceptionsTest.java
```diff
@@ -168,8 +169,9 @@ public void testConcurrentSuccessfulConstruction() {
...
+    // according to javadoc on @ApplicationScoped it is not guaranteed that the TestBean is constructed only once
+    // if this test ever fails/flip-flops we may need to increase the first argument on the following line
```
bschwarzent: Is this comment helpful? Increasing the numbers will not help, because the next run might be different again. -> maybe refer to one of the examples below instead.

org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/BeanCreationWithExceptionsTest.java
```diff
@@ -346,6 +292,18 @@ private static IFuture<TestBean> scheduleGetBean() {
     return future;
   }
 
+  private static void assertTestBeanException(boolean expectedFromConstructor, BeanCreationException firstException, BeanCreationException secondException) {
```
bschwarzent, line 296: This method should not care that the second exception is the "rare" case -> reformulate comments.

org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/BeanCreationWithExceptionsTest.java
```diff
@@ -346,6 +292,18 @@ private static IFuture<TestBean> scheduleGetBean() {
...
+    try {
+      // assertion must match either for firstException
+      assertTestBeanException(expectedFromConstructor, firstException);
+    }
+    catch (AssertionError e) {
+      // or in rare cases for secondException (parts of second future may be run before parts of first future)
+      LOG.info("Fallback, check other exception", e);
```
bschwarzent: This is the same exception we already logged in assertTestBeanException - necessary? Or should output first and second exception instead?

org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/BeanCreationWithExceptionsTest.java
```diff
@@ -346,6 +292,18 @@ private static IFuture<TestBean> scheduleGetBean() {
     return future;
   }
 
+  private static void assertTestBeanException(boolean expectedFromConstructor, BeanCreationException firstException, BeanCreationException secondException) {
+    try {
+      // assertion must match either for firstException
+      assertTestBeanException(expectedFromConstructor, firstException);
+    }
+    catch (AssertionError e) {
```
bschwarzent: Is this AssertionError the one with the missing cause? If yes, why does this method not check which of the two exceptions has a cause first? Or do we really want to try the second exception if something else fails, e.g. wrong "fromConstructor" state?

org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/BeanCreationWithExceptionsTest.java
matthiaso: https://github.com/eclipse-scout/scout.rt/compare/5e0a4c95ecfb32caabd5e85655120332ab4b26de..4e2cd0b948b1793e86d467cb2ba8ec1c394ede49

org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/BeanCreationWithExceptionsTest.java
matthiaso: https://github.com/eclipse-scout/scout.rt/compare/5e0a4c95ecfb32caabd5e85655120332ab4b26de..4e2cd0b948b1793e86d467cb2ba8ec1c394ede49

org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/BeanCreationWithExceptionsTest.java
matthiaso: https://github.com/eclipse-scout/scout.rt/compare/5e0a4c95ecfb32caabd5e85655120332ab4b26de..4e2cd0b948b1793e86d467cb2ba8ec1c394ede49

org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/BeanCreationWithExceptionsTest.java
matthiaso: https://github.com/eclipse-scout/scout.rt/compare/5e0a4c95ecfb32caabd5e85655120332ab4b26de..4e2cd0b948b1793e86d467cb2ba8ec1c394ede49

org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/platform/BeanCreationWithExceptionsTest.java
matthiaso: https://github.com/eclipse-scout/scout.rt/compare/5e0a4c95ecfb32caabd5e85655120332ab4b26de..4e2cd0b948b1793e86d467cb2ba8ec1c394ede49

bschwarzent (approved):

PR #1780, opened 2025-10-30
Fix flip/flop BeanCreationWithExceptionsTest

matthiaso:
There is no guarantee that the code within both futures is actually called concurrently, it may as well be called after each other or in any order. Even the second future may be the future to instantiate the bean while the first future may just return the created bean/throw an exception.


matthiaso:
Cherry-pick of https://github.com/eclipse-scout/scout.rt/pull/1775

PR #1783, opened 2025-10-30
Fix flip/flop PeriodicAsyncMeasurementTest

matthiaso:
ExecutionSemaphore is released asynchronously, therefore competition count must not be asserted immediately, instead we may need to wait a short while until competition count can be checked/asserted.

awaitCondition does implicitly assert that this condition is met otherwise an exception is thrown on timeout.


matthiaso:
Cherry-pick of https://github.com/eclipse-scout/scout.rt/pull/1782#pullrequestreview-3399511993

PR #1785, opened 2025-10-30
Add Order annotation for IDataObjectSerializerProvider impls

paolobazzi:
IDataObjectSerializerProvider is used by BEANS.all() which uses alphabetical order as default. Add fixed order for main scout serializer provider implementation.

### DIFF
--- a/org.eclipse.scout.rt.jackson/src/main/java/org/eclipse/scout/rt/jackson/dataobject/ScoutDataObjectSerializerProvider.java
+++ b/org.eclipse.scout.rt.jackson/src/main/java/org/eclipse/scout/rt/jackson/dataobject/ScoutDataObjectSerializerProvider.java
@@ -35,6 +35,7 @@ import org.eclipse.scout.rt.jackson.dataobject.id.UnqualifiedIIdDeserializer;
 import org.eclipse.scout.rt.jackson.dataobject.id.UnqualifiedIIdMapKeyDeserializer;
 import org.eclipse.scout.rt.jackson.dataobject.id.UnqualifiedIIdMapKeySerializer;
 import org.eclipse.scout.rt.jackson.dataobject.id.UnqualifiedIIdSerializer;
+import org.eclipse.scout.rt.platform.Order;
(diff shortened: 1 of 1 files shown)


org.eclipse.scout.rt.jackson/src/main/java/org/eclipse/scout/rt/jackson/dataobject/ScoutDataObjectSerializerProvider.java
```diff
@@ -51,6 +52,7 @@
 import com.fasterxml.jackson.databind.type.CollectionType;
 import com.fasterxml.jackson.databind.type.ReferenceType;
 
+@Order(4000)
```
fschinkel: 4000 is reserved for projects (see [Tech-Doc](https://eclipsescout.github.io/scout-docs/22/technical-guide/common-concepts/platform.html#order)). Please stick to the range 4001-5999.
Do we need to add a specific order or is 5000 (default) good enough? Maybe we add an order but use the value `IBean.DEFAULT_BEAN_ORDER`. If we add any other value than 5000, please add a migration guide.

fschinkel (changes requested):

org.eclipse.scout.rt.jackson/src/main/java/org/eclipse/scout/rt/jackson/dataobject/ScoutDataObjectSerializerProvider.java
paolobazzi: Changed to 4500. I'll suggest to not use 5000 (default) on purpose, in order that defautl non-annotated custom implementations runs as last provider.

fschinkel (approved):
Looks good.
Please add migration guide and describe how providing a new Serializer/Deserializer works. I think the [docs](https://eclipsescout.github.io/scout-docs/22/technical-guide/working-with-data/data-object.html#extending-with-custom-serializer-and-deserializer) are outdated.

PR #1787, opened 2025-10-31
Fix Group Body

bsifpa:
Don't render the body, if it gets set while collapsed


fschinkel (commented):
Please add a small explanation of the problem to the commit mesage. E.g.:

If the group gets expanded it renders its body. If the body is already rendered when this happens an 'Already rendered'-error is thrown. Therefore, prevent the body from being rendered when it is changed while the group is collapsed.

fschinkel (approved):

PR #1791, opened 2025-11-03
New test for the correct naming of concrete type version classes

michaelze:
When creating a new NamespaceVersion from a type version class name, the name of the class (up until the first underscore) is converted to the namespace id with the first letter being transformed to lower case (the rest of the classname is left as-is). As all namespace ids only consist of lowercase letters, this assumes that the type version class name only consists of lowercase letters (except the first one). When this assumption is not true, problems arise when trying to resolve namespaces by the created NamespaceVersion as the contained namespace id (with incorrect casing) does not match any known namespace id. The problems that can arise are hard to debug because there's no hint at the wrong casing, every class seems to be present and have the right name. This is why we chose to contribute a test case that checks for the correct naming of ITypeVersion implementations. In addition, a test was added to the AbstractDataObjectTestCompletenessTest as a reminder to add the mentioned test class.

### DIFF
--- a/org.eclipse.scout.rt.dataobject.test/src/main/java/org/eclipse/scout/rt/dataobject/testing/AbstractTypeVersionClassNamingTest.java
+++ b/org.eclipse.scout.rt.dataobject.test/src/main/java/org/eclipse/scout/rt/dataobject/testing/AbstractTypeVersionClassNamingTest.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.scout.rt.dataobject.testing;
+
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Matcher;
+
+import org.eclipse.scout.rt.dataobject.AbstractTypeVersion;
+import org.eclipse.scout.rt.dataobject.ITypeVersion;
+import org.eclipse.scout.rt.platform.BEANS;
+import org.eclipse.scout.rt.platform.util.CollectionUtility;
+import org.eclipse.scout.rt.platform.util.StringUtility;
+import org.eclipse.scout.rt.testing.platform.runner.PlatformTestRunner;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(PlatformTestRunner.class)
+public abstract class AbstractTypeVersionClassNamingTest {
+
+  @Test
+  public final void testTypeVersions() {
+    List<String> malformedTypeVersionClassNames = collectMalformedTypeVersionClassNames();
+    if (!malformedTypeVersionClassNames.isEmpty()) {
+      Assert.fail("Malformed type version class names found.\nType version class names must follow the following pattern: "
+          + "<Namespace>_<version1>[_<version2>[_<versionN>]][__<comment>] where <Namespace> must be called after the respective scout namespace with all "
+          + "letters in lower case except for the first letter, which must be capitalized.\n- " + String.join("\n- ", malformedTypeVersionClassNames));
+    }
+  }
+
+  protected abstract String getPackageNamePrefix();
+
+  protected Set<Class<ITypeVersion>> getExclusionList() {
+    return CollectionUtility.emptyHashSet();
+  }
+
+  protected List<String> collectMalformedTypeVersionClassNames() {
+    String packagePrefix = getPackageNamePrefix();
+
+    Set<Class<ITypeVersion>> excludedTypeVersions = getExclusionList();
+
+    return BEANS.all(ITypeVersion.class).stream()
+        .map(Object::getClass)
+        .filter(typeVersionClass -> !excludedTypeVersions.contains(typeVersionClass))
+        .filter(typeVersionClass -> typeVersionClass.getPackageName().startsWith(packagePrefix))
+        .map(Class::getSimpleName)
+        .filter(this::isMalformedTypeVersionClassName)
+        .toList();
+  }
+
+  protected boolean isMalformedTypeVersionClassName(String className) {
+    Matcher matcher = AbstractTypeVersion.CLASS_NAME_PATTERN.matcher(className);
+    if (!matcher.matches()) {
+      return false;
+    }
+
+    String actualNamespaceIdString = matcher.group(1);
+    String expectedNamespaceIdString = StringUtility.uppercaseFirst(actualNamespaceIdString.toLowerCase());
+
+    return !actualNamespaceIdString.equals(expectedNamespaceIdString);
+  }
+}
--- a/org.eclipse.scout.rt.dataobject.test/src/main/java/org/eclipse/scout/rt/dataobject/testing/TypeVersionClassNamingTestSupport.java
+++ b/org.eclipse.scout.rt.dataobject.test/src/main/java/org/eclipse/scout/rt/dataobject/testing/TypeVersionClassNamingTestSupport.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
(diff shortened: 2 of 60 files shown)


org.eclipse.scout.rt.dataobject.test/src/main/java/org/eclipse/scout/rt/dataobject/testing/AbstractTypeVersionClassNamingTest.java
```diff
@@ -0,0 +1,60 @@
...
+    if (!StringUtility.isNullOrEmpty(malformedTypeVersionClassNames)) {
+      Assert.fail("Malformed type version class names found.\nType version class names must follow the following pattern: "
+          + "<Namespace>_<version1>[_<version2>[_<versionN>]][__<comment>] where <Namespace> must be called after the respective scout namespace with all "
+          + "letters in lower case except for the first letter, which must be capitalized.\n- " + malformedTypeVersionClassNames);
+    }
+  }
+
+  protected abstract String getPackageNamePrefix();
+
+  protected String collectMalformedTypeVersionClassNames() {
```
rogersteblerbsi: I think it's better if this method returns a `List` instead of an already concatenated `String`. Then it's easier to override it and manipulate the list if needed. The concatenation can then be done in the error message.

org.eclipse.scout.rt.dataobject.test/src/main/java/org/eclipse/scout/rt/dataobject/testing/AbstractTypeVersionClassNamingTest.java
```diff
@@ -0,0 +1,60 @@
...
+    String packagePrefix = getPackageNamePrefix();
+
+    return BEANS.all(ITypeVersion.class).stream()
+        .filter(typeVersion -> typeVersion.getClass().getPackageName().startsWith(packagePrefix))
+        .map(typeVersion -> typeVersion.getClass().getSimpleName())
+        .filter(this::malformedTypeVersionClassNamePredicate)
+        .collect(Collectors.joining("\n- "));
+  }
+
+  protected boolean malformedTypeVersionClassNamePredicate(String className) {
```
rogersteblerbsi: Since it returns a `boolean` and not a `Predicate` I would name it `isMalformedTypeVersionClassName`

org.eclipse.scout.rt.dataobject.test/src/main/java/org/eclipse/scout/rt/dataobject/testing/TypeVersionClassNamingTestSupport.java
```diff
@@ -0,0 +1,46 @@
...
+package org.eclipse.scout.rt.dataobject.testing;
+
+import java.nio.file.Path;
+import java.util.regex.Pattern;
+
+public class TypeVersionClassNamingTestSupport extends AbstractDataObjectTestSupport {
+
+  @Override
+  protected Pattern createFilePattern() {
+    return Pattern.compile("class \\w+ extends AbstractTypeVersion");
```
rogersteblerbsi: Maybe also check that after `AbstractTypeVersion` is a word boundary? To prevent a match for example for `extends AbstractTypeVersionTest` or something similar.

org.eclipse.scout.rt.dataobject.test/src/main/java/org/eclipse/scout/rt/dataobject/testing/AbstractTypeVersionClassNamingTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(PlatformTestRunner.class)
+public abstract class AbstractTypeVersionClassNamingTest {
+
+  @Test
+  public final void testTypeVersions() {
+    String malformedTypeVersionClassNames = collectMalformedTypeVersionClassNames();
```
rogersteblerbsi: Maybe it also needs the possibility to add exclusions?

org.eclipse.scout.rt.dataobject.test/src/main/java/org/eclipse/scout/rt/dataobject/testing/AbstractTypeVersionClassNamingTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.eclipse.scout.rt.dataobject.ITypeVersion;
+import org.eclipse.scout.rt.platform.BEANS;
+import org.eclipse.scout.rt.platform.util.StringUtility;
+import org.eclipse.scout.rt.testing.platform.runner.PlatformTestRunner;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(PlatformTestRunner.class)
+public abstract class AbstractTypeVersionClassNamingTest {
```
rogersteblerbsi, line 27: `ScoutDataObjectTestCompletenessTest` currently fails, because the `ScoutTypeVersionClassNamingTest` is missing.

org.eclipse.scout.rt.dataobject.test/src/main/java/org/eclipse/scout/rt/dataobject/testing/AbstractTypeVersionClassNamingTest.java
```diff
@@ -0,0 +1,60 @@
...
+  protected boolean malformedTypeVersionClassNamePredicate(String className) {
+    Matcher matcher = AbstractTypeVersion.CLASS_NAME_PATTERN.matcher(className);
+    if (!matcher.matches()) {
+      return false;
+    }
+
+    String actualNamespaceIdString = matcher.group(1);
+    String expectedNamespaceIdString = StringUtility.uppercaseFirst(actualNamespaceIdString.toLowerCase());
+
+    return !actualNamespaceIdString.equals(expectedNamespaceIdString);
```
rogersteblerbsi, line 68: What will we do with all the Fixture Type Versions that we have in Scout? For example `DataObjectFixture_1_0_0` will be invalid. Rename them all? Or just exclude them?

rogersteblerbsi (changes requested):

org.eclipse.scout.rt.dataobject.test/src/main/java/org/eclipse/scout/rt/dataobject/testing/AbstractTypeVersionClassNamingTest.java
michaelze: Thats a good point which also raises the question why they do have a "wrong" name in the first place.
While developing this change I found it increasingly strange, that the rule we are testing for is not enforced more strictly in the code (e.g. when constructing a NamespaceVersion object).
On the other hand: why are we selectively converting only the first letter to lower case and not the whole namespace part of the type version class name?

What are your thoughts on that?

org.eclipse.scout.rt.dataobject.test/src/main/java/org/eclipse/scout/rt/dataobject/testing/AbstractTypeVersionClassNamingTest.java
```diff
@@ -0,0 +1,70 @@
...
+      Assert.fail("Malformed type version class names found.\nType version class names must follow the following pattern: "
+          + "<Namespace>_<version1>[_<version2>[_<versionN>]][__<comment>] where <Namespace> must be called after the respective scout namespace with all "
+          + "letters in lower case except for the first letter, which must be capitalized.\n- " + String.join("\n- ", malformedTypeVersionClassNames));
+    }
+  }
+
+  protected abstract String getPackageNamePrefix();
+
+  protected Set<Class<ITypeVersion>> getExclusionList() {
+    return Collections.emptySet();
```
rogersteblerbsi: I would return a mutable set here, so that you can just use the supercall to add additional exclusions.

```suggestion
    return CollectionUtility.emptyHashSet();
```

org.eclipse.scout.rt.dataobject.test/src/main/java/org/eclipse/scout/rt/dataobject/testing/AbstractTypeVersionClassNamingTest.java
rogersteblerbsi: Not sure why we use these names for the Fixture Type Versions that don't match the naming pattern. Maybe @paolobazzi or @LukasHuser know more about this topic?

paolobazzi: I think those names are a (legacy) shortcut choosen to simplify tests (and using differtent namespaces we did not have to deal with duplicated type names).
Those namespace could be adjusted/fixed (+ check for duplicated type names)

PR #1796, opened 2025-11-04
Stop propagating clicks on chart area

bsifpa:
Clicks on the chart area are no longer bubbling up. Instead, an event is triggered when a click is made within the chart area but outside an element. This allows additional actions to be registered in a more targeted manner.

430441

### DIFF
--- a/eclipse-scout-chart/src/chart/AbstractSvgChartRenderer.ts
+++ b/eclipse-scout-chart/src/chart/AbstractSvgChartRenderer.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -25,6 +25,8 @@ export class AbstractSvgChartRenderer extends AbstractChartRenderer {
 
   $svg: JQuery<SVGElement>;
 
+  protected _nonValueClickHandler: (event: JQuery.ClickEvent) => void;
+
   constructor(chart: Chart) {
     super(chart);
     this.chartBox = null;
@@ -33,6 +35,8 @@ export class AbstractSvgChartRenderer extends AbstractChartRenderer {
     this.maskId = 'Mask-' + ObjectFactory.get().createUniqueId();
 
     this.suppressLegendBox = false;
+
+    this._nonValueClickHandler = this._onNonValueClick.bind(this);
   }
 
   static FONT_SIZE_SMALLEST = 'smallestFont';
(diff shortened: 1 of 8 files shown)


eclipse-scout-chart/src/chart/ChartJsRenderer.ts
```diff
@@ -121,6 +121,7 @@ export class ChartJsRenderer extends AbstractChartRenderer {
   protected _tooltip: Tooltip;
   protected _tooltipTimeoutId: number;
   protected _updatingDatalabels: boolean;
+  protected _preventCanvasClick: boolean;
```
fschinkel: Can we think of a name that describes state of the chart and not the result? Maybe we need this in the future for other use cases. 
If the value is `true` all clicks trigger specific events, so can we say the chart is interactive or these clicks are interactive? So maybe `_interactiveClicks`? What do you think?

eclipse-scout-chart/src/chart/ChartEventMap.ts
```diff
@@ -10,15 +10,27 @@
...
   data: ClickObject;
   originalEvent?: Event;
 }
 
+export interface ChartNonValueClickEvent<C = Chart> extends ScoutEvent<C> {
+  originalEvent?: Event;
+}
+
+export interface ChartLegendItemClickEvent<C = Chart> extends ScoutEvent<C> {
+  legendItem?: LegendItem;
```
fschinkel: We do not want to expose chart.js-types. Maybe its enough to simply add the legendItem-index.

eclipse-scout-chart/src/chart/Chart.ts
```diff
@@ -399,6 +399,19 @@ export class Chart extends Widget implements ChartModel {
...
     });
   }
+
+  handleNonValueClick(originalEvent?: Event) {
+    this.trigger('nonValueClick', {
+      originalEvent
+    });
+  }
+
+  handleLegendClick(originalEvent: Event, legentItem: LegendItem) {
```
fschinkel:
```suggestion+0-0
  handleLegendClick(legentItemIndex: number, originalEvent?: Event) {
```

eclipse-scout-chart/src/chart/Chart.ts
```diff
@@ -399,6 +399,19 @@ export class Chart extends Widget implements ChartModel {
       originalEvent
     });
   }
+
+  handleNonValueClick(originalEvent?: Event) {
```
fschinkel, line 403: Please add these events for all other charts as well (Fulfillment, Salesfunnel, Venn, Speedo)

fschinkel (changes requested):

eclipse-scout-chart/src/chart/AbstractSvgChartRenderer.ts
```diff
@@ -12,6 +12,15 @@ import $ from 'jquery';
 import {AbstractChartRenderer, Chart} from '../index';
 import {ClickObject} from './Chart';
 
+export type ChartBox = {
```
fschinkel: Why did you move this to the top of the file?

fschinkel (approved):

PR #1797, opened 2025-11-06
Improve exception handling of malware scanner

Knicksaas:
The new implementation throws an ProcessingException when the folder is
not available.


jwinkens (commented):
Looks good

PR #1806, opened 2025-11-12
Http Service Tunnel: Adapt API to enable extensibility

paolobazzi:


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/servicetunnel/HttpServiceTunnel.java
```diff
@@ -192,21 +192,9 @@ protected ServiceTunnelResponse tunnel(final ServiceTunnelRequest serviceRequest
...
-            return new ServiceTunnelResponse(new ProcessingException("Response contains no content")
-                .withContextInfo("http-status", "{} {}", resp.getStatus(), resp.getStatusInfo())
-                .withContextInfo("http-headers", resp.getHeaders() + ""));
-          }
-          return response;
-        }
-      }
+      Response response = executeRequest(serviceRequest, requestData);  // response will be closed by #readResponse
+      interceptHttpResponse(response, serviceRequest);
+      return readResponse(response, contentHandler);
```
matthiaso: Previously if `interceptHttpResponse` did fail for some reason we did still close the response as it was included in the try-with-resources; now we would not close the `response` anymore. 

Also if `readResponse` (maybe overridden) fails before the try-with-resources (in this class) for `response` fails we would not close it anymore.

I do not know if that can ever happen though; but still maybe we need to add this information to the javadoc of `interceptHttpResponse`? Or we could add (yet another) try-catch block and in case of an error always close the `response`.

Not sure if this can ever be a problem.

matthiaso (approved):

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/servicetunnel/ServiceTunnelService.java
```diff
@@ -63,38 +65,61 @@ public void init() {
...
+   * Handles incoming request by given {@link InputStream}, delegates handling of {@link ServiceTunnelResponse} to {@link #handleResponse(ServiceTunnelResponse)} default implementation.
    */
-  protected boolean enableSignature(HttpServletRequest servletRequest) {
-    return Boolean.TRUE.toString().equalsIgnoreCase(servletRequest.getHeader(IdSignatureClientRequestFilter.ID_SIGNATURE_HTTP_HEADER));
+  public Response incomingRequest(InputStream in) {
+    return handleIncomingRequest(in, this::handleResponse);
+  }
+
+  /**
+   * Default handling, writes {@link ServiceTunnelResponse} directly to given {@link OutputStream}.
```
matthiaso: Which given `OutputStream`? No `OutputStream` is given to the method. 

Maybe:

```suggestion
   * Default handling, writes {@link ServiceTunnelResponse} directly to the returned {@link Response}.
   * 
   * @see #serializeServiceResponse(OutputStream, ServiceTunnelResponse) 
```

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/servicetunnel/ServiceTunnelService.java
```diff
@@ -63,38 +65,61 @@ public void init() {
...
+    if (enableSignature(IHttpServletRoundtrip.CURRENT_HTTP_SERVLET_REQUEST.get())) {
+      RunContext.CURRENT.get().getPropertyMap().put(ServiceTunnelOptions.ID_SIGNATURE_PROP, true);
+    }
+
+    // FIXME PBZ,MOT: Check if this m_serverRunContextProducer is necessary -> org.eclipse.scout.rt.server.context.ServerRunContextFilter already provided same/similar RunContext
+//    return m_serverRunContextProducer.get()
+//        .getInnerRunContextProducer()
+//        .produce(IHttpServletRoundtrip.CURRENT_HTTP_SERVLET_REQUEST.get(), IHttpServletRoundtrip.CURRENT_HTTP_SERVLET_RESPONSE.get())
+//        //.withProperties(enableSignature(IHttpServletRoundtrip.CURRENT_HTTP_SERVLET_REQUEST.get()) ? Map.of(ServiceTunnelOptions.ID_SIGNATURE_PROP, true) : Map.of())
+//        .call(() -> {
```
matthiaso: We probably do not need the additional `RunContext` here, previous purpose was I believe to add the ìd signature run context property (which we now add directly to the property map, see https://github.com/eclipse-scout/scout.rt/pull/1806/files#diff-76a72b3ca6d38aa5eca2d2ee56f4066ac28f3646599378e9d1f7bc4135e21635R96).

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/servicetunnel/ServiceTunnelService.java
```diff
@@ -63,38 +65,61 @@ public void init() {
...
+    try {
+      ServiceTunnelRequest serviceRequest = deserializeServiceRequest(in);
+      ServiceTunnelResponse serviceResponse = evaluate(serviceRequest);
+      return responseConsumer.handle(serviceResponse);
+    } catch (Exception e) {
+      // FIXME PBZ;MOT check exception handling, remove translator and switch to classic try/catch?
+      // Or just throw out of ProcessResource and let Jersey exception mapper do the job?
+      throw BEANS.get(ServiceTunnelExceptionTranslator.class).translate(e);
+    }
+//        }, ServiceTunnelExceptionTranslator.class);
```
matthiaso: Maybe even alright to remove the translator, it does not seem to contain much special code (maybe not anymore). - We could remove it and run additional tests.

matthiaso (approved):

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/servicetunnel/ServiceTunnelService.java
paolobazzi: done

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/servicetunnel/ServiceTunnelService.java
paolobazzi: removed

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/servicetunnel/ServiceTunnelService.java
paolobazzi: removed

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/servicetunnel/HttpServiceTunnel.java
paolobazzi: added try-catch for intercept response

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/servicetunnel/HttpServiceTunnel.java
```diff
@@ -192,21 +192,15 @@ protected ServiceTunnelResponse tunnel(final ServiceTunnelRequest serviceRequest
...
-        }
+      Response response = executeRequest(serviceRequest, requestData);  // response will be closed by #readResponse
+      try {
+        interceptHttpResponse(response, serviceRequest);
+      }
+      catch (Throwable t) {
+        response.close(); // close response in case of exception in #interceptHttpResponse
+        throw t;
       }
+      return readResponse(response, contentHandler);
```
matthiaso:
```suggestion
        return readResponse(response, contentHandler);
      }
      catch (Throwable t) {
        response.close(); // close response in case of exception in #interceptHttpResponse
        throw t;
      }
```

Not sure if `readResponse` should be within `try` block (what if an exception occurs during `readResponse`, do we not have to close it as well)? At least our implementation of `readResponse` seems to handle this case and has a try-with-resources block around the `response` - maybe not every implementation of `readResponse` will have this `try-with-resources` block for the response (as it will always close the response) but at least in case of an error is it expected that the response will be closed always? Maybe we could also add it to javadoc of `readResponse`?

matthiaso (approved):

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/servicetunnel/HttpServiceTunnel.java
paolobazzi: Moved readResponse within try/catch in order to close response in case of an error.

PR #1810, opened 2025-11-14
Improve exception handling of malware scanner

Knicksaas:
The new implementation throws an ProcessingException when the folder is not available.

422721


Knicksaas:
Cherry-pick from #1810

PR #1814, opened 2025-11-17
MalwareScanner: Add support for RemoteSystemUnavailableException

Knicksaas:
When the MalwareScanner is not available a
RemoteSystemUnavailableException can be thrown to adress this case. This allows a more detailed error handling.

422721

### DIFF
--- a/eclipse-scout-core/src/session/Session.ts
+++ b/eclipse-scout-core/src/session/Session.ts
@@ -1081,7 +1081,9 @@ export class Session extends EventEmitter implements SessionModel, ModelAdapterL
       isFatalError = false; // unsafe upload allows the application to continue
     } else if (jsonError.code === Session.JsonResponseError.REMOTE_SYSTEM_UNAVAILABLE) {
       boxOptions.header = this.optText('NetErrorTitle', boxOptions.header);
-      boxOptions.body = this.optText('NetSystemsNotAvailable', boxOptions.body);
+      boxOptions.body = this.optText('NetSystemsNotAvailable', 'The system is partially unavailable at the moment.')
+        + '\n\n'
+        + this.optText('PleaseTryAgainLater', 'Please try again later.');
       boxOptions.yesButtonText = this.optText('ui.Ok', 'Ok');
       boxOptions.yesButtonAction = null; // NOP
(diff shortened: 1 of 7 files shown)


org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/JsonRequestHelper.java
```diff
@@ -118,6 +118,17 @@ public JSONObject createVersionMismatchResponse() {
     return response.toJson();
   }
 
+  /**
+   * UI Text <code>NetSystemsNotAvailable</code>
```
bschwarzent: This is unnecessary, remove it.

I would remove it from createUnsafeUploadResponse and createRejectedUploadResponse as well.

eclipse-scout-core/src/session/Session.ts
```diff
@@ -1078,6 +1079,12 @@ export class Session extends EventEmitter implements SessionModel, ModelAdapterL
       boxOptions.yesButtonText = this.optText('ui.Ok', 'Ok');
       boxOptions.yesButtonAction = null; // NOP
       isFatalError = false; // unsafe upload allows the application to continue
+    } else if (jsonError.code === Session.JsonResponseError.REMOTE_SYSTEM_UNAVAILABLE) {
+      boxOptions.header = this.optText('NetErrorTitle', boxOptions.header);
```
bschwarzent, line 1083: This NLS key has to be added to the UiTextContributor, otherwise the text might not be available.

eclipse-scout-core/src/session/Session.ts
```diff
@@ -1078,6 +1079,12 @@ export class Session extends EventEmitter implements SessionModel, ModelAdapterL
...
+      boxOptions.body = this.optText('NetSystemsNotAvailable', boxOptions.body);
```
bschwarzent: If possible, add a second line: "Please try again later."

-> analog to ErrorHandler.ts (getMessageBodyForHttpStatus)

bschwarzent (approved):

PR #1818, opened 2025-11-18
Support for Java 25 runtime

rogersteblerbsi:
422904

### DIFF
--- a/org.eclipse.scout.rt/pom.xml
+++ b/org.eclipse.scout.rt/pom.xml
@@ -14,7 +14,7 @@
   <parent>
     <groupId>org.eclipse.scout</groupId>
     <artifactId>maven_rt_plugin_config-master</artifactId>
-    <version>25.2.1</version>
+    <version>25.2.2</version>
     <relativePath />
   </parent>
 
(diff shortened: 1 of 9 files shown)


org.eclipse.scout.rt/pom.xml
```diff
@@ -14,7 +14,7 @@
   <parent>
     <groupId>org.eclipse.scout</groupId>
     <artifactId>maven_rt_plugin_config-master</artifactId>
-    <version>25.2.1</version>
+    <version>25.2.2-SNAPSHOT</version>
```
matthiaso: Same comment as https://github.com/eclipse-scout/scout.docs/pull/349/files/254d491706c0f790709352ee0f25ba72ade713d2#diff-a38ef1a9d6048623381b6008f764fac4772a619d84b440d1b6d4c13ec625e0be

I suppose snapshot version is alright as it will be fixed at some later point of time (during build)?

matthiaso (approved):

org.eclipse.scout.rt/pom.xml
rogersteblerbsi: Changed to 25.2.2. We will build the maven master soon.

PR #1822, opened 2025-11-19
MalwareScanner: Add support for RemoteSystemUnavailableException

Knicksaas:
When the MalwareScanner is not available a
RemoteSystemUnavailableException can be thrown to adress this case. This allows a more detailed error handling.

422721


Knicksaas:
Cherry pick from #1814

PR #1833, opened 2025-11-24
Deprecate FileSystemMalwareScannerImplementor

mvilliger:
This scanner is unsafe and has lots of preconditions. Furthermore, it has issues with large files and is hard to configure correctly (timeout) to ensure no malware is missed.

436096


paolobazzi (approved):
looks good to me

PR #1834, opened 2025-11-24
Allow cancellation w/o session as long as subject is available

matthiaso:
425245

### DIFF
--- a/org.eclipse.scout.rt.platform.test/src/main/java/org/eclipse/scout/rt/testing/platform/mock/RegisterBeanTestRule.java
+++ b/org.eclipse.scout.rt.platform.test/src/main/java/org/eclipse/scout/rt/testing/platform/mock/RegisterBeanTestRule.java
@@ -9,22 +9,30 @@
  */
 package org.eclipse.scout.rt.testing.platform.mock;
 
+import static org.junit.Assert.assertEquals;
+
+import java.util.Objects;
 import java.util.function.Supplier;
 
 import org.eclipse.scout.rt.platform.BeanMetaData;
 import org.eclipse.scout.rt.platform.IBean;
+import org.eclipse.scout.rt.platform.Platform;
 import org.eclipse.scout.rt.testing.platform.BeanTestingHelper;
 import org.eclipse.scout.rt.testing.platform.util.AbstractScoutTestRule;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Shortcut to replace a application scoped bean within a test.
  */
 public class RegisterBeanTestRule<BEAN> extends AbstractScoutTestRule {
 
+  private static final Logger LOG = LoggerFactory.getLogger(RegisterBeanTestRule.class);
+
   private final Class<? super BEAN> m_beanClazz;
-  private final Supplier<BEAN> m_mockSupplier;
+  private final Supplier<? extends BEAN> m_mockSupplier;
 
   private IBean<?> m_temporaryBean;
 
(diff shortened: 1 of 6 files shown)


org.eclipse.scout.rt.platform.test/src/main/java/org/eclipse/scout/rt/testing/platform/mock/RegisterBeanTestRule.java
```diff
@@ -38,8 +39,11 @@ public RegisterBeanTestRule(Class<? super BEAN> beanClazz, Supplier<BEAN> mockSu
   }
 
   public void registerBean() {
+    // load existing bean (optional) to overwrite potential replacements of m_beanClazz (use actual registered beanClazz)
+    // this may even result in an error if our mock does not implement the currently registered beanClazz, it is necessary however as otherwise it would not be guaranteed that our mock would be used
```
paolobazzi: should we add an assertion for this in order to facilitate finding errors in tests?

paolobazzi (approved):

org.eclipse.scout.rt.platform.test/src/main/java/org/eclipse/scout/rt/testing/platform/mock/RegisterBeanTestRule.java
matthiaso: It depends a little bit what we want/what is our expected behavior how this rule is used.

Maybe we should add a test to define this behavior, I tried the following test on the release branch (w/o my change suggested here):

```java
Index: org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/testing/platform/mock/RegisterBeanTestRuleTest.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/testing/platform/mock/RegisterBeanTestRuleTest.java b/org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/testing/platform/mock/RegisterBeanTestRuleTest.java
new file mode 100644
--- /dev/null	(date 1764248538607)
+++ b/org.eclipse.scout.rt.platform.test/src/test/java/org/eclipse/scout/rt/testing/platform/mock/RegisterBeanTestRuleTest.java	(date 1764248538607)
@@ -0,0 +1,65 @@
+/*
+ *
+ *
+ */
+package org.eclipse.scout.rt.testing.platform.mock;
+
+import org.eclipse.scout.rt.platform.Replace;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+public class RegisterBeanTestRuleTest {
+
+  @Test
+  public void test() {
+    // 1
+    new RegisterBeanTestRule<>(IFixtureBean.class, Mockito.mock(FixtureBeanReplacement.class)).run(() -> {
+          System.out.println(BEANS.get(IFixtureBean.class)); // returns my mock
+          System.out.println(BEANS.get(FixtureBean.class)); // returns NOT my mock -> FixtureBeanReplacement
+          System.out.println(BEANS.get(FixtureBeanReplacement.class));  // returns NOT my mock -> FixtureBeanReplacement
+        }
+    );
+
+    // 2
+    new RegisterBeanTestRule<>(FixtureBean.class, Mockito.mock(FixtureBeanReplacement.class)).run(() -> {
+          System.out.println(BEANS.get(IFixtureBean.class));  // returns NOT my mock -> FixtureBeanReplacement
+          System.out.println(BEANS.get(FixtureBean.class));  // returns NOT my mock -> FixtureBeanReplacement
+          System.out.println(BEANS.get(FixtureBeanReplacement.class)); // returns NOT my mock -> FixtureBeanReplacement
+        }
+    );
+
+    // 3
+    new RegisterBeanTestRule<>(FixtureBeanReplacement.class, Mockito.mock(FixtureBeanReplacement.class)).run(() -> {
+          System.out.println(BEANS.get(IFixtureBean.class));  // returns my mock
+          System.out.println(BEANS.get(FixtureBean.class));  // returns my mock
+          System.out.println(BEANS.get(FixtureBeanReplacement.class));  // returns my mock
+        }
+    );
+
+    // 4
+    System.out.println(BEANS.get(IFixtureBean.class));  // returns FixtureBeanReplacement
+    System.out.println(BEANS.get(FixtureBean.class)); // returns FixtureBeanReplacement
+    System.out.println(BEANS.get(FixtureBeanReplacement.class)); // returns FixtureBeanReplacement
+  }
+
+  @ApplicationScoped
+  public interface IFixtureBean {
+  }
+
+  public static class FixtureBean implements IFixtureBean {
+  }
+
+  @Replace
+  public static class FixtureBeanReplacement extends FixtureBean {
+  }
+}
```

See comments for the current behavior.

I think I ran into case 2 above (my replacement was never returned), that's why I decided to suggest this change. With this change my mock is always returned (expect for case 4 where we do not use any test rule).

However we could also use an assertion like

```java
Assert.assertEquals("Some description what is wrong here and how to fix it", bean.getBeanClazz(), m_beanClazz);
```

With this assertion my case 1 and 2 above would fail immediately - and I would fix it in my test.

However we would have to ensure that case 1 and 2 does not exist in our tests today.

matthiaso: Case 1 is valid to just mock the interface (with a completly different implementation not extending the current impl); however assertion added for case 2/3 - and a warning is logged for case 1 now.

PR #1835, opened 2025-11-25
Table: error with active chart filter when removing column

cguglielmo:
440564


fschinkel (changes requested):
Please explain in the commit message why removing the equals method solves this problem.

PR #1836, opened 2025-11-26
Property for maximum thread pool size of Jetty server

rogersteblerbsi:
* Add new config property 'scout.app.threadPool.maxSize' which controls the maximum number of threads for the Jetty server thread pool
* Add low resource monitor to log a warning if the Jetty server is low on threads

441286

### DIFF
--- a/org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ScoutJettyLowResourceMonitorFactory.java
+++ b/org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ScoutJettyLowResourceMonitorFactory.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.scout.rt.app;
+
+import org.eclipse.jetty.server.LowResourceMonitor;
+import org.eclipse.jetty.server.Server;
+import org.eclipse.scout.rt.platform.ApplicationScoped;
+
+/**
+ * Factory to create a {@link LowResourceMonitor} to monitor the Jetty server for low resources.
+ */
+@ApplicationScoped
+public class ScoutJettyLowResourceMonitorFactory {
+
+  public LowResourceMonitor createLowResourceMonitor(Server server) {
+    LowResourceMonitor lowResourceMonitor = new LowResourceMonitor(server);
+    initLowResourceMonitor(lowResourceMonitor);
+    return lowResourceMonitor;
+  }
+
+  protected void initLowResourceMonitor(LowResourceMonitor lowResourceMonitor) {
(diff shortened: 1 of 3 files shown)


rogersteblerbsi:
@matthiaso Do you think we should also increase the default value of the parameter?

matthiaso:
> @matthiaso Do you think we should also increase the default value of the parameter?

@rogersteblerbsi Not sure, but with techniques like the UI notification long-running polling or the /json polling requests, probably each of these does reserve at least one thread (not sure about requests going through the `HttpProxy` however which is a async request processor, but also these requests need to end somewhere). 

Maybe we could increase it just a little bit (but not too high)? As far as I understand we do have the low resource warning now which would warn us if we reach the limit.

matthiaso:
> @matthiaso Do you think we should also increase the default value of the parameter?

@rogersteblerbsi Do we know if we need one thread per incoming connection? If so, we do have other limits which are much higher, e.g. https://github.com/eclipse-scout/scout.rt/blob/0c0ea0ea8908090082248ea6eeb9a76616aa098b/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/servicetunnel/http/HttpServiceTunnelConfigurationProperties.java#L34

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ScoutJettyLowResourceMonitorFactory.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Factory to create a {@link LowResourceMonitor} to monitor the Jetty server for low resources.
+ */
+@ApplicationScoped
+public class ScoutJettyLowResourceMonitorFactory {
+
+  public LowResourceMonitor createLowResourceMonitor(Server server) {
+    LowResourceMonitor lowResourceMonitor = new LowResourceMonitor(server);
+    initLowResourceMonitor(lowResourceMonitor);
+    return lowResourceMonitor;
```
matthiaso, line 25: Do we need additional configuration? This monitor does not only monitor for low resources, it seems to also decrease idle time (when we are low on resources) and possibly (probably disabled) stop incoming connections, see https://github.com/jetty/jetty.project/blob/jetty-12.1.x/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/LowResourceMonitor.java#L352-L368

Is there anything we need to consider here? I am not even sure what the decreased idle time means for us (as also the default idle time seems to be only 30 seconds).

rogersteblerbsi:
> > @matthiaso Do you think we should also increase the default value of the parameter?
> 
> @rogersteblerbsi Not sure, but with techniques like the UI notification long-running polling or the /json polling requests, probably each of these does reserve at least one thread (not sure about requests going through the `HttpProxy` however which is a async request processor, but also these requests need to end somewhere).
> 
> Maybe we could increase it just a little bit (but not too high)? As far as I understand we do have the low resource warning now which would warn us if we reach the limit.

Quickly discussed with Paolo. We will increase the default value to 2000.

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ScoutJettyLowResourceMonitorFactory.java
rogersteblerbsi: We will add the monitor but disabled by default.

matthiaso (approved):

PR #1841, opened 2025-11-28
Support for cancellation of chunked data requests

paolobazzi:
Improved REST cancellation support for chunked REST requests:
- Client allows to cancel request until response is completely read.
- Server allows cancelling request until response is completely written.
- Added (debug) logging in RunMonitor and REST cancellation support

423389


paolobazzi:
Alternative implementation for OutputStreamRestContainerResponseFilter (should be renamed then):
Additionally RestRequestCancellationServletFilter must be removed then.

```
public class OutputStreamRestContainerResponseFilter implements IRestContainerRequestFilter, IRestContainerResponseFilter {

  private static final Logger LOG = LoggerFactory.getLogger(OutputStreamRestContainerResponseFilter.class);

// ---> additional request filter registering run monitor
  @Override
  public void filter(ContainerRequestContext requestContext) throws IOException {
    registerRunMonitor(requestContext);
  }

  @Override
  public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext) throws IOException {
    String requestId = requestContext.getHeaderString(RestHttpHeaders.REQUEST_ID);
    OutputStream out = responseContext.getEntityStream();
    responseContext.setEntityStream(new P_WrappingOutputStream(out, requestId));
  }

  protected void registerRunMonitor(ContainerRequestContext request) {
    String requestId = request.getHeaderString(RestHttpHeaders.REQUEST_ID);
    if (requestId == null) {
      LOG.trace("cancellation not supported by this request: HTTP header '" + RestHttpHeaders.REQUEST_ID + "' is missing");
      return;
    }

    RunContext runContext = RunContext.CURRENT.get();
    if (runContext == null) {
      LOG.trace("cancellation not supported by this request: not running within a run context");
      return;
    }

    Object userId = resolveUserId(request);
    getCancellationRegistry().register(requestId, userId, runContext);
  }

  /**
   * Returns the user id of the given request. May be {@code null}.
   */
  protected Object resolveUserId(ContainerRequestContext request) {
    return BEANS.get(IAccessControlService.class).getUserIdOfCurrentSubject();
  }

  /**
   * Returns the cancellation registry that manages requests passing this filter (e.g. if different entry points are
   * managed independently). The default is {@link RestRequestCancellationRegistry}.
   */
  protected RestRequestCancellationRegistry getCancellationRegistry() {
    return BEANS.get(RestRequestCancellationRegistry.class);
  }

  /**
   * {@link OutputStream} implementation wrapping an {@link OutputStream} and ignoring connection errors when closing the stream.
   *
   * @see ConnectionErrorDetector#isConnectionError(Throwable)
   */
  protected static class P_WrappingOutputStream extends OutputStream {

    protected final OutputStream m_out;
    protected final String m_requestId;

    P_WrappingOutputStream(OutputStream out, String requestId) {
      m_out = out;
      m_requestId = requestId;
    }

    @Override
    public void write(int b) throws IOException {
      m_out.write(b);
    }

    @Override
    public void write(byte[] b) throws IOException {
      m_out.write(b);
    }

    @Override
    public void write(byte[] b, int off, int len) throws IOException {
      m_out.write(b, off, len);
    }

    @Override
    public void close() throws IOException {
      try {
        m_out.close();
      }
      catch (EOFException e) {
        if (BEANS.get(ConnectionErrorDetector.class).isConnectionError(e)) {
          // Ignore disconnect errors: we do not want to throw an exception, if the client closed the connection.
          LOG.debug("EOFException when closing REST response output stream: ", e);
        }
        else {
          throw e;
        }
      }
      finally {
        // unregister cancellation support after closing output stream (e.g. finished to completely write response)
        if (m_requestId != null) {
          BEANS.get(RestRequestCancellationRegistry.class).unregister(m_requestId);
        }
      }
    }

    @Override
    public void flush() throws IOException {
      m_out.flush();
    }
  }
}
```

org.eclipse.scout.rt.rest.jersey.test/src/main/java/org/eclipse/scout/rt/rest/jersey/fixture/ChunkedDataResource.java
```diff
@@ -148,6 +152,69 @@ public Response getDataObjectsScoutDefaultDelimiterNull() {
...
+        writer.write(createFixtureDo(1));
+        LOG.info("Server: Wrote 1 chunk of data.");
+
+        BEANS.get(ChunkedDataCancellationSynchronizer.class).signalDataWritten();
+        LOG.info("Server: Signaled data written.");
+
+        while (!RunMonitor.CURRENT.get().isCancelled()) {
+          // simulate further writing and wait for being cancelled
+          LOG.info("Server: Waiting for cancellation.");
+          SleepUtil.sleepSafe(100, TimeUnit.MILLISECONDS);
```
andibur: You could use another CountDownLatch and await with timeout instead of the sleep-loop: await(5, TimeUnit.Seconds)

andibur (approved):
Looks good!

org.eclipse.scout.rt.rest.jersey.client/src/main/java/org/eclipse/scout/rt/rest/jersey/client/ScoutApacheConnector.java
```diff
@@ -757,6 +757,8 @@ public void close() throws IOException {
       finally {
         // (3) close response
         m_response.close();
+        // (4) unregister cancellable
+        m_cancellableHandle.dispose();
```
matthiaso: Is there a possibility of `m_response.close()` failing (e.g. EOF or other I/O errors) which would then lead to non-execution of our `m_cancellableHandle.dispose()` method? Would this be a problem?

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/cancellation/RestRequestCancellationRegistry.java
```diff
@@ -46,27 +45,22 @@ protected ConcurrentMap<String, RequestCancellationInfo> getRequestCancellationI
     return m_requestCancellationInfos;
   }
 
-  public IRegistrationHandle register(String requestId, Object userId, RunContext runContext) {
+  /**
+   * Rregister request with id {@code requestId} in registry using given {@code runContext} and {@code userId}.
```
matthiaso:
```suggestion
   * Register request with id {@code requestId} in registry using given {@code runContext} and {@code userId}.
```

Maybe we should also add a description that manual unregistration is possible/necessary?

org.eclipse.scout.rt.rest.jersey.test/src/main/java/org/eclipse/scout/rt/rest/jersey/fixture/ChunkedDataResource.java
paolobazzi: done

paolobazzi:
> Alternative implementation for OutputStreamRestContainerResponseFilter (should be renamed then): Additionally RestRequestCancellationServletFilter must be removed then.
> 
> ```
> public class OutputStreamRestContainerResponseFilter implements IRestContainerRequestFilter, IRestContainerResponseFilter 
> ```

Integrated into newest code

matthiaso (approved):

org.eclipse.scout.rt.rest.jersey.client/src/main/java/org/eclipse/scout/rt/rest/jersey/client/ScoutApacheConnector.java
paolobazzi: Changed to (another) try-finally

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/cancellation/RestRequestCancellationRegistry.java
paolobazzi: Added into javadoc

PR #1843, opened 2025-11-30
Avoid triggering double events with non value chart click

bsifpa:


fschinkel (changes requested):
Please add a small explanation in the commit message why triggering the nonValueClick leads to duplicate events. Please include how to reproduce this behaviour.

fschinkel (approved):

PR #1849, opened 2025-12-04
DesktopTab: Only close closable tabs using menu

Knicksaas:
The "CloseAll" and the "CloseOthers" menu on the DesktopTab is only visible when there is a closable view, attached to the desktop. It only closes the closable views.

420603

### DIFF
--- a/eclipse-scout-core/src/desktop/desktoptab/DesktopTab.ts
+++ b/eclipse-scout-core/src/desktop/desktoptab/DesktopTab.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


eclipse-scout-core/src/desktop/desktoptab/DesktopTab.ts
```diff
@@ -57,14 +58,37 @@ export class DesktopTab extends SimpleTab<Form> {
...
+
+    if (!includeThis) {
+      arrays.remove(closeableForms, this.view);
+    }
+
+    return closeableForms;
+  }
+
+  protected _closeAllMenuVisible(): boolean {
+    // When the desktop has at least one tab which is closeable, the menu is visible.
```
jwinkens, line 79: comment necessary? the other method has no comment, either.

jwinkens (approved):
Looks good to me. Test succesful. Not sure if the comment is really necessary, see review.

eclipse-scout-core/src/desktop/desktoptab/DesktopTab.ts
Knicksaas: This comment comes from a discussion I had. We talked about, if we want a "Close All" menu on a tab, which is not closable. It may irritate the user and make him believe, the tab is closable. But for consistency reasons, we agreed on showing it on any tab, as soon as at least one tab is closable. 

This comment should express that it was a conscious decision to show the menu on any tab, as soon as at least one menu is visible.

PR #1854, opened 2025-12-05
Fix possible ArrayIndexOutOfBoundsException in InspectorInfo

mvilliger:
InspectorInfo used to be an ApplicationScoped bean with a SHA256 digest. This digest was therefore shared for the whole application and used in parallel. But MessageDigest is not thread safe which may lead to AIOOBE.

Now each UiSession gets its own InspectorInfo instance with its own MessageDigest. This is fine as requests of a single UiSession are never processed in parallel (see UiSession.processJsonRequest and JsonMessageRequestHandler.handleJsonRequest).

443078


org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/desktop/JsonOutline.java
```diff
@@ -203,7 +200,8 @@ protected JSONObject treeNodeToJson(ITreeNode node, IChildNodeIndexLookup childI
       json.put(PROP_SHOW_TILE_OVERVIEW, page.isShowTileOverview());
       json.put(PROP_COMPACT_ROOT, page.isCompactRoot());
     }
-    BEANS.get(InspectorInfo.class).put(json, page, getUiSession());
+    IUiSession uiSession = getUiSession();
+    uiSession.getInspectorInfo().put(json, page, uiSession);
```
cguglielmo, line 204: Why extract uiSession as variable if it is used only once?

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/form/fields/wizard/JsonWizardProgressField.java
```diff
@@ -96,7 +94,8 @@ protected JSONObject wizardStepToJson(IWizardStep<? extends IForm> wizardStep) {
     jsonStep.put("actionEnabled", wizardStep.isActionEnabled());
     jsonStep.put("cssClass", wizardStep.getCssClass());
     jsonStep.put("finished", wizardStep.isFinished());
-    BEANS.get(InspectorInfo.class).put(jsonStep, wizardStep, getUiSession());
+    IUiSession uiSession = getUiSession();
+    uiSession.getInspectorInfo().put(jsonStep, wizardStep, uiSession);
```
cguglielmo, line 98: Why extract uiSession as variable if it is used only once?

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/table/JsonColumn.java
```diff
@@ -104,7 +102,8 @@ public JSONObject toJson() {
     json.put("headerTooltipText", column.getHeaderCell().getTooltipText());
     json.put("headerTooltipHtmlEnabled", column.getHeaderCell().isTooltipHtmlEnabled());
     json.put("headerIconId", BinaryResourceUrlUtility.createIconUrl(column.getHeaderCell().getIconId()));
-    BEANS.get(InspectorInfo.class).put(json, column, getUiSession());
+    IUiSession uiSession = getUiSession();
+    uiSession.getInspectorInfo().put(json, column, uiSession);
```
cguglielmo, line 106: Why extract uiSession as variable if it is used only once?

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/desktop/JsonOutline.java
mvilliger: It is used twice on the same line.

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/form/fields/wizard/JsonWizardProgressField.java
mvilliger: It is used twice on the same line.

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/table/JsonColumn.java
mvilliger: It is used twice on the same line.

PR #1856, opened 2025-12-05
DesktopTab: Only close closable tabs using menu

Knicksaas:
The "CloseAll" and the "CloseOthers" menu on the DesktopTab is only visible when there is a closable view, attached to the desktop. It only closes the closable views.

420603


Knicksaas:
Cherry pick from #1849

PR #1864, opened 2025-12-10
PageWithNodes: update row with text from child jsPage

fschinkel:
If the child page of an AbstractPageWithNodes is an IJsPage the text, cssClass, etc. is not known to the UI server when the child page is created. Therefore, send the information about the child page back to the UI server after the child page was initialized.

427023

### DIFF
--- a/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/outline/pages/IPageWithNodes.java
+++ b/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/outline/pages/IPageWithNodes.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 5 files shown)


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/outline/pages/IPageWithNodes.java
```diff
@@ -25,4 +25,9 @@
 public interface IPageWithNodes extends IPage<ITable> {
 
   void rebuildTableInternal();
+
+  /**
+   * Updates the table row with the content of the page .
```
cguglielmo: remove space between page and .

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/outline/pages/IPageWithNodes.java
```diff
@@ -25,4 +25,9 @@
...
+   */
+  void updateTableRowFromPage(IPage page);
```
cguglielmo, line 32: Please update JavaDoc of updateCellFromPageCell. Remove "Called when the table gets rebuilt" because this is not always true anymore. Also remove the _row._ at the end

PR #1865, opened 2025-12-10
Add overviewText and overviewHtmlEnabled to Page

fschinkel:
Introduce the new properties overviewText and overviewHtmlEnabled on the Page. These properties are used by the PageTileButton to display a tile outline overview like overviewIconId. The PageTileButton uses these properties if they are set (i.e. not undefined). If they are not set it text, iconId and htmlEnabled are used to display the tile overview.

427023

### DIFF
--- a/eclipse-scout-core/src/desktop/outline/pages/PageModel.ts
+++ b/eclipse-scout-core/src/desktop/outline/pages/PageModel.ts
@@ -32,9 +32,27 @@ export interface PageModel extends TreeNodeModel, ObjectModelWithUuid<TreeNode> 
    */
   drillDownOnRowClick?: boolean;
   /**
-   * The icon id which is used for icons in the tile outline overview.
+   * Configures the text which is used in the tile outline overview.
+   *
+   * If the value is `undefined`, the {@link PageTileButton} uses the {@link PageModel.text} instead.
+   * If no text should be displayed, set the value to `null`.
+   */
+  overviewText?: string;
+  /**
+   * Configures the icon which is used in the tile outline overview.
+   *
+   * If the value is `undefined`, the {@link PageTileButton} uses the {@link PageModel.iconId} instead.
+   * If no icon should be displayed, set the value to `null`.
    */
   overviewIconId?: string;
+  /**
+   * Configures whether HTML code in the {@link PageModel.overviewText} property should be interpreted. If set to false, the HTML will be encoded.
+   *
+   * If the value is `undefined`, the {@link PageTileButton} uses {@link PageModel.htmlEnabled} instead.
+   *
+   * Default is false.
+   */
+  overviewHtmlEnabled?: boolean;
   showTileOverview?: boolean;
   /**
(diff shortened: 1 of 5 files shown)


eclipse-scout-core/src/desktop/outline/pages/PageModel.ts
```diff
@@ -31,10 +31,18 @@ export interface PageModel extends TreeNodeModel, ObjectModelWithUuid<TreeNode>
    * True to select the page linked with the selected row when the row was selected. May be useful on touch devices.
    */
   drillDownOnRowClick?: boolean;
+  /**
+   * The text which is used in the tile outline overview.
```
cguglielmo:
```suggestion
   * Configures the text which is used in the tile outline overview.
   * 
   * If the value is `undefined`, the {@link PageTileButton} uses the {@link text} instead.
   * If no text should be displayed, set the value to `null`.
```

eclipse-scout-core/src/desktop/outline/pages/PageModel.ts
```diff
@@ -31,10 +31,18 @@ export interface PageModel extends TreeNodeModel, ObjectModelWithUuid<TreeNode>
...
+   */
+  overviewText?: string;
   /**
    * The icon id which is used for icons in the tile outline overview.
```
cguglielmo: Adjust js doc as for the text

eclipse-scout-core/src/desktop/outline/pages/PageModel.ts
```diff
@@ -31,10 +31,18 @@ export interface PageModel extends TreeNodeModel, ObjectModelWithUuid<TreeNode>
...
    */
   overviewIconId?: string;
+  /**
+   * The htmlEnabled which is used in the tile outline overview.
```
cguglielmo:
```suggestion
   * Configures whether HTML code in the {@link overviewText} property should be interpreted. If set to false, the HTML will be encoded.
   * 
   * If the value is `undefined`, the {@link PageTileButton} uses {@link htmlEnabled} instead.
   *
   * Default is false.
```

eclipse-scout-core/src/desktop/outline/pages/PageTileButton.ts
```diff
@@ -24,18 +24,39 @@ export class PageTileButton extends TileButton implements PageTileButtonModel {
...
+      return this.page.overviewText;
+    }
+    return this.page?.text;
+  }
+
+  protected get _iconId(): string {
+    if (this.page?.overviewIconId !== undefined) {
+      return this.page.overviewIconId;
+    }
+    return this.page?.iconId;
```
cguglielmo, line 53: This changes the behavior for Java pages. It is not possible anymore to _not_ have an overviewIcon. Maybe set overviewIcon to null in OutlineAdapter?

cguglielmo (commented):
Commit message has a type: If they are not set _it_ text

PR #1875, opened 2025-12-16
SearchOutline

fschinkel:
In order to support hybrid or js only search pages move the search-logic to SearchOutline.ts. All pages added to AbstractSearchOutline in Java need to implement ISearchPage and provide an ISearchState containing the information about the search results of this page. After the search of a page was completed it must update the ISearchState with information about the resultCount and whether the page is limited. In addition, the ISearchState must be marked as pending=false.
This information is collected by the SearchOutline.ts and displayed in the status of the outline.

427023


eclipse-scout-core/src/desktop/outline/SearchOutline.ts
```diff
@@ -120,29 +166,82 @@ export class SearchOutline extends Outline implements SearchOutlineModel {
     this.$searchPanel.toggleClass('has-status', !!this.searchStatus);
   }
 
-  focusQueryField() {
-    this.validateFocus();
+  setMaxSearchFieldLength(maxSearchFieldLength: number) {
+    if (maxSearchFieldLength <= 0) {
```
cguglielmo, line 175: This won't wok if property is set in init model -> also call in _init

eclipse-scout-core/src/desktop/outline/SearchOutline.ts
```diff
@@ -120,29 +166,82 @@ export class SearchOutline extends Outline implements SearchOutlineModel {
...
+      maxSearchFieldLength = 200;
+    }
+    this.setProperty('maxSearchFieldLength', maxSearchFieldLength);
+  }
+
+  setMinSearchTokenLength(minSearchTokenLength: number) {
```
cguglielmo, line 181: This won't wok if property is set in init model -> also call in _init

eclipse-scout-core/src/desktop/outline/SearchOutline.ts
```diff
@@ -120,29 +166,82 @@ export class SearchOutline extends Outline implements SearchOutlineModel {
...
+    if (this.searchQuery.length > this.maxSearchFieldLength) {
+      this.setSearchStatus(this.session.text('ui.SearchTermTooLong'));
+      this._searchQueryValid = false;
+      return;
+    }
+
+    // remove wildcards and split into tokens (i.e. non-empty strings without spaces)
+    const tokens = this.searchQuery.replaceAll(/\*/g, '').split(' ').filter(Boolean);
+
+    // at least one token must hav min length
```
cguglielmo: hav -> have

eclipse-scout-core/src/desktop/outline/SearchOutline.ts
```diff
@@ -7,30 +7,47 @@
...
-  constructor() {
-    super();
-    this.hasText = false;
-    this.$searchPanel = null;
-    this.$clearIcon = null;
-    this.$searchStatus = null;
-    this.$queryField = null;
+  protected _searchQueryValid = true;
+  protected _searchStates = new Set<SearchState>();
+  protected _searchStateResultCountChangedHandler = this._onSearchStateResultCountChanged.bind(this);
```
cguglielmo: Changed -> Change. Because event is called propertyChange

eclipse-scout-core/src/desktop/outline/SearchOutline.ts
```diff
@@ -186,4 +270,133 @@ export class SearchOutline extends Outline implements SearchOutlineModel {
...
+    if (!searchState) {
+      return;
+    }
+    searchState.off('propertyChange:resultCount', this._searchStateResultCountChangedHandler);
+    searchState.off('propertyChange:limited', this._searchStateLimitedChangedHandler);
+    searchState.off('propertyChange:pending', this._searchStatePendingChangedHandler);
+    searchState.off('destroy', this._searchStateDestroyHandler);
+  }
+
+  protected _onSearchStateResultCountChanged(event: PropertyChangeEvent<number>) {
```
cguglielmo: Maybe just use one handler onSearchStateChange instead of 3 because they all do the same?

eclipse-scout-core/src/desktop/outline/SearchOutline.ts
```diff
@@ -186,4 +270,133 @@ export class SearchOutline extends Outline implements SearchOutlineModel {
...
+  get resultCount(): number {
+    let resultCount = 0;
+    for (const searchState of this._searchStates) {
+      resultCount += searchState.resultCount;
+    }
+    return resultCount;
+  }
+
+  get limited(): boolean {
+    for (const searchState of this._searchStates) {
```
cguglielmo: Or if you prefer it more: return Array.from(this._searchStates).some(state => state.limited); For me, both is fine

eclipse-scout-core/src/desktop/outline/SearchOutline.ts
```diff
@@ -186,4 +270,133 @@ export class SearchOutline extends Outline implements SearchOutlineModel {
...
+      if (searchState.limited) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  get pending(): boolean {
+    for (const searchState of this._searchStates) {
```
cguglielmo: See above

eclipse-scout-core/src/desktop/outline/SearchOutline.ts
```diff
@@ -186,4 +270,133 @@ export class SearchOutline extends Outline implements SearchOutlineModel {
...
+        continue;
+      }
+      searchStatesToRemove.add(searchState);
+    }
+
+    searchStatesToAdd.forEach(searchState => this._addSearchState(searchState));
+    searchStatesToRemove.forEach(searchState => this._removeSearchState(searchState));
+  }
+
+  protected _addSearchState(searchState: SearchState) {
```
cguglielmo: I would integrate add and removeSearchState into setSearchStates so updateSearchStates is called only once. onSearchStateDestroy can call a public removeSearchState method that copies the existing states, remove the given state and calls setSearchStates()

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/outline/SearchState.java
```diff
@@ -0,0 +1,99 @@
...
+package org.eclipse.scout.rt.client.ui.desktop.outline;
+
+import org.eclipse.scout.rt.client.ModelContextProxy;
+import org.eclipse.scout.rt.client.ModelContextProxy.ModelContext;
+import org.eclipse.scout.rt.client.ui.AbstractWidget;
+import org.eclipse.scout.rt.platform.BEANS;
+import org.eclipse.scout.rt.platform.classid.ClassId;
+
+@ClassId("4b28a0dc-1121-4ad5-b1f8-899cacca7a02")
+public class SearchState extends AbstractWidget implements ISearchState {
```
cguglielmo: extending from AbstractPropertyObserver would probably be sufficient. This would also allow to make it a @Bean, at least existing code already does it, see PopupManager

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/desktop/JsonSearchOutline.java
```diff
@@ -54,17 +135,45 @@ protected String modelValue() {
...
+  }
+
+  protected void handleUiResetSearch(JsonEvent event) {
+    getModel().getUIFacade().fireResetSearchFromUI();
+  }
+
+  @Override
+  protected void handleUiPropertyChange(String propertyName, JSONObject data) {
+    switch (propertyName) {
+      case ISearchOutline.PROP_SEARCH_QUERY -> {
```
cguglielmo, line 156: The new switch expression looks nice if every case has exactly one line as in handleUiEvent, but here I don't see any benefits. It would stick to the existing pattern to keep it consistent with the other json adapters.

PR #1558, opened 2025-06-13
Define a maximum runtime for test methods

matthiaso:
422336

PR #1876, opened 2025-12-17
Define a maximum runtime for test methods

matthiaso:
Runtime is checked after execution has finished and if maximum time has been exceeded test fails with an error. No actual timeout is defined as timeouts would result in a behavior change for several tests (methods would be executed in different threads); however with the maximum runtime we should still be able to identify newly long-running tests faster.

422336


matthiaso:
Replaces https://github.com/eclipse-scout/scout.rt/pull/1558

PR #1882, opened 2025-12-22
Table: Fix scrollToSelection on tables with async validating rows

Knicksaas:
When a table has at least one row that validates its value async, the scrollToSelection flag did not have an effect on newly inserted rows. This is because the insert event gets packed into the table buffer, and the rendering of the viewport is paused. At the end of the pause, the selection was not rendered. This got fixed with this change.

422941

### DIFF
--- a/eclipse-scout-core/src/table/TableUpdateBuffer.ts
+++ b/eclipse-scout-core/src/table/TableUpdateBuffer.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


eclipse-scout-core/src/table/TableUpdateBuffer.ts
```diff
@@ -85,6 +85,10 @@ export class TableUpdateBuffer extends EventEmitter {
     if (this.table._isDataRendered()) {
       this.table._renderViewport();
     }
+    // Update scroll position after viewport rendering
```
dwl-bsi: Should the scroll position only be updated if the viewport was actually rendered? Or should it be updated every time?
I.e., should this code block be moved into the `if(this.table._isDataRendered())` block above?
If not, then I would change this comment because right now it sounds like this should only happen if the viewport was rendered.

Knicksaas: Good point, it makes more sense to move it into the if-block above.

PR #1888, opened 2026-01-07
Add SLF4JBridgeHandler for tests

rogersteblerbsi:
Adds SLF4JBridgeHandler in logback configuration for tests to ensure that log messages are redirected to SLF4J, providing consistent logging

445808


rogersteblerbsi:
~Looks like we can't just add it to the general logback-test-scout.xml. Requires the jul-to-slf4j dependency.~

Edit: Added check if the optional jul-to-slf4j dependency is available.

andibur (approved):

PR #1900, opened 2026-01-08
Add WAV MimeType

michaelze:
This change introduces the new MimeType `WAV` with content type string
"audio/wav" and file extension "wav".

As checking the magic bytes of wav blobs involves skipping byte 5-8,
which was previously not possible with the IMimeMagic pattern matching,
this change also extends the magic bytes pattern matching mechanism to
allow for skipping bytes while checking the magic byte pattern.
Checking wav blob magic bytes makes use of this new enhancement.

### DIFF
--- a/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
+++ b/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -9,8 +9,11 @@
  */
 package org.eclipse.scout.rt.platform.resource;
 
+import java.util.Arrays;
+
+import org.eclipse.scout.rt.platform.exception.ProcessingException;
 import org.eclipse.scout.rt.platform.security.MalwareScanner;
-import org.eclipse.scout.rt.platform.util.HexUtility;
+import org.eclipse.scout.rt.platform.util.Assertions;
 
 /**
  * Typically used in combination with {@link MalwareScanner}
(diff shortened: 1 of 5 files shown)


org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
```diff
@@ -63,45 +72,137 @@ default boolean matches(BinaryResource res) {
...
   static IMimeMagic createMagic(int pos, String... hexMagics) {
-    byte[][] magics = new byte[hexMagics.length][];
-    int maxLen = 0; // longest magic marker
-    for (int i = 0; i < hexMagics.length; i++) {
-      magics[i] = HexUtility.decode(hexMagics[i]);
-      maxLen = Math.max(maxLen, magics[i].length);
-    }
-    int magicsMaxLen = maxLen;
+    return createMagic(Arrays.stream(hexMagics)
+        .map(hexMagic -> ImmutablePair.of(pos, hexMagic))
```
mvilliger: Why use Pair here? We could also directly use BytePattern?

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
```diff
@@ -63,45 +72,137 @@ default boolean matches(BinaryResource res) {
...
+  }
+
+  static IMimeMagic createMagic(List<Pair<Integer, String>> patterns) {
+    Assertions.assertTrue(patterns.size() > 0, "At least one pattern must be provided");
+
+    BytePattern[] magics = patterns.stream()
+        .map(pattern -> new BytePattern(pattern.getLeft(), pattern.getRight()))
+        .toArray(BytePattern[]::new);
+    int length = Arrays.stream(magics)
+        .map(bytePattern -> bytePattern.getPos() + bytePattern.getLength())
```
mvilliger: Use mapToInt to prevent auto boxing

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
```diff
@@ -63,45 +72,137 @@ default boolean matches(BinaryResource res) {
...
+}
 
-      private boolean matchesMagic(byte[] content, byte[] magic) {
-        for (int i = 0; i < magic.length; i++) {
-          if (content[pos + i] != magic[i]) {
-            return false;
-          }
-        }
-        return true;
+class BytePattern {
```
mvilliger: Maybe use a more specific name like e.g. "MagicBytePattern" or something similar?

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
```diff
@@ -63,45 +72,137 @@ default boolean matches(BinaryResource res) {
...
-    };
+      if (content[m_pos + i] != m_bytes[i]) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  // mainly used for testing
+  byte[] getBytes() {
```
mvilliger: If these methods exist, maybe also use them instead of directly accessing the member?

mvilliger (changes requested):

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
michaelze: I used Pair, because using `BytePattern` would expose its name as part of the public API of `IMimeMagic` (Warning). On second thought, that's probably not a bad thing, because creating `IMimeMagic` instances through the `createMagic()` method from outside Scout's code is intended. I'll move the `BytePattern` class to its own file, make it `public` and use it in the `createMagic()` method.

PR #1905, opened 2026-01-09
Enable JDK25 compact headers JVM flag

paolobazzi:
Added "-XX:+UseCompactObjectHeaders" to launch configs for apps and test


matthiaso (changes requested):
What about launch/run configurations in https://github.com/eclipse-scout/scout.rt/blob/7f6f3e56db9096491f7c1ec40cdeb3bde5333fa1/scout-hellojs-app/src/main/resources/archetype-resources/__rootArtifactId__/run-configs - shouldn't we add the option there as well?

matthiaso (approved):

PR #1916, opened 2026-01-13
ApiDocGenerator: Add HTTP PATCH method

matthiaso:
446509


matthiaso:
We could think about adding custom colors for the method as well, see https://github.com/eclipse-scout/scout.rt/blob/d221672e70c72d7450ae6cb4ced3b2e7bd075838/org.eclipse.scout.rt.rest/src/main/resources/org/eclipse/scout/rt/rest/doc/res/doc.css#L109-L134

But as this method is not as common - and other methods are also not colored as well, I would skip this for now.

paolobazzi (approved):

paolobazzi:
> But as this method is not as common - and other methods are also not colored as well, I would skip this for now.
Skipping is OK.

PR #1919, opened 2026-01-13
CI: Use Java 25 image

matthiaso:
422904


matthiaso:
Yet another place which needed to be adjusted for Java 25 - I believe we should use the same image here which we are also using for continuous jobs on Jenkins (after seeing pull requests with failing tests because of old image).

rogersteblerbsi (approved):

PR #1920, opened 2026-01-14
DoSet, DoCollection: Fix equals if nested data structure is manipulated

paolobazzi:
- Implement equals for DoSet and DoCollection supporting changes in nested data objects which causes corruption of internal hashcode-based data structure.
- DataObjectHelper recreate internal structure for set-attributes when cleaning data object

446243

### DIFF
--- a/org.eclipse.scout.rt.dataobject.test/src/test/java/org/eclipse/scout/rt/dataobject/DoCollectionTest.java
+++ b/org.eclipse.scout.rt.dataobject.test/src/test/java/org/eclipse/scout/rt/dataobject/DoCollectionTest.java
@@ -23,6 +23,8 @@ import java.util.Set;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
+import org.eclipse.scout.rt.dataobject.DataObjectHelperTest.FixtureDoEntity;
+import org.eclipse.scout.rt.platform.BEANS;
 import org.eclipse.scout.rt.platform.util.CollectionUtility;
 import org.junit.Before;
 import org.junit.Test;
@@ -434,6 +436,7 @@ public class DoCollectionTest {
 
     assertEquals(collection1, collection2); // order of elements is not relevant for equality
     assertEquals(collection2, collection1);
+    //noinspection EqualsWithItself
     assertEquals(collection1, collection1);
     assertEquals(collection1.hashCode(), collection2.hashCode());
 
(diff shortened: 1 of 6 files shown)


org.eclipse.scout.rt.dataobject.test/src/test/java/org/eclipse/scout/rt/dataobject/DoCollectionTest.java
```diff
@@ -445,9 +448,33 @@ public void testEqualsHashCode() {
...
+
+    assertEquals(col1, col2);
+    assertEquals(col1.hashCode(), col2.hashCode());
+
+    // manipulate nested dataobject, this causes the hashCode of nested object to change which corrupts internal LinkedHashSet structure
+    col1.iterator().next().withName(null);
+    col2.iterator().next().withName(null);
+
+    assertEquals(col1, col2);
+    assertEquals(col1.hashCode(), col2.hashCode());
```
matthiaso, line 502: Even though you did not change https://github.com/eclipse-scout/scout.rt/blob/0c0ea0ea8908090082248ea6eeb9a76616aa098b/org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/DoCollection.java#L60-L75:

So `col1` and `col2` are equal (as we just asserted this).

In this case I would expect that I can find every object contained in `col1` also in `col2` (and vice versa) - however the following statements seem to be `false`:

```java
col1.contains(col2.iterator().next());
col2.contains(col1.iterator().next());
```

Is this really the intended behavior?

org.eclipse.scout.rt.dataobject.test/src/test/java/org/eclipse/scout/rt/dataobject/DoSetTest.java
```diff
@@ -382,9 +385,33 @@ public void testEqualsHashCode() {
...
+
+    assertEquals(set1, set2);
+    assertEquals(set1.hashCode(), set2.hashCode());
+
+    // manipulate nested dataobject, this causes the hashCode of nested object to change which corrupts internal LinkedHashSet structure
+    set1.iterator().next().withName(null);
+    set2.iterator().next().withName(null);
+
+    assertEquals(set1, set2);
+    assertEquals(set1.hashCode(), set2.hashCode());
```
matthiaso, line 411: So `set1` and `set2` are equal (as we just asserted this).

In this case I would expect that I can find every object contained in `set1` also in `set2` (and vice versa) - however the following statements seem to be false:

```java
set1.contains(set2.iterator().next());
set2.contains(set1.iterator().next());
```

Is this really the intended behavior?

I can even add the element from `set2` to `set1`:

```java
set1.add(set2.iterator().next());
```

Now `set1` has two elements (even though I just added an element from `set2` which was supposed to be equal with `set1`).

org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/DoSet.java
```diff
@@ -53,7 +56,20 @@ public void set(Set<V> newValue) {
...
+   * which uses containsAll-based equality based on hashCode of contained objects which may be corrupted by mutating nested data objects.
+   */
+  @Override
+  protected boolean valueEquals(DoNode other) {
+    if (!exists() && !other.exists()) {
+      return true;
+    }
+    //noinspection unchecked
+    return CollectionUtility.equalsCollection(get(), (Collection<V>) other.get(), false);
+  }
```
matthiaso, line 89: This code is already used for https://github.com/eclipse-scout/scout.rt/blob/0c0ea0ea8908090082248ea6eeb9a76616aa098b/org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/DoCollection.java#L60-L75 so it must be alright; I am actually quite suprised about the code used there (see other comments) - this makes our `DoCollection` and `DoSet` not just wrapper classes for their regular counterparts anymore.

If we really keep this code here (and the already existing code for `DoCollection`) maybe we should adjust the classes javadoc?

org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/DoSet.java
```diff
@@ -53,7 +56,20 @@ public void set(Set<V> newValue) {
     super.set(emptySetIfNull(newValue));
   }
 
-  // LinkedHashSet already implemented hashCode/equals without considering element position, thus no need to override valueHashCode/valueEquals.
+  // LinkedHashSet already implements hashCode without considering element position, thus no need to override valueHashCode/valueEquals.
```
matthiaso:
```suggestion
  // LinkedHashSet already implements hashCode without considering element position, thus no need to override valueHashCode.
```

At first I was suprised that you did not overwrite `valueHashCode` but it seems to be alright as the `AbstractSet.hashCode()` function always seems to use the actual current objects to calculate the `hashCode`.

org.eclipse.scout.rt.dataobject.test/src/test/java/org/eclipse/scout/rt/dataobject/DoCollectionTest.java
paolobazzi: Not really intended. See patched #contains implementation.

org.eclipse.scout.rt.dataobject.test/src/test/java/org/eclipse/scout/rt/dataobject/DoSetTest.java
paolobazzi: Not really intended. See patched #contains implementation.

org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/DoSet.java
paolobazzi: Updated javadoc for DoSet and DoCollection

matthiaso (approved):

PR #1924, opened 2026-01-15
ObjectExtensions: Reduce memory usage

matthiaso:
447011

### DIFF
--- a/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/ExtensionStack.java
+++ b/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/ExtensionStack.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -9,7 +9,6 @@
  */
 package org.eclipse.scout.rt.shared.extension;
 
-import java.util.ArrayList;
 import java.util.Deque;
 import java.util.LinkedList;
 import java.util.List;
@@ -54,10 +53,6 @@ public class ExtensionStack {
     return m_extensions.isEmpty();
   }
 
-  /**
-   * @param declaringClass
-   * @return
-   */
   public Object findContextObjectByClass(Class<?> declaringClass) {
     for (List<? extends IExtension<?>> segment : m_extensions) {
       for (IExtension<?> modelExtension : segment) {
@@ -69,7 +64,10 @@ public class ExtensionStack {
     return null;
   }
 
+  /**
+   * Create an unmodifiable snapshot of the current state which can be used to re-create an equal, modifiable {@link ExtensionStack} using {@link ExtensionStack#ExtensionStack(List)}.
+   */
   public List<List<? extends IExtension<?>>> snapshot() {
-    return new ArrayList<>(m_extensions);
+    return List.copyOf(m_extensions);
   }
 }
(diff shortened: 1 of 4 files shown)


org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/ExtensionStack.java
```diff
@@ -69,7 +64,10 @@ public Object findContextObjectByClass(Class<?> declaringClass) {
     return null;
   }
 
+  /**
+   * Create a unmodifiable snapshot of the current state which can be used to re-create an equal, modifiable {@link ExtensionStack} using {@link ExtensionStack#ExtensionStack(List)}.
```
paolobazzi:
```suggestion
   * Create an unmodifiable snapshot of the current state which can be used to re-create an equal, modifiable {@link ExtensionStack} using {@link ExtensionStack#ExtensionStack(List)}.
```

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/ObjectExtensions.java
```diff
@@ -119,6 +119,18 @@ private List<EXTENSION> loadExtensions(EXTENSION localExtension) {
     if (extensions.isEmpty()) {
       return Collections.singletonList(localExtension);
     }
+    else if (extensions.size() == 1) {
```
paolobazzi, line 123: add javadoc explaining why this size-dependent impl. are used

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/ObjectExtensions.java
```diff
@@ -119,6 +119,18 @@ private List<EXTENSION> loadExtensions(EXTENSION localExtension) {
...
+      return List.of(extensions.get(0), localExtension);
```
paolobazzi, line 124: those list implementations do not support null element - did you check that org.eclipse.scout.rt.shared.extension.IInternalExtensionRegistry#createExtensionsFor returns lists without null elements?

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/ObjectExtensions.java
paolobazzi: maybe add unit test for ObjectExtensions.initConfig ensuring that special list instances are used?

paolobazzi (commented):
Looks good, see inputs for minor improvements.

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/ObjectExtensions.java
matthiaso: As far as I understand no null items should be contained here, especially null items are also conisdered invalid here: https://github.com/eclipse-scout/scout.rt/blob/53bd992e693d584c0a1e041cca33653769ddbffe/org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/ExtensionRegistryItem.java#L74-L76

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/ObjectExtensions.java
matthiaso: Added a comment

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/extension/ObjectExtensions.java
matthiaso: I added a new ObjectExtensionsTest class, what do you think about it?

paolobazzi (approved):

PR #1926, opened 2026-01-16
OpenTelemetry shutdown hook fix

matthiaso:
Only the last registered bean implementing IPlatformListener directly is executed on shutdown. See BeanHierarchy#sortedBeanCopy. The application scoped OpenTelemetryInitializer is now shut down by a static platform listener.

445822


matthiaso:
See https://github.com/eclipse-scout/scout.rt/commit/cb3f972060b25d6fed32e92a6fe5558c881ea28d

PR #1931, opened 2026-01-19
Allow reading of boundary string of REST multipart messages

MartinSoemerBSI:
When sending REST multipart messages, the boundary identifier has to be included in the Content-Type header.
Using BEANS.get(MultipartMessage.class) an identifier will be set and the Content-Type header can be set like:
"multipart/form-data; boundary=" + multipartMessage.getBoundary();


torstentrompler (approved):
looks good to me 👍

PR #1953, opened 2026-01-28
SearchOutline: fix updateSearchStates for mobile

fschinkel:
The SearchOutline collects searchStates from relevant pages. If the outline is compact and has a compactRootNode these relevant pages are the child pages of the compactRootNode.


bschwarzent (approved):
LGTM

PR #1957, opened 2026-01-30
JSONObject - Prevent unnecessary escaping of '/'

rogersteblerbsi:
According to RFC 4627 it's optional to escape the '/' character in JSON. For historical reasons '/' still gets escaped following a '<' to prevent accidental HTML tag termination.

372618

### DIFF
--- a/org.eclipse.scout.json/src/main/java/org/json/JSONStringer.java
+++ b/org.eclipse.scout.json/src/main/java/org/json/JSONStringer.java
@@ -30,6 +30,7 @@ import java.util.List;
  * Changes to the original code:
  * -----------------------------
  * - Applied Scout code formatting rules
+ * - Prevent unnecessary escaping of '/'
  *
  * Copyright (c) 2015 BSI Business Systems Integration AG.
  */
@@ -311,7 +312,6 @@ public class JSONStringer {
       switch (c) {
         case '"':
         case '\\':
-        case '/':
           out.append('\\').append(c);
           break;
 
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.json/src/main/java/org/json/JSONStringer.java
```diff
@@ -311,10 +312,16 @@ private void string(String value) {
       switch (c) {
         case '"':
         case '\\':
-        case '/':
           out.append('\\').append(c);
           break;
 
+        case '/':
+          if (i > 0 && value.charAt(i - 1) == '<') {
+            out.append('\\'); // escape only if previous char is '<'
```
bschwarzent: I'm not sure that preserving the escaping in this special case is beneficial or even necessary. The supposed use case (embedding a JSON literal inside a <script> tag) feels quite niche and somewhat outdated. None of the modern JSON libraries (Jackson, JavaScript, Python, jq etc.)h andle this in the same way.

I would suggest aligning the behavior with Jackson, as demonstrated by this unit test:

```
  @Test
  public void testEscapeSash() {
    // Jackson
    DoEntity o1 = BEANS.get(DoEntity.class);
    o1.put("a1", "https://www.example.org/data?q=123");
    o1.put("a2", "<b>bold</b>");
    o1.put("a3", "c:\\temp\\readme.txt");
    o1.put("a4", "hello \"world\"");
    String s1 = BEANS.get(IDataObjectMapper.class).writeValue(o1);

    // org.json
    JSONObject o2 = new JSONObject();
    o2.put("a1", "https://www.example.org/data?q=123");
    o2.put("a2", "<b>bold</b>");
    o2.put("a3", "c:\\temp\\readme.txt");
    o2.put("a4", "hello \"world\"");
    String s2 = o2.toString();

    assertEquals("{\"a1\":\"https://www.example.org/data?q=123\",\"a2\":\"<b>bold</b>\",\"a3\":\"c:\\\\temp\\\\readme.txt\",\"a4\":\"hello \\\"world\\\"\"}", s2);
    assertEquals(s1, s2);
  }
```

bschwarzent (changes requested):

bschwarzent (approved):

PR #1961, opened 2026-01-30
FileChooserField: propagate properties to FileInput

fschinkel:
The properties "acceptTypes", "displayText" and "maximumUploadSize" of the FileChooserField are propagated to the inner FileInput. This propagation was previously done in the setters after the default implementation has completed. In the meantime the changed properties are e.g. rendered.
Rendering the "displayText" property updates the "hasText" property of a value field and for a FileChooserField the "hasText" property is set to true iff the fileInputs property "text" is truthy. Therefore, the propagation of the "displayText" property to the fileInput needs to be done before the property is rendered. This is achieved by changing the propagation to a property change listener as the listener is executed directly after the property was changed and before the property is rendered.

435099

### DIFF
--- a/eclipse-scout-core/src/form/fields/filechooserfield/FileChooserField.ts
+++ b/eclipse-scout-core/src/form/fields/filechooserfield/FileChooserField.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/src/form/fields/filechooserfield/FileChooserField.ts
```diff
@@ -30,12 +30,17 @@ export class FileChooserField extends ValueField<File> implements FileChooserFie
     super._init(model);
 
     this.fileInput.on('change', this._onFileChange.bind(this));
-    this.on('propertyChange', event => {
-      if (event.propertyName === 'enabledComputed') {
-        // Propagate "enabledComputed" to inner widget
-        this.fileInput.setEnabled(event.newValue);
+    // Propagate properties "acceptTypes", "displayText", "enabledComputed" and "maximumUploadSize"to inner widget
```
cguglielmo: missing space after maximumUploadSize

cguglielmo (commented):
The commit message does not really explain the actual problem. The problem is that hasText property is not updated when changing using setValue() (or setDisplayText()), so the clear icon is missing.

I would change at least the title to for example: FileChooserField: ensure hasText property is updated when changing display text 

Also, there is a type in the message: iff

PR #1965, opened 2026-02-02
Features/cgu/26.1/443405 tree perf

cguglielmo:


eclipse-scout-core/src/tree/Tree.ts
```diff
@@ -2522,11 +2528,11 @@ export class Tree extends Widget implements TreeModel, Filterable<TreeNode> {
       }
       this._destroyTreeNode(node);
       deletedNodes.push(node);
-      this._updateMarkChildrenChecked(node);
 
       // remove children from node map
       Tree.visitNodes(this._destroyTreeNode.bind(this), node.childNodes);
-    });
+    }
+    this._updateChildrenChecked(nodes);
```
fschinkel, line 2534: Why do you call `_updateChildrenChecked` for the nodes that are deleted? The child nodes will be skipped, because of the destroyed-check in `_updateChildrenChecked`, but common parent nodes will be processed multiple times. I think `_updateChildrenChecked` needs to be called with a Set of parentNodes.

eclipse-scout-core/src/tree/Tree.ts
```diff
@@ -2576,8 +2582,8 @@ export class Tree extends Widget implements TreeModel, Filterable<TreeNode> {
       nodes = this.nodes;
       this.nodes = [];
     }
-    Tree.visitNodes(updateNodeMap.bind(this), nodes);
-
+    Tree.visitNodes(node => this._destroyTreeNode(node), nodes);
+    this._updateChildrenChecked(nodes);
     this.deselectNodes(nodes, {collectChildren: true});
```
fschinkel, line 2585: See comment in `deleteNodes`. Only call for parentNode?

eclipse-scout-core/src/util/arrays.ts
```diff
@@ -308,9 +315,19 @@ export const arrays = {
     return 0;
   },
 
+  /**
+   * Pushes all elements of `arr2` to `arr`.
+   *
+   * The implementation uses a loop to push the elements rather than `Array.apply` or the spread operator so it can handle very large arrays
+   * (Chrome throws a Maximum call stack size exceeded when using spread operator at around 125'000 items, Firefox can handle around 500'000).
+   *
+   * In terms of performance, using a a loop is about the same as with the spread operator.
```
fschinkel: using a a loop

eclipse-scout-core/src/util/arrays.ts
```diff
@@ -308,9 +315,19 @@ export const arrays = {
...
+   */
   pushAll<T>(arr: T[], arr2: T | T[]) {
     let a = arrays.ensure(arr2);
```
fschinkel: consider `arr` and `arr2` being the same array, same for insertAll

eclipse-scout-core/src/tree/Tree.ts
cguglielmo: updateChildrenChecked now only processes the same parentNodes once.

eclipse-scout-core/src/tree/Tree.ts
cguglielmo: See comment in delete nodes

eclipse-scout-core/src/util/arrays.ts
cguglielmo: pushAll and insertAll can now handle this case

fschinkel (approved):

PR #1973, opened 2026-02-03
Jetty: Enable graceful shutdown

matthiaso:
431595

### DIFF
--- a/org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/Application.java
+++ b/org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/Application.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -50,6 +50,7 @@ import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
 import org.eclipse.jetty.server.SslConnectionFactory;
 import org.eclipse.jetty.server.handler.ContextHandler;
+import org.eclipse.jetty.server.handler.GracefulHandler;
 import org.eclipse.jetty.util.Callback;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 import org.eclipse.jetty.util.thread.QueuedThreadPool;
@@ -58,6 +59,7 @@ import org.eclipse.scout.rt.app.ApplicationProperties.ScoutApplicationCertificat
 import org.eclipse.scout.rt.app.ApplicationProperties.ScoutApplicationConsoleInputHandlerEnabledProperty;
 import org.eclipse.scout.rt.app.ApplicationProperties.ScoutApplicationContextHandlerExtendedResourceLookup;
 import org.eclipse.scout.rt.app.ApplicationProperties.ScoutApplicationContextPathProperty;
+import org.eclipse.scout.rt.app.ApplicationProperties.ScoutApplicationGracefulShutdownProperty;
 import org.eclipse.scout.rt.app.ApplicationProperties.ScoutApplicationGzipEnabled;
 import org.eclipse.scout.rt.app.ApplicationProperties.ScoutApplicationGzipExcludedInflatePaths;
 import org.eclipse.scout.rt.app.ApplicationProperties.ScoutApplicationGzipExcludedMethods;
@@ -83,6 +85,7 @@ import org.eclipse.scout.rt.app.ApplicationProperties.ScoutApplicationSessionCoo
 import org.eclipse.scout.rt.app.ApplicationProperties.ScoutApplicationSessionCookieConfigSameSiteProperty;
 import org.eclipse.scout.rt.app.ApplicationProperties.ScoutApplicationSessionCookieConfigSecureProperty;
 import org.eclipse.scout.rt.app.ApplicationProperties.ScoutApplicationSessionTimeoutProperty;
+import org.eclipse.scout.rt.app.ApplicationProperties.ScoutApplicationStopTimeoutProperty;
 import org.eclipse.scout.rt.app.ApplicationProperties.ScoutApplicationThreadPoolMaxSizeProperty;
 import org.eclipse.scout.rt.app.ApplicationProperties.ScoutApplicationUseTlsProperty;
 import org.eclipse.scout.rt.app.handler.ScoutJettyErrorHandler;
@@ -215,6 +218,8 @@ public class Application {
 
     Handler handler = createHandler();
     server.setHandler(handler);
+
+    server.setStopTimeout(CONFIG.getPropertyValue(ScoutApplicationStopTimeoutProperty.class));
 
     return server;
   }
@@ -425,6 +430,7 @@ public class Application {
     String contextPath = CONFIG.getPropertyValue(ScoutApplicationContextPathProperty.class);
     handler.setContextPath(contextPath);
 
+    Handler returnHandler = handler;
     if (!"/".equals(contextPath)) {
       /*
        * Wraps the given <code>webAppContext</code> in a {@link P_RedirectToContextPathHandler} which redirects all GET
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/Application.java
```diff
@@ -435,13 +444,18 @@ protected Handler createHandler() {
...
+      returnHandler = new P_RedirectToContextPathHandler(handler);
     }
 
     if (BooleanUtility.nvl(CONFIG.getPropertyValue(ScoutApplicationGzipEnabled.class), true)) {
-      return createCompressionHandler(handler);
+      returnHandler = createCompressionHandler(returnHandler);
+    }
+
+    if (BooleanUtility.nvl(CONFIG.getPropertyValue(ScoutApplicationGracefulShutdownProperty.class), true)) {
+      returnHandler = createGracefulHandler(returnHandler);
```
paolobazzi, line 452: according to doc: https://jetty.org/docs/jetty/12.1/programming-guide/server/http.html#handler-use-graceful
graceful handler should be added as first handler -> move this code at top of this method?

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ApplicationProperties.java
```diff
@@ -633,4 +634,40 @@ public String description() {
...
+    }
+
+    @Override
+    public Boolean getDefaultValue() {
+      return Boolean.TRUE;
+    }
+
+    @Override
+    public String description() {
+      return "Enables graceful shutdown of the Jetty server (additional monitoring of all active requests). Default value is " + getDefaultValue() + ".";
```
paolobazzi:
```suggestion
      return "Enables graceful shutdown of the Jetty server (allowing active requests to terminate). Default value is " + getDefaultValue() + ".";
```

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ApplicationProperties.java
```diff
@@ -633,4 +634,40 @@ public String description() {
...
+  public static class ScoutApplicationStopTimeoutProperty extends AbstractPositiveLongConfigProperty {
+
+    @Override
+    public String getKey() {
+      return "scout.app.stopTimeout";
+    }
+
+    @Override
+    public Long getDefaultValue() {
+      return TimeUnit.MINUTES.toMillis(2);
```
paolobazzi: I'll would use a lowe value as default, e.g. 20-25 seconds.
Note: in a defautl kubernetes environment, a pod gets 30s default timeout for proper termination before gettign killed by SIGKILL.

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ApplicationProperties.java
```diff
@@ -633,4 +634,40 @@ public String description() {
...
+    }
+
+    @Override
+    public String description() {
+      return "The server stop timeout in milliseconds to allow a graceful shutdown (with less exceptions). Default value is " + getDefaultValue() + ".";
```
paolobazzi:
```suggestion
      return "The server stop timeout in milliseconds to allow a proper shutdown. Default value is " + getDefaultValue() + ".";
```

Changes:
- do not use "graceful" since this could indcate that property has just an effect when using graceful shutdown, according to jetty code this is not the case, this property affects shutdown even without graceful handler
- remove note about exceptions, this is just an effect of graceful shutdown

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/Application.java
```diff
@@ -216,6 +219,11 @@ protected Server createServer() {
     Handler handler = createHandler();
     server.setHandler(handler);
 
+    Long stopTimeout = CONFIG.getPropertyValue(ScoutApplicationStopTimeoutProperty.class);
+    if (stopTimeout != null) {
```
paolobazzi: is it possible for a property declaring a default value to return null here?

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/Application.java
```diff
@@ -435,13 +444,18 @@ protected Handler createHandler() {
...
-      return new P_RedirectToContextPathHandler(handler);
+      returnHandler = new P_RedirectToContextPathHandler(handler);
     }
 
     if (BooleanUtility.nvl(CONFIG.getPropertyValue(ScoutApplicationGzipEnabled.class), true)) {
-      return createCompressionHandler(handler);
+      returnHandler = createCompressionHandler(returnHandler);
+    }
+
+    if (BooleanUtility.nvl(CONFIG.getPropertyValue(ScoutApplicationGracefulShutdownProperty.class), true)) {
```
paolobazzi: is it possible for a property declaring a default value to return null here?  (same for gzip in line above?)

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/Application.java
matthiaso: Probably not, will remove null-check

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/Application.java
matthiaso: It is the outermost handler, isn't it? 

We wrap our previous handler here and the returned handler (the `GracefulHandler`) is registered with `Server.setHandler(...)`.

Therefore it is already the first handler in my opinion - no need to change anything.

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ApplicationProperties.java
matthiaso: Agree

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ApplicationProperties.java
matthiaso: Agree - for this input I decided to ask you to review this change 😉

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ApplicationProperties.java
matthiaso: Agree

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/Application.java
matthiaso: Probably not, will remote it

matthiaso: See https://github.com/eclipse-scout/scout.rt/compare/aa4ff3dca3b073a992424c8e6ad5a04c6f721d72..e277e52b22509414b054defb7e2afbcf60c7c18b

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/Application.java
matthiaso: See https://github.com/eclipse-scout/scout.rt/compare/aa4ff3dca3b073a992424c8e6ad5a04c6f721d72..e277e52b22509414b054defb7e2afbcf60c7c18b

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ApplicationProperties.java
matthiaso: https://github.com/eclipse-scout/scout.rt/compare/aa4ff3dca3b073a992424c8e6ad5a04c6f721d72..e277e52b22509414b054defb7e2afbcf60c7c18b

paolobazzi (approved):

PR #1976, opened 2026-02-03
File chooser: Correct handling of JS-files in Firefox

swg-bsi:
Firefox does not support the MIME type "text/javascript" as defined in RFC 9239 in the accept-attribute of input fields of type "file". This means that setting "js" as an allowed file extension on the Scout field is being ignored. This presents the user with a faulty default setting on the file chooser.

See also Bugzilla 2014180.

432191

### DIFF
--- a/org.eclipse.scout.rt.ui.html.test/src/test/java/org/eclipse/scout/rt/ui/html/json/basic/filechooser/JsonFileChooserAcceptAttributeBuilderTest.java
+++ b/org.eclipse.scout.rt.ui.html.test/src/test/java/org/eclipse/scout/rt/ui/html/json/basic/filechooser/JsonFileChooserAcceptAttributeBuilderTest.java
@@ -9,11 +9,10 @@
  */
 package org.eclipse.scout.rt.ui.html.json.basic.filechooser;
 
-import java.util.Arrays;
-import java.util.HashSet;
 import java.util.Set;
 
 import org.eclipse.scout.rt.platform.resource.MimeType;
+import org.eclipse.scout.rt.platform.util.CollectionUtility;
 import org.junit.Assert;
 import org.junit.Test;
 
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.ui.html.test/src/test/java/org/eclipse/scout/rt/ui/html/json/basic/filechooser/JsonFileChooserAcceptAttributeBuilderTest.java
```diff
@@ -81,13 +82,41 @@ public void testMimeWithCsv() {
         .build());
   }
 
+  @Test
+  public void testMimeWithJs() {
+    Assert.assertEquals(setOf(".js", ".mjs"), new JsonFileChooserAcceptAttributeBuilder()
```
bschwarzent: It should be possible to use `Set.of()` (none of the arguments are null, and the set does not have to be modified later)

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/basic/filechooser/JsonFileChooserAcceptAttributeBuilder.java
```diff
@@ -11,28 +11,39 @@
...
+ *     Therefore as a fallback for IE only we use file extensions for selected white-listed types.
+ *   </li>
+ *   <li>
+ *     Mozilla Firefox has a bug (Bugzilla 201480) with the MIME type text/javascript.
+ *     <p>
+ *     As a fallback we also use the file extensions defined in RFC 9239 for this MIME type.
+ *   </li>
+ * </ul>
+ *
  * This bean builds the content of the accept attribute in
```
bschwarzent: This paragraph should probably be on top.

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/basic/filechooser/JsonFileChooserAcceptAttributeBuilder.java
```diff
@@ -11,28 +11,39 @@
...
 /**
- * Microsoft internet explorer is not correctly handlig mime types in the 'accept' atttribute.
- * <p>
- * For example valid text/csv is not recognized.
- * <p>
- * Therefore as a fallback for IE only we use file extensions for selected white-listed types.
- * <p>
+ * In this class we correct for browser-side bugs such as:
+ * <ul>
+ *   <li>
```
bschwarzent, line 27: My IDE formats this differently, please check.

bschwarzent (approved):

PR #1982, opened 2026-02-05
Support for Session Nonce in PopupWindow

mvilliger:
A popup-window gets its own CSP nonce which is different from the main window opening the popup.
If a HtmlField belongs to a form running in such a popup-window (see PopupWindow.ts) and contains an inline script having a SessionNonce (see SessionNonce.java and JsonHtmlField#initJsonProperties), the initial nonce of the script tag is the one from the main window (because it comes from the main window).
To allow the script to be executed in the popup-window, the nonce must be switched to the one of the popup-window. This is safe to do with a simple replace as the nonce is a 256bit secure random (collisions almost impossible).

### DIFF
--- a/eclipse-scout-core/src/form/fields/htmlfield/HtmlField.ts
+++ b/eclipse-scout-core/src/form/fields/htmlfield/HtmlField.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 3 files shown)


eclipse-scout-core/src/form/fields/htmlfield/HtmlField.ts
```diff
@@ -48,12 +48,27 @@ export class HtmlField extends ValueField<string> implements HtmlFieldModel {
     this._renderEmpty();
   }
 
+  protected _prepareDisplayText(displayText: string): string {
+    const popupWindowNonce = Form.findNonWrappedForm(this)?.popupWindow?.nonce;
```
bschwarzent: FormFields have an instance method for that:

```suggestion
    const popupWindowNonce = this.findNonWrappedForm()?.popupWindow?.nonce;
```

eclipse-scout-core/src/form/fields/htmlfield/HtmlField.ts
```diff
@@ -48,12 +48,27 @@ export class HtmlField extends ValueField<string> implements HtmlFieldModel {
...
+    if (!popupWindowNonce) {
+      return displayText;
+    }
+
+    // In case this HtmlField belongs to a form running in a popup-window (see PopupWindow.ts).
```
bschwarzent, line 57: Optional: add a jasmine test (e.g. in PopupWindowSpec.ts)

bschwarzent (approved):

PR #1983, opened 2026-02-05
Improve logging for client session awaitJobs

matthiaso:
421614


matthiaso:
https://github.com/eclipse-scout/scout.rt/pull/1980

PR #1988, opened 2026-02-05
Avoid multi invalidation of HTTP session

matthiaso:
If session is already invalidated do not call invalidate again; also if invalidate fails because session is already invalidated ignore it.

Similar code used in SessionStore.checkHttpSessionOutsideWriteLock

377031


matthiaso:
https://github.com/eclipse-scout/scout.rt/pull/1987

PR #1989, opened 2026-02-05
TileGrid: fix text selection behavior

cguglielmo:
The tile table removes all selection ranges when tiles are selected to prevent text being selected when using shift to multi select tiles. This may remove the cursor from a focused input in Chrome because all selection ranges are removed not just the ones from the selected tiles. This is also applied when tiles are selected using keyboard even though the problem only exists when using mouse with shift. Furthermore, a regular tile grid could have the same issue, so why only apply it to the tile table?

The new solution just disables text-selection completely if a tile grid is selectable. This also prevents text from being selected when double-clicking a tile in order to perform the tile action.

Downside: text selection is now not possible anymore for selectable tiles. But it's the same for table and tree.

448709

### DIFF
--- a/eclipse-scout-core/src/tile/TileGrid.less
+++ b/eclipse-scout-core/src/tile/TileGrid.less
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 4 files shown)


eclipse-scout-core/src/tile/TileGrid.less
```diff
@@ -21,6 +21,11 @@
     padding-right: 15px;
   }
 
+  &.selectable * {
+    /* Text selection interferes with tile selection (e.g. double click, shift selection */
```
bschwarzent:
```suggestion
    /* Text selection interferes with tile selection (e.g. double click, shift selection) */
```

bschwarzent (approved):

PR #1991, opened 2026-02-06
TileGrid: fix text selection behavior

cguglielmo:
The tile table removes all selection ranges when tiles are selected to prevent text being selected when using shift to multi select tiles. This may remove the cursor from a focused input in Chrome because all selection ranges are removed not just the ones from the selected tiles. This is also applied when tiles are selected using keyboard even though the problem only exists when using mouse with shift. Furthermore, a regular tile grid could have the same issue, so why only apply it to the tile table?

The new solution prevents text selection when tiles are selected with the mouse while pressing shift without generally disabling text selection on selectable tiles.

448709

### DIFF
--- a/eclipse-scout-core/src/util/events.ts
+++ b/eclipse-scout-core/src/util/events.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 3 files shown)


eclipse-scout-core/src/util/events.ts
```diff
@@ -214,6 +214,16 @@ export const events = {
...
   },
 
+  /**
+   * Prevents text selection for the given `$container` until a mouse up occurs anywhere on the window.
+   */
+  preventSelectionStartUntilMouseUp($container: JQuery) {
+    const preventSelectionHandler = event => event.preventDefault();
+    $container
+      .on('selectstart', preventSelectionHandler)
+      .window().one('mouseup', () => $container?.off('selectstart', preventSelectionHandler));
```
bschwarzent, line 224:
```suggestion
      .window().one('mouseup', () => $container.off('selectstart', preventSelectionHandler));
```
$container is a local variable and cannot become null. (If at all, the check should be at the start of the method.)

bschwarzent (approved):

cguglielmo:
Merged with https://github.com/eclipse-scout/scout.rt/pull/2001

PR #1992, opened 2026-02-06
Make reloadReason accessible in TableMaxResultsHelper

mvilliger:
This allows to customize the max row counts based on the reason of the reload. E.g. to load more data for reload reason 'overrideRowLimit' (see TableFooter._onInfoLoadAction).

432873

### DIFF
--- a/eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
+++ b/eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
@@ -22,6 +22,7 @@ export class PageWithTable extends Page implements PageWithTableModel {
   searchRequired = false;
   searchFilterCompleted = false;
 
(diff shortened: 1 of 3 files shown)


eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
```diff
@@ -21,6 +21,7 @@ export class PageWithTable extends Page implements PageWithTableModel {
   alwaysCreateChildPage: boolean;
   searchRequired = false;
   searchFilterCompleted = false;
+  reloadReason: TableReloadReason;
```
fschinkel: Does this need to be a public property?
Please add some js doc or change the variable name so it is clear that this is the most recent reload reason that was passed to `loadTableData`.

fschinkel (changes requested):

PR #2002, opened 2026-02-09
Immediately create and add Collection/List/Set nodes for DoEntity

matthiaso:
It is not possible to create/put them later as this would lead to concurrency issues (e.g. multiple threads trying to get the same entity which might trigger concurrent putNode calls).

396980

### DIFF
--- a/org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/DoEntity.java
+++ b/org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/DoEntity.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/DataObjectHelper.java
matthiaso, line 2:
```suggestion
 * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
```

org.eclipse.scout.rt.jackson/src/main/java/org/eclipse/scout/rt/jackson/dataobject/DoEntitySerializer.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
```
matthiaso:
```suggestion
 * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
```

org.eclipse.scout.rt.dataobject/src/main/java/org/eclipse/scout/rt/dataobject/DoEntity.java
```diff
@@ -208,7 +248,7 @@ public boolean equals(Object o) {
 
     DoEntity doEntity = (DoEntity) o;
 
-    if (!m_attributes.equals(doEntity.m_attributes)) {
+    if (!allNodes().equals(doEntity.allNodes())) {
```
matthiaso, line 251: hashCode?

PR #2006, opened 2026-02-10
Improve support for password managers

mvilliger:
- LoginBox: Use autocomplete attributes to help the password managers to
  identify the input fields.
- Disable spellchecker for masked fields and username field.
- Add aria-labels on LoginBox fields.

449883

### DIFF
--- a/eclipse-scout-core/src/form/fields/stringfield/StringField.ts
+++ b/eclipse-scout-core/src/form/fields/stringfield/StringField.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -244,7 +244,10 @@ export class StringField extends BasicField<string> implements StringFieldModel 
     if (this.multilineText) {
       return;
     }
-    this.$field.attr('type', this.inputMasked ? 'password' : 'text');
+
+    this.$field
+      .toggleAttr('spellcheck', this.inputMasked, 'false')
+      .attr('type', this.inputMasked ? 'password' : 'text');
   }
 
   protected _renderInputObfuscated() {
--- a/eclipse-scout-core/src/login/LoginBox.ts
+++ b/eclipse-scout-core/src/login/LoginBox.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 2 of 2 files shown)


eclipse-scout-core/src/form/fields/stringfield/StringField.ts
```diff
@@ -244,7 +244,13 @@ export class StringField extends BasicField<string> implements StringFieldModel
     if (this.multilineText) {
       return;
     }
-    this.$field.attr('type', this.inputMasked ? 'password' : 'text');
+
+    this.$field
+      .toggleAttr('autocomplete', this.inputMasked, 'new-password')
```
cguglielmo: Are these attributes really necessary? If a browser autocapitalized a letter if type is password, it would be a very stupid browser...

eclipse-scout-core/src/login/LoginBox.ts
```diff
@@ -88,11 +88,16 @@ export class LoginBox extends Box {
...
     }
+    const userText = this.texts.get('ui.User');
     this.$user = $('<input>')
       .attr('type', 'text')
+      .attr('id', 'username')
+      .attr('autocomplete', 'username')
       .attr('autocapitalize', 'off')
-      .attr('autocorrect', 'off')
-      .placeholder(this.texts.get('ui.User'))
+      .attr('autocorrect', 'false')
```
cguglielmo: According to mdn 'autocorrect' can be set to on or off, not false
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/autocorrect

eclipse-scout-core/src/login/LoginBox.ts
```diff
@@ -88,11 +88,16 @@ export class LoginBox extends Box {
...
+      .attr('spellcheck', 'false')
+      .attr('aria-label', userText)
```
cguglielmo, line 98: Is aria-label necessary if there is a placeholder?

eclipse-scout-core/src/login/LoginBox.ts
```diff
@@ -101,14 +106,20 @@ export class LoginBox extends Box {
       .addClass('login-button button default')
       .text(this.texts.get('ui.Login'))
       .appendTo(this.$form);
-
-    this.$user.focus();
+    this.$user.trigger('focus');
```
cguglielmo: I assume you replaced it because of the deprecation? I did not know that trigger('focus') actually focuses the element. This is very unexpected. I normally use this.$user[0].focus()

eclipse-scout-core/src/form/fields/stringfield/StringField.ts
cguglielmo: I don't think it is safe to assume that a string field is always used for the new-password case. E.g. we already have a LoginForm.js which is used to login, so in that case autocomplete should be set to current-password.

eclipse-scout-core/src/login/LoginBox.ts
mvilliger: Actually I am not sure. Major login forms e.g. from Microsoft or Google have both: placeholder and aria-attributes. Therefore I assumed it is necessary?

eclipse-scout-core/src/form/fields/stringfield/StringField.ts
mvilliger: Agree

PR #2011, opened 2026-02-10
Do return plain input stream for images in MailHelper.getPartContent

matthiaso:
Especially avoid potentially registered data content handlers to wrap this input stream (providing an input stream as well in the end) leading to performance loss.

362182


matthiaso:
Cherry-pick of https://github.com/eclipse-scout/scout.rt/pull/1998

PR #2013, opened 2026-02-11
UiPreferences: improve applying column preferences

fschinkel:
When the column preferences are applied the column is updated directly, but some preferences can only be applied if the column is of a specific type (e.g. NumberColumn). If there are preferences set for these cases the preferences are lost. Normally this is ok, because the column can not handle these preferences, but it makes it impossible to e.g. use placeholder columns if the real column can only be created asynchronously. Therefore, introduce the interface ColumnPreferencesColumn which contains a getter and a setter for the whole preferences. If preferences are applied to such a column the whole TableColumnClientUiPreferenceDo is passed to the setter and later on these preferences are used as a fallback for preference values that can only be extracted from specific column types when the preferences are stored.

449721

### DIFF
--- a/eclipse-scout-core/src/prefs/TableUiPreferences.ts
+++ b/eclipse-scout-core/src/prefs/TableUiPreferences.ts
@@ -256,8 +256,16 @@ export class TableUiPreferences implements ObjectWithType {
           sortOrder: column.sortIndex,
           sortAscending: column.sortAscending,
           groupingActive: column.grouped,
-          aggregationFunctionId: column instanceof NumberColumn ? column.aggregationFunction : undefined,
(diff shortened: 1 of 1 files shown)


eclipse-scout-core/src/prefs/TableUiPreferences.ts
```diff
@@ -571,3 +587,8 @@ UiPreferences.registerHandler(TableUiPreferences, {
     preferences.tablePreferences = tableUiPreferences._exportTablePreferences();
   }
 });
+
+export interface ColumnPreferencesColumn<TValue = any> extends Column<TValue> {
```
bschwarzent, line 599: Maybe add a short jsdoc explaining this interface?

bschwarzent (approved):

PR #2016, opened 2026-02-12
TileTable: fix focus handling

cguglielmo:
The tile table should only show selection and menubar if the focus is inside the tile. This broke because the table data now has the focus and not the container.

The change also fixes a bug, that was already there before the focus refactoring:
When the table is focused and tab pressed, the focus went to the body instead of focusing the first menubar item. This happened because the menubar was made invisible as soon as focus left the table. Now, the menubar is only made invisible, when the focus leaves the tile, it may be on the table data, a menu bar item, a header item or the text filter.

450639

### DIFF
--- a/eclipse-scout-core/src/tile/fields/tablefield/TileTableField.ts
+++ b/eclipse-scout-core/src/tile/fields/tablefield/TileTableField.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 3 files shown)


eclipse-scout-core/src/tile/fields/tablefield/TileTableField.ts
bschwarzent: This does not work properly. When 'focusout' is triggered, the focus is still at the old element. If we hide the menu bar now, the focus will still be get put back to the start. What we want is to let the focus move to the next element, and _then_ hide the menubar. Easiest way would probably be to use setTimeout.

cguglielmo: Thank you, I now use requestAnimationFrame to hide it

bschwarzent (approved):

PR #2017, opened 2026-02-13
BookmarkDoBuilder: use nicer search filter text in bookmark description

bschwarzent:
Instead of showing low-level JSON values, use the text computed by PageWithTable#getSearchFilterText.

447101

### DIFF
--- a/eclipse-scout-core/src/bookmark/BookmarkDoBuilder.ts
+++ b/eclipse-scout-core/src/bookmark/BookmarkDoBuilder.ts
@@ -8,9 +8,8 @@
  * SPDX-License-Identifier: EPL-2.0
  */
 import {
-  arrays, BaseDoEntity, BookmarkDo, BookmarkDoBuilderModel, BookmarkSupport, BookmarkTableRowIdentifierDo, ChartTableControlConfigHelper, Desktop, Form, IBookmarkDefinitionDo, IBookmarkDo, IBookmarkPageDo, IChartTableControlConfigDo,
-  InitModelOf, NodeBookmarkPageDo, objects, ObjectWithType, OutlineBookmarkDefinitionDo, Page, PageBookmarkDefinitionDo, PageWithTable, scout, Session, strings, TableBookmarkPageDo, TableClientUiPreferencesDo, TableUiPreferences,
-  tableUiPreferences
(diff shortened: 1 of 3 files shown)


eclipse-scout-core/src/bookmark/BookmarkDoBuilder.ts
```diff
@@ -275,18 +287,23 @@ export class BookmarkDoBuilder implements ObjectWithType, BookmarkDoBuilderModel
...
-          .filter(key => !strings.startsWith(key, '_'))
-          .filter(key => !objects.isNullOrUndefinedOrEmpty(searchData[key]))
-          .map(key => key + ': ' + JSON.stringify((searchData)[key]))
-          .forEach(searchLine => lines.push(prefix + '  ' + searchLine));
+      if (bookmarkPage instanceof TableBookmarkPageDo) {
+        // Get search filter text added by _tablePageToTableBookmarkPage()
+        let searchFilterText = this._searchFilterTexts.get(bookmarkPage);
+        if (searchFilterText) {
+          // Normalize line endings and remove trailing whitespace
+          searchFilterText = searchFilterText.replaceAll('\r', '').replace(/[\s\n]+$/g, '');
```
cguglielmo: Why is this necessary? Should this be done by the SearchFilterTextBuilder itself?

eclipse-scout-core/src/bookmark/BookmarkDoBuilder.ts
```diff
@@ -275,18 +287,23 @@ export class BookmarkDoBuilder implements ObjectWithType, BookmarkDoBuilderModel
...
+          // Split search filter text into lines and add them to the description (with additional indentation)
+          if (searchFilterText) {
+            lines.push(...searchFilterText.split('\n').map(s => prefix + indent + s));
```
cguglielmo: Should we change the SearchFilterTextBuilder so it returns an array of strings?

eclipse-scout-core/src/bookmark/BookmarkDoBuilder.ts
bschwarzent: Removed since normalization is not necessary here.

eclipse-scout-core/src/bookmark/BookmarkDoBuilder.ts
bschwarzent: Since splitting is very simple, we keep the current API.

PR #2020, opened 2026-02-16
Make AbstractMomSubscriber thread-safe

rogersteblerbsi:
The list of subscriptions should be thread-safe, so that registrations can be done in parallel.

411785

### DIFF
--- a/org.eclipse.scout.rt.mom.api.test/src/test/java/org/eclipse/scout/rt/mom/api/MomSubscriberTest.java
+++ b/org.eclipse.scout.rt.mom.api.test/src/test/java/org/eclipse/scout/rt/mom/api/MomSubscriberTest.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.scout.rt.mom.api;
+
+import static org.mockito.Mockito.*;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.mom.api.test/src/test/java/org/eclipse/scout/rt/mom/api/MomSubscriberTest.java
```diff
@@ -0,0 +1,78 @@
...
+      public void subscribe() {
+        for (ISubscription subscriptionMock : subscriptionMocks) {
+          registerSubscription(subscriptionMock);
+        }
+      }
+    };
+
+    subscriber.subscribe(); // insert initial subscriptions
+    IFuture<Void> subscriptionFuture = Jobs.schedule(subscriber::subscribe, Jobs.newInput()); // insert more subscriptions async
+    subscriber.dispose(); // start disposing while new subscriptions are being added
```
matthiaso, line 75: Not sure if you can really be sure that at least one of the IFutures scheduled is already running when you are calling dispose.

In a similar test I wrote lines like this: https://github.com/eclipse-scout/scout.rt/pull/2025/changes#diff-c5f30f594908d9c266392aed47c0b2dbab36c21e68e01aa9a91cac725ebce0b2R991-R995 (but on the other hand I am also not sure if this is always the case) - maybe your dispose lasts long enough (because there are a lot of mocks) so that at least one IFuture is usually running.

I suppose this test is failing w/o your change?

matthiaso (approved):

org.eclipse.scout.rt.mom.api.test/src/test/java/org/eclipse/scout/rt/mom/api/MomSubscriberTest.java
rogersteblerbsi: As discussed we leave it like this. It's not 100% reliable. But I run the test without the fix 1000 times and it only was "mistakenly" green 1 time. With the fix it should be green all the time.

PR #2022, opened 2026-02-16
Add lifecycle to FormModel

fschinkel:
The lifecycle or the lifecycle model may be adjusted depending on the context a form is used in. Add the possibility to pass an object type or a child model of a FormLifecycle in the model of a form.

449831

### DIFF
--- a/eclipse-scout-core/src/form/Form.ts
+++ b/eclipse-scout-core/src/form/Form.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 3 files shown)


eclipse-scout-core/src/form/Form.ts
```diff
@@ -321,13 +321,40 @@ export class Form extends Widget implements FormModel, DisplayParent {
...
+    // object type
+    if (typeof this.lifecycle === 'string' || typeof this.lifecycle === 'function') {
+      return scout.create(this.lifecycle, {
+        ...defaultModel,
+        widget: this
+      });
+    }
+
+    // child model
+    if (objects.isObject(this.lifecycle)) {
```
cguglielmo: I think isPojo() would be better

eclipse-scout-core/src/form/Form.ts
```diff
@@ -321,13 +321,40 @@ export class Form extends Widget implements FormModel, DisplayParent {
...
+      return scout.create({
+        objectType: FormLifecycle,
+        ...defaultModel,
+        ...(this.lifecycle as ChildModelOf<FormLifecycle>),
```
cguglielmo: Brackets are not necessary

eclipse-scout-core/src/form/FormModel.ts
```diff
@@ -230,4 +230,8 @@ export interface FormModel extends WidgetModel, DisplayParentModel {
    * By default, the list is empty.
    */
   validators?: FormValidator[];
+  /**
+   * Defines the {@link FormLifecycle} that is used for the validation lifecycle of the form.
```
cguglielmo: It is not only used for validation, also for load, save and reset etc.

"Defines the {@link FormLifecycle} for this form" should be sufficient.

PR #2025, opened 2026-02-17
Immediately create and add Collection/List/Set nodes for DoEntity

matthiaso:
It is not possible to create/put them later as this would lead to concurrency issues (e.g. multiple threads trying to get the same entity which might trigger concurrent putNode calls).

396980


matthiaso:
This PR replaces https://github.com/eclipse-scout/scout.rt/pull/2002

matthiaso:
Replaced with https://github.com/eclipse-scout/scout.rt/pull/2323 (new target version)

PR #2028, opened 2026-02-18
Calendar: avoid opening menu on unselected component

Knicksaas:
The javascript 'contextmenu' event is not consistent with its order. On some machines, it is triggered right after right-click 'mousedown', on other machines after 'mouseup'. This can lead to an unexpected behavior, when the right-click is started outside a component and is finished on a component. Because the selection is applied on 'mousedown', no component is selected, but the context menu of the component is opened. This change cancels the opening of the popup window When this case occurs.

451116


jwinkens (approved):
Review ok

PR #2032, opened 2026-02-19
Do not send popups with unresolvable anchors to Scout JS

fschinkel:
The PopupManager contains a set of popups, which themselves may contain anchors. When a JsonPopup builds its json it may happen that there is no json adapter for the anchor of the popup, because the anchor is e.g. part of an outline that is currently not the active outline. The JsonPopup can not leave its anchor property empty as a popup without anchor will be rendered directly which would be incorrect. In addition, the JsonPopup can not return null when creating its json as this leads to other errors.
Therefore, the JsonPopupManager needs to filter its set of popups and ensure that no popup containing an anchor without a json adapter is sent to Scout JS. To ensure correct updates of its property "popups" the JsonPopupManager needs to observe all current anchors of its popups and perform an update of its property "popups" if json adapters are created for one of these observed anchors. This is relevant e.g. when the JsonPopupManager is processed before the json adapter of one of its popups anchors is processed which may happen e.g. when the browser is reloaded.
As the PopupManager.ts only renders newly added popups and as a formerly rendered popup will be removed when its anchor is removed, it is not necessary to listen on json adapter disposals. The only time the PopupManager.ts will try to rerender these popups with potentially unavailable anchors is when the browser is reloaded and in this case the property "popups" will be sent again and is therefore correctly filtered.
Additionally, close a Popup if its anchor is disposed to prevent memory leaks.
Some popup anchors may be part of tiles. Therefore, reposition those popups after the layout animation of the TileGrid is done.

448950

### DIFF
--- a/org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/popup/JsonPopupManager.java
+++ b/org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/popup/JsonPopupManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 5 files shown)


org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/popup/JsonPopupManager.java
```diff
@@ -35,11 +52,247 @@ public String getObjectType() {
...
+
+  /**
+   * Checks whether the given {@link IWidget} is part of the {@link #getAnchorsWithoutJsonAdapter()}.
+   */
+  protected boolean isAnchorWithoutJsonAdapter(IWidget anchor) {
+    return getAnchorsWithoutJsonAdapter().contains(anchor);
+  }
+
+  /**
+   * Adds the given anchor to the {@link #getAnchorsWithoutJsonAdapter()} if there does not exist an {@link IJsonAdapter} for it.
```
cguglielmo:
```suggestion
   * Adds the given anchor to the {@link #getAnchorsWithoutJsonAdapter()} if there is no {@link IJsonAdapter} for it.
```

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/popup/JsonPopupManager.java
```diff
@@ -35,11 +52,247 @@ public String getObjectType() {
...
+   * If this changes the {@link #getAnchorsWithoutJsonAdapter()} the {@link #m_anchorAdapterChangeListener} is updated if necessary.
+   *
+   * @return <code>true</code> if {@link #getAnchorsWithoutJsonAdapter()} changed
+   */
+  protected boolean addAnchorWithoutJsonAdapter(IWidget anchor) {
+    if (anchor == null) {
+      return false;
+    }
+
+    // there exists a json adapter for the anchor -> nothing to observe
```
cguglielmo:
```suggestion
    // there is a json adapter for the anchor -> nothing to observe
```

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/popup/JsonPopupManager.java
```diff
@@ -35,11 +52,247 @@ public String getObjectType() {
...
+      return false;
+    }
+
+    // anchor is already observed
+    if (!getAnchorsWithoutJsonAdapter().add(anchor)) {
+      return false;
+    }
+
+    // nothing to observe or listener is created already
+    if (getAnchorsWithoutJsonAdapter().isEmpty() || m_anchorAdapterChangeListener != null) {
```
cguglielmo: How can the list be empty if add was called at line 191?

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/json/popup/JsonPopupManager.java
```diff
@@ -35,11 +52,247 @@ public String getObjectType() {
...
+      updatePopupsJsonProperty();
+    }
+  }
+
+  /**
+   * Updates the json property {@link PopupManager#PROP_POPUPS} (see {@link #updatePopupsJsonProperty()}) when a json adapter is created for one of the {@link #getAnchorsWithoutJsonAdapter()}.
+   * This is relevant e.g. when the {@link JsonPopupManager} is processed before the json adapter of one of its popups anchors is processed which may happen e.g. when the browser is reloaded.
+   * As the <code>PopupManager.ts</code> only renders newly added popups and as a formerly rendered popup will be removed when its anchor is removed, it is not necessary to listen on json adapter disposals.
+   * The only time the <code>PopupManager.ts</code> will try to rerender these popups with potentially unavailable anchors is when the browser is reloaded
+   * and in this case the property {@link PopupManager#PROP_POPUPS} will be sent again and is therefore correctly  filtered.
```
cguglielmo:
```suggestion
   * and in this case the property {@link PopupManager#PROP_POPUPS} will be sent again and is therefore correctly filtered.
```

PR #2037, opened 2026-02-19
ClipboardForm: add menu to copy text by mouse

bschwarzent:
Instead of pressing ctrl-c or using the context menu, the user can now also click the "copy" menu to copy the selected text to the clipboard.

Additional changes:
- Remove EscapeKeyStroke, instead mark the form as closable. This has the same effect but the F1 keystroke hint is positioned much nicer.
- Cleanup some clipboard related texts.
- Make instruction label non-selectable, just like any other form field label.

378797


eclipse-scout-core/src/jquery/jquery-scout.js
```diff
@@ -1449,25 +1449,84 @@ $.fn.unfocusable = function() {
 };
 
 $.fn.selectAllText = function() {
-  let range,
-    myDocument = this.document(true),
-    myWindow = this.window(true),
-    element = this[0];
+  let element = this[0];
+  let document = element?.ownerDocument;
```
cguglielmo: element.ownerDocument is not exactly the same as this.document(true). If `this` is the document, it does not work. Currently it is possible to select all text in the document using $(document).selectAllText(). It is probably never used like this, but still...

eclipse-scout-core/src/jquery/jquery-scout.js
```diff
@@ -1449,25 +1449,84 @@ $.fn.unfocusable = function() {
...
+    range.selectNodeContents(element);
+    selection.removeAllRanges();
+    selection.addRange(range);
+  }
   return this;
 };
 
+$.fn.selectedText = function() {
+  let element = this[0];
+  let document = element?.ownerDocument;
```
cguglielmo: Maybe use $(element).document(true), see above

cguglielmo (approved):

eclipse-scout-core/src/jquery/jquery-scout.js
bschwarzent: Agreed. Code and Spec were adjusted accordingly.

eclipse-scout-core/src/jquery/jquery-scout.js
bschwarzent: Agreed. Code and Spec were adjusted accordingly.

bschwarzent:
Submitted to 26.2: https://github.com/eclipse-scout/scout.rt/commit/b21b07f80955c4742c6d596cfc781c51e5201c56

PR #2038, opened 2026-02-19
OWASP dependency check

crispin-kirchner:


crispin-kirchner:
no cherry-pick necessary, because 25.2 already uses a fixed version

PR #2050, opened 2026-02-23
Lifecycle _validateWidget is skipped if _validateElements has warnings

mvilliger:
If _validateElements returns only warnings, the _validateWidget is skipped and the warning status is directly returned. This might be wrong in case _validateWidget would return error status.

423378


org.eclipse.scout.rt.nls/src/main/resources/org/eclipse/scout/rt/nls/texts/ScoutTexts_de.properties
```diff
@@ -132,7 +132,9 @@ FormSaveChangesQuestion=Möchten Sie die Änderungen speichern?
 FormStateLoad=Öffnen
 FormStateStore=Speichern
 FormStateStoreAs=Speichern unter
+FormValidationErrorMessage=Die folgenden Probleme müssen behandelt werden\:
 FormValidationFailedTitle=Bitte Eingaben überprüfen
+FormValidationWarningMessage=Darüber hinaus wurden folgende Warnungen entdeckt\:
```
cguglielmo:
```suggestion
FormValidationWarningMessage=Zudem wurden folgende Warnungen entdeckt\:
```

eclipse-scout-core/src/status/Status.ts
```diff
@@ -151,29 +177,41 @@ export class Status implements StatusModel, ObjectWithType {
...
 
-      function calcPriority(status) {
-        let multiplier = 1;
-        if (status instanceof ParsingFailedStatus) {
-          multiplier = 4;
-        } else if (status instanceof ValidationFailedStatus) {
-          multiplier = 2;
-        }
-        return multiplier * status.severity;
+    const all = this.asFlatList();
```
cguglielmo, line 204: Is this different to the existing sort-logic?

mvilliger: No it should be the same logic as before. Only the implementation has changed. There is no need to fully sort an array just to find a maximum.

PR #2061, opened 2026-03-02
Outline: add keystroke to reload the root page

bschwarzent:
If a user wishes to recreate a page (e.g. because its content has been changed, but it offers no means to be refreshed), they can usually just reload the detail table of the parent page. However, since the outline in JS has no root page, this is not possible for top-level pages. For a pure JS outline, this can be resolved by recreating the entire application (reload page in browser, Ctrl-R), but for classic outlines, the server-side model has to be updated.

A new keystroke (F5) provides this function with the help of a hybrid action. It is only active for remote outlines, and only when no page is selected.

411299

### DIFF
--- a/eclipse-scout-core/src/desktop/outline/keystrokes/OutlineReloadKeyStroke.ts
+++ b/eclipse-scout-core/src/desktop/outline/keystrokes/OutlineReloadKeyStroke.ts
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {HybridActionContextElements, HybridManager, keys, KeyStroke, Outline, scout, ScoutKeyboardEvent} from '../../../index';
+
+/**
+ * Keystroke to reload all nodes of an outline on the UI server via hybrid action. Only active when no node is selected.
+ */
+export class OutlineReloadKeyStroke extends KeyStroke {
+  declare field: Outline;
+
+  protected _reloading = false;
+
+  constructor(outline: Outline) {
+    super();
+    this.field = outline;
+    this.which = [keys.F5];
+
+    this.renderingHints.offset = 8; // increase distance from left edge
+    this.renderingHints.$drawingArea = ($drawingArea, event) => this.field.$title || this.field.$data; // same as OutlineNavigateToTopKeyStroke
+  }
+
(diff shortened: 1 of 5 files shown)


eclipse-scout-core/src/desktop/outline/keystrokes/OutlineReloadKeyStroke.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  override handle(event: JQuery.KeyboardEventBase) {
+    this._reloading = true;
+    this._handleAsync(event)
+      .finally(() => {
+        this._reloading = false;
+      });
+  }
+
+  async _handleAsync(event: JQuery.KeyboardEventBase): Promise<void> {
```
fschinkel: Mark as `protected`. Have you considered moving the `this._reloading`-guard into this method, so you can use a classic finally-block?

eclipse-scout-core/src/desktop/outline/keystrokes/OutlineReloadKeyStroke.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    this.renderingHints.offset = 8; // increase distance from left edge
+    this.renderingHints.$drawingArea = ($drawingArea, event) => this.field.$title || this.field.$data; // same as OutlineNavigateToTopKeyStroke
+  }
+
+  protected override _accept(event: ScoutKeyboardEvent): boolean {
+    if (this._reloading) {
+      return false; // prevent parallel hybrid actions
+    }
+    if (!this.field.modelAdapter) {
```
fschinkel: As this key stroke only reloads remote outlines, it must not be present on js only outlines. Therefore, the registration of this key stroke needs to be moved to the `OutlineAdapter`. Please, check with @cguglielmo if this key stroke needs to be part of the `OutlineKeyStrokeContext` and why this context not accessible on the `Outline`.

fschinkel (changes requested):

eclipse-scout-core/src/desktop/outline/keystrokes/OutlineReloadKeyStroke.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  override handle(event: JQuery.KeyboardEventBase) {
+    // noinspection JSIgnoredPromiseFromCall
+    this._handleAsync(event);
+  }
+
+  protected async _handleAsync(event: JQuery.KeyboardEventBase): Promise<void> {
+    this._reloading = true;
+    try {
+      let hybridManager = await HybridManager.get();
```
fschinkel: nothing to await here

fschinkel (approved):

bschwarzent:
Closed in favor of https://github.com/eclipse-scout/scout.rt/pull/2117

PR #2069, opened 2026-03-04
Revert jetty to version 12.0.33

paolobazzi:
Due to the following issues a downgrade to latest 12.0.32 version

- https://github.com/jetty/jetty.project/issues/14264
- https://github.com/jetty/jetty.project/issues/14564

451954

### DIFF
--- a/org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/Application.java
+++ b/org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/Application.java
@@ -25,12 +25,6 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 
 import org.eclipse.jetty.alpn.server.ALPNServerConnectionFactory;
-import org.eclipse.jetty.compression.gzip.GzipCompression;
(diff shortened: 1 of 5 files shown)


org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/Application.java
matthiaso, line 492: This method should be reverted as well, see https://github.com/eclipse-scout/scout.rt/commit/c1eab82ee26baf7e57298ab2b77eaff24dcbd90e#diff-2598791ee2d3d8ea84e41e8b766773319707e3b689848372f1a400f17e5892a6L456-R475


```suggestion
  protected void setStringPropertyValueIfFilled(Class<? extends IConfigProperty<List<String>>> propertyClass, Consumer<String[]> setter) {
    List<String> propertyValue = CONFIG.getPropertyValue(propertyClass);
    if (!CollectionUtility.isEmpty(propertyValue)) {
      setter.accept(propertyValue.toArray(n -> new String[n]));
    }
  }
```

Otherwise if multiple properties are set in the `List<String>` only the last one would actually be set.

matthiaso (changes requested):

matthiaso (approved):

PR #2078, opened 2026-03-06
BasicTransaction: addFailure must be thread-safe

matthiaso:
Multiple threads may call this method at the same time, e.g. transaction
 is cancelled by one thread and a failure (maybe because of
 cancellation) occurs in another thread.

453426


matthiaso:
https://github.com/eclipse-scout/scout.rt/pull/2073

PR #2085, opened 2026-03-09
HybridManager: improve widget disposal

fschinkel:
Add api to HybridManager.ts for widget disposal. Widgets that need to be disposed are collected and the scout.DisposeWidgets-hybrid action is only called once. In addition, send a scout.DisposeWidgets-hybrid action and stop destroying for all widgets belonging to the HybridManager if they are destroyed.
All widgets for which a disposal was scheduled are excluded from property change events in the future as destroyed widgets may no longer be resolved.
Improve performance of DisposeWidgetsHybridAction. The HybridManager adds dispose-listeners to all its widgets and removes them from itself when they are disposed. For this the HybridManager needs to build a new map of widgets, each time its "widgets" property is updated. If a great amount of widgets is disposed, each disposal triggers a rebuild of the HybridManagers "widgets" property. The DisposeWidgetsHybridAction will now remove all widgets that are to be disposed from the HybridManager at once and dispose them afterward. With this the "widgets" property of the HybridManager is only updated once which improves performance when great amounts of widgets are disposed.

379008

### DIFF
--- a/eclipse-scout-core/src/desktop/hybrid/HybridManager.ts
+++ b/eclipse-scout-core/src/desktop/hybrid/HybridManager.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -8,8 +8,8 @@
  * SPDX-License-Identifier: EPL-2.0
  */
 import {
-  AnyDoEntity, App, Event, EventHandler, EventListener, EventMapOf, Form, HybridActionContextElements, HybridActionEvent, HybridManagerEventMap, HybridManagerWidgetAddEvent, HybridManagerWidgetRemoveEvent, InitModelOf, ObjectOrChildModel,
-  scout, Session, UuidPool, Widget
+  AnyDoEntity, App, arrays, DisposeWidgetsHybridActionDo, Event, EventHandler, EventListener, EventMapOf, Extension, Form, HybridActionContextElements, HybridActionEvent, HybridManagerEventMap, HybridManagerWidgetAddEvent,
+  HybridManagerWidgetRemoveEvent, InitModelOf, ObjectOrChildModel, objects, scout, Session, strings, UuidPool, Widget
 } from '../../index';
 
 /**
@@ -21,6 +21,17 @@ export class HybridManager extends Widget {
   declare self: HybridManager;
 
   widgets: Record<string, Widget>;
+
+  /**
+   * Set of {@link HybridManagerWidget}s that will be disposed in the next batch.
+   * See {@link #_callDisposeWidgets} for more information.
+   */
+  protected _widgetsToBeDisposed: Set<HybridManagerWidget> = null;
+  /**
+   * Set of ids that were scheduled in former batches.
+   * See {@link #_callDisposeWidgets} for more information.
+   */
+  protected _widgetIdsBeingDisposed = new Set<string>();
 
   constructor() {
     super();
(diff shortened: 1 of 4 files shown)


eclipse-scout-core/src/desktop/hybrid/HybridManager.ts
```diff
@@ -79,8 +120,16 @@ export class HybridManager extends Widget {
 
     this._setProperty('widgets', widgets);
 
-    Object.entries(addedWidgets).forEach(([id, widget]) => this._triggerWidgetAdd(id, widget));
-    Object.entries(removedWidgets).forEach(([id, widget]) => this._triggerWidgetRemove(id, widget));
+    Object.entries(addedWidgets).forEach(([id, widget]) => {
+      this._installHybridManagerWidget(widget, id);
+      // trigger widgetAdd event
```
cguglielmo: There is no need to add a comment if the method call is meaningful enough and basically the same as the comment

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/hybrid/DisposeWidgetsHybridAction.java
```diff
@@ -30,13 +32,20 @@ public class DisposeWidgetsHybridAction extends AbstractHybridAction<DisposeWidg
...
-    for (String id : data.getIds()) {
-      IWidget widget = hybridManager().getWidgetById(id);
-      if (widget != null) {
-        widget.dispose();
-        LOG.debug("Disposed hybrid widget with id {}", id);
-      }
-    }
+    // get all widgets that need to be disposed
+    Map<String, IWidget> widgets = hybridManager().getWidgets().entrySet().stream()
+        .filter(entry -> data.getIds().contains(entry.getKey()) && entry.getValue() != null)
```
cguglielmo: Why not loop over data.getIds() and call getWidgetById()? It would be faster because looping over getWidgets() and using getIds().contains is O(n^2)

eclipse-scout-core/src/desktop/hybrid/HybridManager.ts
```diff
@@ -221,6 +290,55 @@ export class HybridManager extends Widget {
...
+    }
+
+    // if next batch was not created already, create it and queue microtask to sent hybrid action
+    if (!this._disposeWidgetsNextBatch) {
+      this._disposeWidgetsNextBatch = new Set<HybridManagerWidget>();
+      queueMicrotask(() => {
+        // collect remote ids, transfer widget ids to this._disposeWidgetsScheduledWidgetIds and reset next batch
+        const remoteIds: string[] = [];
+        for (const widget of [...this._disposeWidgetsNextBatch]) {
+          remoteIds.push(widget.__remoteId);
```
cguglielmo, line 314: - widget.__removeId is null for forms, because _uninstallHybridManagerWidget is called before the microtask runs
- Forms are disposed on server, do we even need a destroy listener for forms?

eclipse-scout-core/src/desktop/hybrid/HybridManager.ts
```diff
@@ -22,6 +22,18 @@ export class HybridManager extends Widget {
...
+  /**
+   * Set of {@link HybridManagerWidget}s that will be disposed in the next batch.
+   * See {@link #_callDisposeWidgets} for more information.
+   */
+  protected _disposeWidgetsNextBatch: Set<HybridManagerWidget> = null;
+  /**
+   * Set of ids that were scheduled in former batches.
+   * See {@link #_callDisposeWidgets} for more information.
+   */
+  protected _disposeWidgetsScheduledWidgetIds = new Set<string>();
```
cguglielmo: the name is a little cumbersome... Maybe call it widgetIdsBeingDisposed and the other one _widgetsToBeDisposed?

eclipse-scout-core/src/desktop/hybrid/HybridManager.ts
fschinkel: Added null check.
How are forms that are destroyed in Scout JS disposed on the server?

PR #2090, opened 2026-03-12
Security: Force-stop user's sessions when changing password

rogersteblerbsi:
Force-stop other active sessions when a user updates their password to prevent potential hijacked sessions from persisting.

The current session gets terminated as well, but it allows the user to save unsaved work before logout.

299509


org.eclipse.scout.rt.nls/src/main/resources/org/eclipse/scout/rt/nls/texts/ScoutTexts.properties
```diff
@@ -43,6 +43,7 @@ CancelButton=Cancel
 CancelShutdownAndReturnToTheApplication=Cancel logout and return to the application.
 CannotCreateBookmarkAtThisLocation=Favorite cannot be created at this location.
 ChangePassword=Change password
+ChangePasswordLogoutInfo=After changing your password, you will be logged out automatically.
```
rogersteblerbsi, line 46: TODO: Check texts and request translations

matthiaso (approved):

PR #2091, opened 2026-03-13
Remove deprecated code

cguglielmo:
427035


eclipse-scout-core/src/login/LoginBox.less
```diff
@@ -24,31 +24,6 @@
   & > .message-box {
     margin: 0 20px 12px 20px;
   }
-
-  & > input.alternative:-webkit-autofill {
```
cguglielmo, line 28: is this not necessary anymore?

eclipse-scout-core/src/style/animations.less
```diff
@@ -10,51 +10,30 @@
 #scout {
   /* args = keyframeName, timingFunction, iterationCount */
   .animation(@args) {
```
cguglielmo: I would say we delete these mixins because they are now obsolete.

eclipse-scout-core/src/style/mixins.less
```diff
@@ -109,16 +108,11 @@
 
   // noinspection CssUnknownProperty
   .transform(@args) {
```
cguglielmo: Same as for animations: delete the obsolete mixins

PR #2092, opened 2026-03-13
Reduce number of FutureCancelledError creations in BasicTransaction

rogersteblerbsi:
Initialize error object once and reuse it to improve performance

314695

### DIFF
--- a/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/transaction/BasicTransaction.java
+++ b/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/transaction/BasicTransaction.java
@@ -25,6 +25,12 @@ import org.slf4j.LoggerFactory;
 @SuppressWarnings("squid:S1181")
 public class BasicTransaction implements ITransaction {
   private static final Logger LOG = LoggerFactory.getLogger(BasicTransaction.class);
+  private static final FutureCancelledError FUTURE_CANCELLED_ERROR;
+
+  static {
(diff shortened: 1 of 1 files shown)


org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/transaction/BasicTransaction.java
```diff
@@ -25,6 +25,7 @@
 @SuppressWarnings("squid:S1181")
 public class BasicTransaction implements ITransaction {
   private static final Logger LOG = LoggerFactory.getLogger(BasicTransaction.class);
+  private static final FutureCancelledError FUTURE_CANCELLED_ERROR = new FutureCancelledError("Transaction cancelled");
```
matthiaso: Not sure if it is a good idea to throw an static error instance, see also: https://stackoverflow.com/questions/14489655/throwing-always-the-same-exception-instance-in-java

matthiaso (changes requested):

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/transaction/BasicTransaction.java
rogersteblerbsi: As discussed, the stack trace is probably not relevant for the `FutureCancelledError`. Additionally working with the stack trace from the static construction of the error does not make much sense. So we remove it completely.

matthiaso (approved):

PR #2093, opened 2026-03-13
Page: improve pagesForTableRows

fschinkel:
Page.pagesForTableRows is used e.g. when the OutlineMediator updates the order of child pages after a rowOrderChanged-event. If a page occurs multiple times in the list of pages for rows the order update of the child pages will fail. This may happen if there are multiple rows, linked to the same page (e.g. because the primary key is not unique among all rows and the child pages were created by the JsPageHelper). Therefore, the method Page.pagesForTableRows needs to remove all duplicate pages from its returned list.


bschwarzent (approved):
LGTM

PR #2102, opened 2026-03-17
Table: improve tile mode

fschinkel:
Table: improve tile mode

There may not always be a tile for a row. Therefore, implement Table.createTiles more lenient in order to avoid null pointer exceptions. In addition, the TableTileGridMediator ensures that null values are removed from its list of tiles.
The list of tiles of the TableTileGridMediator needs to be unique as the tiles will be rendered one after another. If there are duplicates in the list of tiles an 'Already rendered'-error will be thrown if the second occurrence is rendered. Therefore, the TableTileGridMediator removes all duplicates from its list of tiles.

379008

---

Align sort lookup call for table in tile mode

The Scout JS implementation of the sort lookup call for tables in tile mode uses lookup row icons to indicate whether a selected entry sorts ascending or descending while the Scout Classic implementation has added these icons to its text. Align these two implementations and use the style from Scout JS as it looks cleaner.

379008

### DIFF
--- a/eclipse-scout-core/src/table/TableTileGridMediator.ts
+++ b/eclipse-scout-core/src/table/TableTileGridMediator.ts
@@ -190,6 +190,7 @@ export class TableTileGridMediator extends Widget implements TableTileGridMediat
   }
 
   setTiles(tiles: ObjectOrChildModel<Tile>[]) {
+    tiles = [...new Set(tiles)].filter(Boolean); // remove duplicates and nulls
     this.setProperty('tiles', tiles);
   }
 
@@ -435,7 +436,7 @@ export class TableTileGridMediator extends Widget implements TableTileGridMediat
     this.table.loadingSupport.renderLoading(true);
   }
 
-  override destroy() {
+  protected override _destroy() {
     // destroy tiles manually since owner is the mediator thus the tileGrid can't destroy them
     this.tiles.forEach(tile => tile.destroy());
     this.tileAccordion.destroy();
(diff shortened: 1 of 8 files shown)


eclipse-scout-core/src/table/TableTileGridMediator.ts
```diff
@@ -190,6 +190,8 @@ export class TableTileGridMediator extends Widget implements TableTileGridMediat
   }
 
   setTiles(tiles: ObjectOrChildModel<Tile>[]) {
+    tiles = arrays.ensure(tiles)
+      .filter((value, index, array) => value && array.indexOf(value) === index); // remove duplicates
```
cguglielmo: This is O(n^2), using a Set should be faster: [...new Set(tiles)]

eclipse-scout-core/src/table/TableTileGridMediator.ts
```diff
@@ -528,7 +530,8 @@ export class TableTileGridMediator extends Widget implements TableTileGridMediat
     if (!this.table.tileMode || $.isEmptyObject(this.tilesMap)) {
       return;
     }
-    this.tiles = this.table.rows.map(row => this.tilesMap[row.id]);
+    this.tiles = this.table.rows.map(row => this.tilesMap[row.id])
```
cguglielmo: See above

cguglielmo (commented):
There is a typo in the commit message:

ensures that null values _**are**_ removed from its list of tiles.

cguglielmo (approved):

PR #2089, opened 2026-03-12
HttpProxy fix when using session-specific cookies

paolobazzi:
HttpProxy.SpecificSessionCookieStore must use
ClientRunContext.withSession() to correctly setup a contex with the client session.

454436

PR #2103, opened 2026-03-18
HttpProxy fix when using session-specific cookies

paolobazzi:
HttpProxy.SpecificSessionCookieStore must use
ClientRunContext.withSession() to correctly setup a contex with the client session.

454436


paolobazzi:
Review see https://github.com/eclipse-scout/scout.rt/pull/2089

PR #2104, opened 2026-03-18
Add @TypeVersion to UiPreferencesDo

bschwarzent:
To support the migration of a persisted UiPreferencesDo in the future, an initial @TypeVersion was added.

454123

### DIFF
--- a/org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/prefs/UiPreferencesDo.java
+++ b/org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/prefs/UiPreferencesDo.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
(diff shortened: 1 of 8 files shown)


org.eclipse.scout.rt.api.data/src/main/java/org/eclipse/scout/rt/api/data/prefs/UiPreferencesDo.java
```diff
@@ -17,9 +17,12 @@
 import org.eclipse.scout.rt.api.data.table.TableClientUiPreferencesDo;
 import org.eclipse.scout.rt.dataobject.DoEntity;
 import org.eclipse.scout.rt.dataobject.DoList;
+import org.eclipse.scout.rt.dataobject.ScoutTypeVersions.Scout_25_2_001;
 import org.eclipse.scout.rt.dataobject.TypeName;
+import org.eclipse.scout.rt.dataobject.TypeVersion;
 
 @TypeName("scout.UiPreferences")
+@TypeVersion(Scout_25_2_001.class)
```
fschinkel: Use a new version. Otherwise, the migration handler will not be executed.

fschinkel (changes requested):

fschinkel (approved):

PR #2106, opened 2026-03-19
Introduce userobject instead of separate user ids

rogersteblerbsi:


org.eclipse.scout.rt.security.test/src/test/java/org/eclipse/scout/rt/security/AccessControlServiceTest.java
```diff
@@ -64,37 +66,37 @@ public void tearDown() {
   }
 
   @Test
-  public void testGetUserId() {
-    assertNull(m_accessControlService.getUserId(null));
+  public void testGetUser() {
+    assertNull(m_accessControlService.getUser(null));
```
rogersteblerbsi, line 71: Here the user object itself exists, but does not contain any values. In case we don't have any user information should the user object be null or should it be an empty object?

Probably `IAccessControlService#getUser` should directly return `null` in case `AbstractAccessControlService#extractUserId` is null

org.eclipse.scout.rt.security.test/src/test/java/org/eclipse/scout/rt/security/AccessControlServiceTest.java
rogersteblerbsi, line 127: Here `@RunWithSubject` runs the `SubjectStatement` which does not setup the user object. So the access control service returns a new `NonePermissionCollection` everytime because the user is null.

Maybe we also need to set the user object in the `SubjectStatement`. But there we currently don't have access to the `IAccessControlService`.

Other way would be to provide new way to override the old `AbstractAccessControlService#getUserIdForCacheKey` for tests. For example a protected `getUser` method, that returns `User.current()` and tests can override it and provide their own user object.

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/security/User.java
```diff
@@ -59,10 +59,12 @@ protected void setData(String property, Serializable value) {
     m_data.put(property, value);
   }
 
+  // FIXME PBZ USER read only nur für einmaliges readOnly markieren, ähnlich wie java.security.PermissionCollection.setReadOnly
```
rogersteblerbsi: Currently we may also need to enrich the User object when loading the client session.
This is needed because
`ClientSessionProvider#provide`
creates the ModelJob with the new, not initialized yet, session. So when applying the user info from the session in `ClientRunContext#withSession`, the user data is not available yet.
Maybe we need to check in this case if the session is already active in `ClientRunContext#createUserFromSession`, and if not, then we don't set it as read only yet and will seal it AFTER the session has been started instead.

org.eclipse.scout.rt.security/src/main/java/org/eclipse/scout/rt/security/IAccessControlService.java
```diff
@@ -16,22 +16,18 @@
...
  */
 public interface IAccessControlService extends IService {
 
-  /**
-   * @return current UserId extracted from current {@link Subject}
-   */
-  String getUserIdOfCurrentSubject();
-
   /**
    * @return current UserId extracted from the provided {@link Subject}
```
rogersteblerbsi: Update JavaDoc

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/authentication/DefaultAuthTokenSigner.java
```diff
@@ -44,7 +44,8 @@ protected byte[] getPrivateKey() {
   }
 
   protected String getDefaultUserId() {
-    return BEANS.get(IAccessControlService.class).getUserIdOfCurrentSubject();
+    // FIXME PBZ USER: we would like to use User.currentUserId(); here -> this is a problem when used in a filter which is before the HttpRunContextFilter which sets up User threadlocal
+    return BEANS.get(IAccessControlService.class).getUser(Subject.current()).getUserId();
```
rogersteblerbsi: Maybe in this special case we can just use the thread local and use the access control service as fallback in case it's not set yet.

Another way would be to lazy create the user object in `User#currentUserId` in case it does not exist yet. But I think it's better that the object should always be setup correctly first. We don't want to just assume/hope that the object will be created as soon as you access it.

org.eclipse.scout.rt.security.test/src/test/java/org/eclipse/scout/rt/security/AccessControlServiceTest.java
rogersteblerbsi, line 140: Same reason for test failure:

https://github.com/eclipse-scout/scout.rt/pull/2106/changes/9b2df42d88695a55a6fa7289f84804d780aabfa4#r3001892534

org.eclipse.scout.rt.security.test/src/test/java/org/eclipse/scout/rt/security/AccessControlServiceTest.java
rogersteblerbsi: `getUser` returns `null` now, if the subject is `null`

org.eclipse.scout.rt.security.test/src/test/java/org/eclipse/scout/rt/security/AccessControlServiceTest.java
rogersteblerbsi: `SubjectStatement` sets up a simple `User` object now

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/security/User.java
rogersteblerbsi: Client now also uses `IAccessControlService` to get the user object, instead of setting it up when the session is loaded.

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/context/RunContext.java
```diff
@@ -79,6 +80,7 @@ public class RunContext implements IAdaptable {
   protected String m_correlationId;
   protected PropertyMap m_propertyMap = new PropertyMap();
   protected Context m_openTelemetryContext;
+  protected User m_user;
```
paolobazzi: move variable up directly after m_subject

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/context/RunContext.java
paolobazzi, line 548: add m_user

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/context/RunContext.java
```diff
@@ -283,6 +286,22 @@ public RunContext withSubject(final Subject subject) {
...
+   * @see #withUser(User)
+   */
+  public User getUser() {
+    return m_user;
+  }
+
+  /**
+   * Associates this context with the given {@link User}.
+   */
+  public RunContext withUser(final User user) {
```
paolobazzi, line 299: add testcases in org.eclipse.scout.rt.platform.context.RunContextTest

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/security/User.java
```diff
@@ -0,0 +1,97 @@
...
+
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.eclipse.scout.rt.platform.Bean;
+import org.eclipse.scout.rt.platform.context.RunContext;
+
+@Bean
+public class User implements Serializable {
```
paolobazzi, line 20: add unit tests, e.g. get/with and readOnly on/off, access to CURRENT and currentUserId() if null etc.

org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/cancellation/CancellationResource.java
paolobazzi: fix javadoc

org.eclipse.scout.rt.security/src/main/java/org/eclipse/scout/rt/security/AbstractAccessControlService.java
```diff
@@ -126,81 +125,70 @@ protected ICacheBuilder<K, IPermissionCollection> createCacheBuilder() {
...
-  public List<ICacheInvalidationListener<Object, IPermissionCollection>> getInvalidationListeners() {
+  public List<ICacheInvalidationListener<User, IPermissionCollection>> getInvalidationListeners() {
     return m_invalidationListeners.list();
   }
 
-  protected ICacheValueResolver<K, IPermissionCollection> createCacheValueResolver() {
-    return this::execLoadPermissions;
+  protected ICacheValueResolver<User, IPermissionCollection> createCacheValueResolver() {
+    return user -> {
+      Assertions.assertTrue(user.isReadOnly(), "User must be read only before accessing the permission cache");
```
paolobazzi: use static import for Assertion.

paolobazzi: add unit test for this case

org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/services/common/security/PermissionsInvalidationNotificationListener.java
```diff
@@ -87,22 +87,27 @@ public boolean needsCommit() {
       return true;
     }
 
+    public ICacheEntryFilter<User, IPermissionCollection> getFilter() {
```
paolobazzi: change to protected

org.eclipse.scout.rt.server.session/src/main/java/org/eclipse/scout/rt/server/session/ServerSessionProviderWithCache.java
```diff
@@ -152,7 +154,9 @@ protected CompositeObject newSessionCacheKey(final String sessionId, final Subje
       return new CompositeObject(sessionId);
     }
     if (subject != null) {
-      return new CompositeObject(BEANS.get(IAccessControlService.class).getUserId(subject));
+      User user = BEANS.get(IAccessControlService.class).getUser(subject);
+      Assertions.assertTrue(user.isReadOnly(), "User must be read only before using it as cache key");
```
paolobazzi: use static import

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/ISessionVariable.java
```diff
@@ -11,5 +11,4 @@
 
 public interface ISessionVariable {
```
paolobazzi: remove empty interface? or move to contacts app as example?

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/authentication/DefaultAuthTokenSigner.java
```diff
@@ -44,7 +45,9 @@ protected byte[] getPrivateKey() {
   }
 
   protected String getDefaultUserId() {
-    return BEANS.get(IAccessControlService.class).getUserIdOfCurrentSubject();
+    // user thread local is set after client session has been created.
```
paolobazzi: add unit test for this behavior

paolobazzi (commented):
Looks good, see inputs for minor improvements.

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/security/User.java
rogersteblerbsi: `UserTest` added

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/ISessionVariable.java
rogersteblerbsi: Removed

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/authentication/DefaultAuthTokenSigner.java
rogersteblerbsi: Added `DefaultAuthTokenSignerTest`

org.eclipse.scout.rt.security/src/main/java/org/eclipse/scout/rt/security/AbstractAccessControlService.java
rogersteblerbsi: Added test case in `AccessControlServiceTest`

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/filter/UiServletUserBootstrapFilter.java
```diff
@@ -0,0 +1,78 @@
...
+  public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
+    HttpServletRequest httpRequest = (HttpServletRequest) request;
+    HttpServletResponse httpResponse = (HttpServletResponse) response;
+
+    try {
+      Subject subject = Subject.current();
+      if (subject != null && m_helper.getUserOnSession(httpRequest) == null) {
+        User user = ClientRunContexts.empty()
+            .withSubject(subject)
+            .withUser(BEANS.get(User.class).withUserId(BEANS.get(IAccessControlService.class).extractUserId(subject)).setReadOnly()) // dummy user to load the real user object
```
paolobazzi:
```suggestion
            .withUser(BEANS.get(User.class).withUserId(BEANS.get(IAccessControlService.class).extractUserId(subject)).setReadOnly()) // use ad-hoc user instance to load the real user object
```

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/filter/UiServletUserBootstrapFilter.java
```diff
@@ -0,0 +1,78 @@
...
+            .withUserAgent(UserAgents.createDefault())
+            .call(() -> BEANS.get(IAccessControlService.class).getUser(subject));
+
+        // cache user for this session (see above)
```
paolobazzi:
```suggestion
        // cache user for this HTTP session (see above)
```

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/filter/UiServletUserBootstrapFilter.java
```diff
@@ -0,0 +1,78 @@
...
+        m_helper.putUserOnSession(httpRequest, user);
+      }
+
+      chain.doFilter(request, response);
+    }
+    catch (Throwable t) {
+      // always invalidate session on any error
+      BEANS.get(ServletFilterHelper.class).doLogout(httpRequest);
+
+      if (t instanceof AccessForbiddenException) { // pass-through HTTP error 403
```
paolobazzi:
```suggestion
      if (t instanceof AccessForbiddenException) { // pass-through HTTP error 403, may be thrown by IAccessControlService when resolving user
```

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/UiServlet.java
```diff
@@ -68,6 +68,8 @@ public class UiServlet extends AbstractHttpServlet {
   private final HttpServletControl m_httpServletControl;
   private final UiThreadInterruption m_uiThreadInterruption;
 
+  protected ServletFilterHelper m_filterHelper = BEANS.get(ServletFilterHelper.class);
```
paolobazzi: move instantiation to constructor like other two members

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/filter/UiServletUserBootstrapFilter.java
```diff
@@ -0,0 +1,78 @@
...
+  protected ServletFilterHelper m_helper = BEANS.get(ServletFilterHelper.class);
+
+  @Override
+  public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
+    HttpServletRequest httpRequest = (HttpServletRequest) request;
+    HttpServletResponse httpResponse = (HttpServletResponse) response;
+
+    try {
+      Subject subject = Subject.current();
+      if (subject != null && m_helper.getUserOnSession(httpRequest) == null) {
```
paolobazzi, line 51: Add javadoc like:
```
// Requests excluded by AuthFilter run without Subject, exclude from resolving user
```

paolobazzi (approved):
See inputs for minor improvements

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/authentication/DefaultAuthTokenSigner.java
```diff
@@ -44,7 +44,7 @@ protected byte[] getPrivateKey() {
   }
 
   protected String getDefaultUserId() {
-    return BEANS.get(IAccessControlService.class).getUserIdOfCurrentSubject();
+    return User.currentUserId();
```
paolobazzi, line 51: log warn if null

org.eclipse.scout.rt.shared/src/main/java/org/eclipse/scout/rt/shared/authentication/DefaultAuthTokenSigner.java
```diff
@@ -67,6 +71,7 @@ public <T extends DefaultAuthToken> T createDefaultSignedToken(Class<T> tokenCla
     }
     String userId = getDefaultUserId();
     if (StringUtility.isNullOrEmpty(userId)) {
+      LOG.warn("No user id could be determined");
```
paolobazzi:
```suggestion
      LOG.warn("No user id could be determined, check user in current run context. A null-token is returned.");
```

org.eclipse.scout.rt.ui.html/src/main/java/org/eclipse/scout/rt/ui/html/context/ClientHttpRunContextProducer.java
```diff
@@ -0,0 +1,39 @@
...
+import org.eclipse.scout.rt.platform.security.User;
+import org.eclipse.scout.rt.server.commons.authentication.ServletFilterHelper;
+import org.eclipse.scout.rt.server.commons.context.HttpRunContextProducer;
+import org.eclipse.scout.rt.ui.html.filter.UiServletUserBootstrapFilter;
+
+/**
+ * Producer for {@link ClientRunContext} instances handling HTTP requests in Scout client.
+ */
+@Replace
+public class ClientHttpRunContextProducer extends HttpRunContextProducer {
```
paolobazzi, line 24: Do not use replace here, but similar to org.eclipse.scout.rt.server.context.ServerHttpRunContextFilter create a ClientHttpRunContextFilter which returns the ClientHttpRunContextProducer (without replace-annotation)

paolobazzi (approved):

PR #2075, opened 2026-03-05
Use request subject also for its cancellation (if no other is set)

matthiaso:
410378

PR #2109, opened 2026-03-23
Use request subject also for its cancellation (if no other is set)

matthiaso:
410378


matthiaso:
Cherry-pick of https://github.com/eclipse-scout/scout.rt/pull/2075

PR #2115, opened 2026-03-25
Remove IRemoteFileService.getRemoteFiles directory listing operation

matthiaso:
Non-serializable parameter was used already a while making it unusable with this parameter set anyway.

453953


org.eclipse.scout.rt.server/src/main/java/org/eclipse/scout/rt/server/services/common/file/RemoteFileService.java
```diff
@@ -186,11 +185,6 @@ private String[][] getFiles(String folderBase, FilenameFilter filter) {
     return retVal;
   }
 
-  @Override
-  public RemoteFile[] getRemoteFiles(String folderPath, FilenameFilter filter, RemoteFile[] existingFileInfoOnClient) {
-    return getRemoteFiles(folderPath, filter, existingFileInfoOnClient, StandardCharsets.UTF_8.name());
-  }
-
   public RemoteFile[] getRemoteFiles(String folderPath, FilenameFilter filter, RemoteFile[] existingFileInfoOnClient, String charsetName) {
```
avandorp: Why don't we remove the other two getRemoteFiles methods? They are not in the interface, but also not used otherwise.

matthiaso: Thanks for the input, I just removed three additional methods (two mentioned above and one private now unused).

rogersteblerbsi (approved):

PR #2117, opened 2026-03-26
Outline: add keystroke to reload the root page

bschwarzent:
If a user wishes to recreate a page (e.g. because its content has been changed but the page offers no means to be refreshed), they can usually just reload the detail table of the parent page. However, since the outline in JS has no root page, this is not possible for top-level pages. For a pure JS outline, this can be resolved by recreating the entire application (reload page in browser, Ctrl-R), but for classic outlines, the server-side model has to be updated.

A new keystroke (F5) provides this function with the help of a hybrid action. It is only active for remote outlines, and only when no page is selected.

411299

### DIFF
--- a/eclipse-scout-core/src/desktop/outline/keystrokes/OutlineReloadKeyStroke.ts
+++ b/eclipse-scout-core/src/desktop/outline/keystrokes/OutlineReloadKeyStroke.ts
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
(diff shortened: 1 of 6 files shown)


eclipse-scout-core/src/desktop/outline/keystrokes/OutlineReloadKeyStroke.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  override handle(event: JQuery.KeyboardEventBase) {
+    // noinspection JSIgnoredPromiseFromCall
+    this._handleAsync(event);
+  }
+
+  protected async _handleAsync(event: JQuery.KeyboardEventBase): Promise<void> {
+    this._reloading = true;
+    try {
+      let hybridManager = await HybridManager.get();
```
fschinkel: nothing to await here

fschinkel (changes requested):

fschinkel (approved):

PR #2129, opened 2026-04-02
Features/cgu/25.2/organize columns 454064

cguglielmo:


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/columns/AbstractColumn.java
```diff
@@ -1502,6 +1502,11 @@ public boolean isEmpty() {
   }
 
   public void setColumnIndexInternal(int index) {
+    int currentColIndex = getColumnIndex();
+    if (currentColIndex >= 0) {
+      LOG.warn(null, new IllegalArgumentException("Unexpected column index " + currentColIndex + " for column " + this));
```
bschwarzent: Maybe print infos in log message instead of exception to be able to use {} placeholders?

```suggestion
      LOG.warn("Unexpected column index {} for column {}", currentColIndex, this, new Exception("stacktrace for further analysis"));
```

bschwarzent (approved):

PR #2132, opened 2026-04-07
PageWithTable: fix error on drill-down to JS page with search required

bschwarzent:
If a JS table page is marked with searchRequired=true and the user activates it with a double click on the corresponding row in the parent node's detail table, an error occurred in _resetTableData(), because the detail table was not yet created. This has been fixed by calling ensureDetailTable() before calling _resetTableData().

457043

### DIFF
--- a/eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
+++ b/eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
@@ -223,7 +223,18 @@ export class PageWithTable extends Page implements PageWithTableModel {
     this._resetTableData();
   }
 
+  /**
+   * Reverts the data in the {@link detailTable} to the original state by reloading it. For tables with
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
```diff
@@ -223,6 +223,13 @@ export class PageWithTable extends Page implements PageWithTableModel {
     this._resetTableData();
   }
 
+  /**
+   * Reverts the data in the {@link detailTable} to the original state by reloading it. For tables with
+   * {@link searchRequired} = true, the data is not reloaded. Instead, the old rows are deleted, a table
+   * status is displayed and {@link searchFilterCompleted} is set to false.
+   *
+   * This method must only be called after the detail table has been created ({@link ensureDetailTable}).
+   */
```
fschinkel, line 232: Please add null check for `this.detailTable` (see e.g. `PageWithTable.loadChildren`)

fschinkel (approved):

PR #2141, opened 2026-04-10
405136 add scout.mom.jms.requestMaxMessageSize setting

simon-bsi:
* if the scout.mom.jms.requestMaxMessageSize for a message is reached, an exception is thrown.


org.eclipse.scout.rt.mom.jms/src/main/java/org/eclipse/scout/rt/mom/jms/JmsMessageWriter.java
```diff
@@ -91,10 +98,18 @@ public JmsMessageWriter writeTransferObject(final Object transferObject) throws
 
     switch (m_marshaller.getMessageType()) {
       case MESSAGE_TYPE_TEXT:
-        writeTextMessage((TextMessage) m_message, (String) transportObject);
+        String text = (String) transportObject;
+        if (m_maxMessageSize > 0 && text != null && m_maxMessageSize < text.length()) {
```
lardser, line 102: I don't think we can (reliably) derive the actual message size from the text.length() since some charactes of _text_ might need more bytes when encoded than others.

lardser (changes requested):

PR #2145, opened 2026-04-15
HTML tooltip on form fields

seesack:


eclipse-scout-core/src/form/fields/FormField.ts
```diff
@@ -486,6 +488,16 @@ export class FormField extends Widget implements FormFieldModel {
     this._updateTooltip();
   }
 
+  /** @see FormFieldModel.tooltipText */
```
cguglielmo, line 491: The JS doc should link tooltipHtmlEnabled not tooltipText

eclipse-scout-core/src/form/fields/FormField.ts
```diff
@@ -486,6 +488,16 @@ export class FormField extends Widget implements FormFieldModel {
...
+  setTooltipHtmlEnabled(tooltipHtmlEnabled: boolean) {
+    this.setProperty('tooltipHtmlEnabled', tooltipHtmlEnabled);
+  }
+
+  /** @internal */
+  _renderTooltipHtmlEnabled() {
```
cguglielmo, line 497: Use protected instead and remove the internal JS doc because it is not used from outside of the class.

eclipse-scout-core/src/form/fields/FieldStatus.ts
```diff
@@ -201,6 +203,7 @@ export class FieldStatus extends Widget implements FieldStatusModel {
         parent: this,
         $anchor: this.$container,
         text: this.status.message,
+        htmlEnabled: this.tooltipHtmlEnabled,
```
cguglielmo, line 206: Also call the setter after line 196 so it can be changed while the tooltip is open.

cguglielmo (commented):
- Please add specs
- The SequenceBox pulls the tooltip from the last field up, this property should be pulled up as well, otherwise the tooltip content would be encoded if the field is the last field in a SequenceBox.
- Please add a checkbox for the new property in the [JS-Widgets-App](https://github.com/eclipse-scout/scout.docs/blob/releases/26.1/code/widgets/org.eclipse.scout.jswidgets.ui.html/src/main/js/formfield/FormFieldPropertiesBox.ts)

The field status creation changed with 26.1. I created a PR for 26.2 with the necessary adjustments and also added specs. However, the changes for the SequenceBox are still open: https://github.com/eclipse-scout/scout.rt/pull/2158

PR #2148, opened 2026-04-17
Use reentrant semaphore lock for BoundedResolveCacheWrapper

matthiaso:
One (and the same thread) should not acquire multiple semaphores, if it holds already the semaphore it may continue w/o semaphore.

419260

### DIFF
--- a/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/cache/BoundedResolveCacheWrapper.java
+++ b/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/cache/BoundedResolveCacheWrapper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -13,8 +13,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.Semaphore;
-
-import org.eclipse.scout.rt.platform.util.concurrent.ThreadInterruptedError;
+import java.util.function.Supplier;
 
 /**
  * This wrapper bounds the maximum concurrent resolve operation of a cache. In case a waiting operation is interrupted,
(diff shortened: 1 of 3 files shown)


org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/cache/AbstractCacheWrapper.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
```
matthiaso:
```suggestion
 * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
```

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/cache/BoundedResolveCacheWrapper.java
```diff
@@ -38,20 +43,7 @@ public V get(K key) {
     if (value != null) {
       return value;
     }
-    try {
-      m_semaphore.acquire();
-    }
-    catch (InterruptedException e) {
-      // interrupted, mark thread again as interrupted and resolve without a semaphore anyway
-      Thread.currentThread().interrupt();
-      throw new ThreadInterruptedError("Interrupted during acquire", e);
```
rogersteblerbsi, line 47: Before we threw an error here, now we resolve the value without a semaphore anyway (as written in the comment). Is this ok?

The error was introduced in: https://github.com/eclipse-scout/scout.rt/commit/adfb0be7807f7900e5f7e10f34217a5bc218bbad#diff-1d3a845c4fa07c4df9f491f54cf936045002e69bf451999c538f672443e0bac3

rogersteblerbsi (approved):

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/cache/BoundedResolveCacheWrapper.java
matthiaso: I assumed this was alright; of course we could easily restore the previous behavior `get` throws an error on interruption; `getAll` does not throw an error on interruption.

However as the javadoc and comments also stated the different behavior, on interruption value will be resolved - and in my opinion both methods should behave the same way, I also browsed through the history, found the same commit and assumed it was changed by error to `ThreadInterruptedError` (even though this error did lastg a long time).

In every case: As the thread is still interrupted if the actual resolve operation is not simple (e.g. spawns a different thread or relies on I/O operations) it is probably going to fail anyways.

PR #2149, opened 2026-04-20
Improve performance of EventSupport

fschinkel:
Store the registered events in a Map where the single event type a
listener is registered for is the key. So when an event is triggered
not all listeners need to be checked whether they need to be triggered
but only those with the correct types. This gives a huge performance
boost when dealing with huge amounts of listeners and triggered events.
Consider e.g. a hybrid action creating several thousands of widgets.
Typically, there will be a listener for each of these widgets and the
HybridManager will trigger a widgetAdd-event for each of these widgets.
Storing the listeners in an array like the former implementation led to
the array of several thousand listeners being processed for every of the
several thousand widgets which is O(n^2).

458889

### DIFF
--- a/eclipse-scout-core/src/events/EventSupport.ts
+++ b/eclipse-scout-core/src/events/EventSupport.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -104,21 +104,36 @@ export class EventSupport {
     return deferred.promise();
   }
 
+  /**
+   * Adds the given {@link EventListener} if it is not present.
+   */
   addListener(listener: EventListener) {
-    this._eventListeners.push(listener);
+    arrays.pushSet(this._eventListeners, listener);
   }
 
+  /**
+   * Removes the given {@link EventListener} if it is present.
+   */
   removeListener(listener: EventListener) {
     arrays.remove(this._eventListeners, listener);
   }
 
+  /**
+   * Counts all {@link EventListener}s that match the given filters.
+   *
+   * The type-filter will match all listeners that are triggered by all given types.
+   * For example if there are two listeners registered with type 'lorem' and 'lorem ipsum dolor' the filter 'lorem' will match 2 listeners.
+   * In this example, the type-filter 'ipsum dolor' will match 1 listener like 'dolor' or 'dolor ipsum lorem'.
+   *
+   * The func-filter will match all listeners that are registered using this exact {@link EventHandler}.
+   */
   count(type?: string, func?: EventHandler): number {
     let count = 0;
     this._eventListeners.forEach(listener => {
-      if (type && type !== listener.type) {
+      if (type && (!listener.type || !arrays.containsAll(listener.type.split(' '), type.split(' ').filter(Boolean)))) {
         return;
       }
-      if (func && func !== listener.func) {
+      if (func && !(func === listener.func || func === listener.origFunc)) {
         return;
       }
       count++;
@@ -127,13 +142,19 @@ export class EventSupport {
   }
 
   /**
-   * @returns the event types of the registered listeners
+   * Returns a list of unique single types of all registered listeners.
+   * Example: If there are two listeners registered for 'lorem' and 'lorem ipsum dolor' the multi-type is split and ['lorem', 'ipsum', 'dolor'] is returned.
    */
   types(): string[] {
     let types = new Set<string>();
     for (const listener of this._eventListeners) {
-      if (listener.type) {
-        types.add(listener.type);
+      if (!listener.type) {
+        continue;
+      }
+      for (const type of listener.type.split(' ')) {
+        if (type) {
+          types.add(type);
+        }
       }
     }
     return Array.from(types);
--- a/eclipse-scout-core/src/events/EventSupport.ts
+++ b/eclipse-scout-core/src/events/EventSupport.ts
@@ -7,19 +7,13 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-import {arrays, Event, EventHandler, EventListener, objects, scout} from '../index';
+import {Event, EventHandler, EventListener, scout, strings} from '../index';
 import $ from 'jquery';
 
-export type EventSubTypePredicate = (type: Event, subType: string) => boolean;
-
 export class EventSupport {
-  protected _eventListeners: EventListener[];
-  protected _subTypePredicates: EventSubTypePredicate[];
-
-  constructor() {
-    this._eventListeners = [];
-    this._subTypePredicates = objects.createMap();
-  }
+  protected _eventListenersByType = new Map<string, Set<EventListener>>();
+  protected _eventListenersByHandler = new Map<EventHandler, Set<EventListener>>();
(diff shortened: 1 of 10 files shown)


eclipse-scout-core/src/events/EventSupport.ts
```diff
@@ -104,21 +104,34 @@ export class EventSupport {
...
 
+  /**
+   * Removes the given {@link EventListener} if it is present.
+   */
   removeListener(listener: EventListener) {
     arrays.remove(this._eventListeners, listener);
   }
 
+  /**
+   * Counts all {@link EventListener}s that match the given filters.
```
cguglielmo, line 203: I would add an empty line after 122 and after 125 to structure it better.

eclipse-scout-core/test/events/EventSupportSpec.ts
```diff
@@ -47,10 +43,70 @@ describe('EventSupport', () => {
...
+      expect(eventSupport.count(null, handler)).toBe(3);
+
+      eventSupport.trigger('foo');
+      eventSupport.trigger('foo');
+      eventSupport.trigger('bar');
+      eventSupport.trigger('ipsum');
+      expect(eventSupport.count(null, handler)).toBe(3);
+    });
+
+    it('does add a handler twice', () => {
```
cguglielmo: just 'add' instead of 'does add'

eclipse-scout-core/test/events/EventSupportSpec.ts
```diff
@@ -78,11 +134,74 @@ describe('EventSupport', () => {
...
+      expect(eventSupport.count(null, handler)).toBe(1);
+
+      eventSupport.one('bar', handler);
+      expect(eventSupport.count(null, handler)).toBe(2);
+
+      eventSupport.one('lorem ipsum dolor', handler);
+      expect(eventSupport.count(null, handler)).toBe(3);
+    });
+
+    it('does remove handler after event is triggered', () => {
```
cguglielmo: just 'remove' instead of 'does remove'

eclipse-scout-core/test/events/EventSupportSpec.ts
```diff
@@ -78,11 +134,74 @@ describe('EventSupport', () => {
...
+
+      eventSupport.trigger('bar');
+      expect(eventSupport.count(null, handler)).toBe(1);
+
+      // removes all events that were registered together
+      eventSupport.trigger('ipsum');
+      expect(eventSupport.count(null, handler)).toBe(0);
     });
 
+    it('does add a handler twice', () => {
```
cguglielmo: just 'add' instead of 'does add'

eclipse-scout-core/test/events/EventSupportSpec.ts
```diff
@@ -143,7 +262,439 @@ describe('EventSupport', () => {
...
+      expect(eventSupport.count('dolor', handler1)).toBe(2);
+
+      eventSupport.off('ipsum');
+      expect(eventSupport.count(null, handler1)).toBe(5);
+      expect(eventSupport.count('lorem ipsum dolor', handler1)).toBe(1);
+      expect(eventSupport.count('lorem', handler1)).toBe(4);
+      expect(eventSupport.count('ipsum', handler1)).toBe(1);
+      expect(eventSupport.count('dolor', handler1)).toBe(2);
+
+      // type need to match exact
```
cguglielmo: exact -> exactly

eclipse-scout-core/src/events/EventSupport.ts
```diff
@@ -71,25 +65,73 @@ export class EventSupport {
...
-      let funcMatches = (func === listener.func || func === listener.origFunc);
-      let typeMatches = (type === listener.type);
-      let remove = false;
-      if (func && type) {
-        remove = (funcMatches && typeMatches);
-      } else if (func) {
-        remove = funcMatches;
-      } else { // always type. all other cases have been checked above
-        remove = typeMatches;
+    // validates a listeners types against the given type-filter
```
cguglielmo: 'the listener's types' or 'the types of the listeners'

eclipse-scout-core/src/events/EventSupport.ts
```diff
@@ -126,93 +207,234 @@ export class EventSupport {
...
+
+    if (!relevantTypes.size) {
+      // there are no relevant listeners -> return
+      return;
+    }
+
+    // collect all listeners for the relevant types in a set in order to remove duplicates
+    const listeners = [...relevantTypes]
+      .map(t => this._getListenersByType(t))
+      .reduce((a, b) => new Set([...a, ...b]), new Set());
```
cguglielmo, line 299: Why do you create a new Set for each iteration? The first parameter is the accumulator, the second is the currentValue -> .reduce((accumulator, currentValue) => accumulator.add(currentValue), new Set());

Actually, I would just write new Set(listeners) on a separate line instead of using reduce. It is cleaner and probably even faster because the Set-constructor is probably optimized by the compiler and no callback function is necessary, even though it is probably negligible.

eclipse-scout-core/src/events/EventSupport.ts
```diff
@@ -126,93 +207,234 @@ export class EventSupport {
...
+    if (this._eventListenersByType.size === 0) {
+      return 0;
+    }
+
+    type = strings.trim(type);
+
+    if (!type && !func) {
+      // no type- and no func-filter -> combine all listener sets in order to remove duplicates (e.g. because a listener was added for multiple types) and return size
+      return [...this._eventListenersByType.values()]
+        .reduce((a, b) => new Set([...a, ...b]), new Set())
```
cguglielmo, line 221: see comment in trigger

eclipse-scout-core/src/events/EventSupport.ts
```diff
@@ -71,25 +65,73 @@ export class EventSupport {
...
+    // type set -> remove from all matching listenersByType
+    for (const t of this._splitTypes(type)) {
+      // get all listeners for type
+      const listenersByType = this._getListenersByType(t);
+
+      // remove all listeners matching the type predicate
+      for (const listener of [...listenersByType].filter(typePredicate)) {
+        listenersByType.delete(listener);
+
+        // remove from listeners for handlers
```
cguglielmo: listeners by handler

eclipse-scout-core/src/events/EventSupport.ts
fschinkel: The problem is, that this does not combine `EventListener` but `Set<EventListener>` and the accumulator and the current value are of the same type. So adding a `Set<EventListener>` to a `Set<EventListener>` results in a `Set<EventListener | Set<EventListener>>` .
What I want to write is:
```
const listeners = [...relevantTypes]
  .map(t => this._getListenersByType(t));
  .reduce((accumulator, currentValue) => accumulator.union(currentValue), new Set());
```
But union is not available for our target browser versions.

Alternatives are
```
// adding everything to the accumulator
const listeners = [...relevantTypes]
  .map(t => this._getListenersByType(t))
  .reduce((accumulator, currentValue) => {
    for (const listener of currentValue) {
      accumulator.add(listener);
    }
    return accumulator;
  }, new Set());
```

```
// building a large array and use it to create a set
const listeners = new Set([...relevantTypes]
  .map(t => [...this._getListenersByType(t)])
  .reduce((accumulator, currentValue) => [...accumulator, ...currentValue])
);
```

```
// creating a set and add each listener
const listeners = new Set();
for (const s of [...relevantTypes].map(t => this._getListenersByType(t))) {
  for (const l of s) {
    listeners.add(l);
  }
}
```

What do you think is best?

cguglielmo (approved):

PR #2151, opened 2026-04-20
Improvements for webpack build during karma

fschinkel:


eclipse-scout-chart/webpack.config.global.js
```diff
@@ -10,6 +10,9 @@
 
 const baseConfig = require('@eclipse-scout/cli/scripts/webpack-defaults');
 module.exports = (env, args) => {
+  // type checking is done by default build
+  args.typeCheck = 'false';
+
```
mvilliger, line 15: I think we should enhance webpack-defaults to be able to handle strings and booleans here.
The jsdoc also states type boolean|'fork' which would be wrong then?

eclipse-scout-cli/scripts/karma-defaults.js
```diff
@@ -134,3 +182,30 @@ function searchSpecEntryPoint(specEntryPoint) {
...
   return path.resolve('test', 'test-index.js');
 }
+
+function findScoutExternals() {
+  // check whether scout browser build exists
+  const scoutEsmPath = require.resolve('@eclipse-scout/core');
+  if (!scoutEsmPath.endsWith('.esm.js')) {
+    return [];
+  }
+  const scoutExternalPath = scoutEsmPath.substring(0, scoutEsmPath.length - 7) + '.js';
```
mvilliger: instead of magic number 7 create a const for the suffix and use suffix.length()

eclipse-scout-cli/scripts/karma-defaults.js
```diff
@@ -120,10 +142,36 @@ module.exports = (config, specEntryPoint) => {
...
+    // add source maps for scout externals if present
+    ...(
+      scoutExternals?.length
+        ? {
+          sourceMapLoader: {
+            onlyWithURL: true
+          }
+        }
+        : {}
+    )
```
mvilliger, line 151: maybe this block is more readable on one line? looks strange like this :)

eclipse-scout-cli/scripts/webpack-defaults.js
mvilliger: I think this type is not complete: it should be:
`boolean|'fork'|'auto'|'true'|'false'`, right?

eclipse-scout-core/package.json
```diff
@@ -30,11 +30,11 @@
     "exports": {
       ".": {
         "types": "./dist/d.ts/src/index.d.ts",
-        "import": "./dist/eclipse-scout-core.esm.js"
+        "default": "./dist/eclipse-scout-core.esm.js"
```
cguglielmo, line 33: Why is this necessary? "import" is the correct way to export an esm module

eclipse-scout-cli/scripts/webpack-defaults.js
```diff
@@ -31,7 +31,7 @@ const ts = require('typescript');
  * @param {[]} args.resDirArray an array containing directories which should be copied to dist/res
  * @param {object} args.tsOptions a config object to be passed to the ts-loader
  * @param {object} args.forkTypeCheckOptions a config object to be passed to the ForkTsCheckerWebpackPlugin
- * @param {boolean|'fork'} args.typeCheck
+ * @param {boolean|'fork'|'auto'|'true'|'false'} args.typeCheck
```
cguglielmo: Even though 'true' and 'false' are possible, it should not be called like this. real booleans are the preferred way just like for the other properties

eclipse-scout-core/webpack.config.testing.global.js
```diff
@@ -0,0 +1,41 @@
...
+const baseConfig = require('@eclipse-scout/cli/scripts/webpack-defaults');
+module.exports = (env, args) => {
+  // type checking is done by default build
+  args.typeCheck = false;
+
+  const config = baseConfig(env, args);
+
+  // This build creates resources that can directly be included in a html file without needing a build stack (webpack, babel etc.).
+  // The resources are available by a CDN that provides npm modules (e.g. https://www.jsdelivr.com/package/npm/@eclipse-scout/core)
+  return {
```
cguglielmo, line 14: Can we import from webpack.config.global.js and just set entry and add aseConfig.rewriteIndexImports('scout', 'testing'), to externals?

eclipse-scout-cli/scripts/webpack-defaults.js
```diff
@@ -620,7 +620,7 @@ function getModuleName() {
  */
 function rewriteIndexImports(newImport, excludedFolder) {
   // If an import ends by one of these names, it is considered an index import.
-  let indexImports = ['index', 'main/js'];
+  let indexImports = ['index', 'main/js', 'src'];
```
cguglielmo: Why is this necessary? In your commit comment you write "This is the case e.g. in modules like eclipse-scout-core and eclipse-scout-chart." This is not true. src imports added by the IDE may only happen from code outside the source folder which is the test folder. But this is irrelevant as the test code is not built. All code in src/testing correctly imports from 'index'.

eclipse-scout-core/package.json
fschinkel: Yes, it is correct to add esm modules to "import" but it is as well correct to add them to "default". The [documentation](https://nodejs.org/api/packages.html#conditional-exports) says _When using environment branches, always include a "default" condition where possible_. And if "default" is set and contains an esm module, "import" can be omitted.
It is necessary to set "default" as we are trying to resolve the modules _@eclipse-scout/core_ and _@eclipse-scout/core/testing_ during the webpack build. This only looks for commonjs modules and will fail if there is no "default" or "require". As we do not want to use it, we do not care if "default" points to an esm module in this case but it needs to point somewhere in order to derive the global build.

eclipse-scout-cli/scripts/webpack-defaults.js
fschinkel: But it is called with the string literals. Running a script like `scout-scripts build:dev --typeCheck false` passes the argument as string and not as boolean. So this change enhanced the function to work with real booleans in case they are set in a _webpack.config.js_-file and @mvilliger asked to [add all possible input values to the js doc](https://github.com/eclipse-scout/scout.rt/pull/2151#discussion_r3121938921).

eclipse-scout-cli/scripts/webpack-defaults.js
fschinkel: Yes you are right, this is no longer necessary. This was a prerequisite for using the global builds as external in _@eclipse-scout/core_, which was not implemented in the end.
So this not wrong but also not sufficient to use rewriteIndexImports for the specs of _@eclipse-scout/core_ and how their imports are written today. Do you want it do be removed as it is not used right now? Or do we leave it this way and skip half of the import migration work when we want to use the global builds for _@eclipse-scout/core_ as well?

PR #2152, opened 2026-04-20
DataColumn: add support for DateGroupType

bschwarzent:
269392


eclipse-scout-core/src/prefs/TableUiPreferences.ts
```diff
@@ -519,6 +542,10 @@ export class TableUiPreferences implements ObjectWithType {
       column.setBackgroundEffect(columnPreferences.backgroundEffectId as NumberColumnBackgroundEffect, false); // false = don't redraw
     }
 
+    if (column instanceof DateColumn) {
+      column.groupType = columnPreferences.dateGroupType;
```
cguglielmo: add setter?

eclipse-scout-core/src/table/columns/DateColumn.ts
```diff
@@ -99,4 +129,41 @@ export class DateColumn extends Column<Date> implements DateColumnModel {
       hasTime: this.hasTime
     });
   }
+
+  override compare(row1: TableRow, row2: TableRow): number {
```
cguglielmo, line 157: Is there a spec for this sorting?

eclipse-scout-core/src/table/columns/DateColumnTableHeaderMenu.ts
```diff
@@ -0,0 +1,363 @@
...
+        if (menu.groupType === currentGroupType) {
+          // already grouped by this type
+          return;
+        }
+        updateFilter(menu.groupType);
+      });
+    });
+
+    // Create context menu
+    this._contextMenu = scout.create(ContextMenuPopup, {
```
cguglielmo: Maybe don't save contextMenu as member but use action as parent. To get it, just use action.findChild(ContextMenuPopup), this should make owner check obsolete. Also maybe create DateGroupContextMenuPoup. Both things should reduce code

eclipse-scout-core/src/table/columns/DateColumnTableHeaderMenu.ts
```diff
@@ -0,0 +1,363 @@
...
+
+interface DateColumnWithFilterType {
+  /**
+   * Temporarily holds the selected group type of the filter table when no filter is active.
+   * Allows restoring the previous group type when the header menu is opened again.
+   */
+  __filterGroupType: DateGroupType;
+}
+
+// FIXME bsh: Add test
```
cguglielmo: obsolete because there are tests?

eclipse-scout-core/src/table/Table.ts
```diff
@@ -1648,89 +1648,88 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
           // column was not yet present: insert at determined position
           // and move all subsequent nodes by one.
           // add just after all other grouping columns in column set.
-          column.sortIndex = sortIndex + 1;
-          arrays.eachSibling(this.columns, column, siblingColumn => {
-            if (siblingColumn.sortActive && !(siblingColumn.initialAlwaysIncludeSortAtBegin || siblingColumn.initialAlwaysIncludeSortAtEnd) && siblingColumn.sortIndex > sortIndex) {
-              siblingColumn.sortIndex++;
+          sortedSiblingColumns.forEach(siblingColumn => {
+            if (siblingColumn.sortIndex > sortIndex) {
+              siblingColumn.sortIndex++; //
```
cguglielmo: remove empty comment or write one

eclipse-scout-core/src/table/Table.ts
```diff
@@ -1648,89 +1648,88 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
       return;
     }
+    if (column.grouped) {
+      return; // column is already grouped, nothing to do (otherwise, the sort index would be increased unnecessarily)
+    }
 
     direction = scout.nvl(direction, column.sortAscending ? 'asc' : 'desc');
     multiGroup = scout.nvl(multiGroup, true);
+
+    // do not update sort index for permanent head/tail sort columns, their order is fixed (see ColumnSet.java)
```
cguglielmo, line 1661: are the upcoming changes just a refactoring or what is the actual change and why was it necessary? -> add spec

eclipse-scout-core/src/table/TableHeaderMenu.ts
```diff
@@ -899,20 +908,20 @@ export class TableHeaderMenu extends Popup implements TableHeaderMenuModel {
     let sortMode = TableHeaderMenu.SortMode;
     if (this.filterSortMode === sortMode.ALPHABETICALLY) {
       // sort by amount
-      this.filterTable.sort(this.filterTable.columns[1], 'desc');
+      this.filterTable.sort(this.filterTable.columnById('AmountColumn'), 'desc');
+      this.filterTable.sort(this.filterTable.columnById('SortByColumn'), 'asc', true); // if two rows have the same amount, sort them by original order
```
cguglielmo, line 920: Maybe use sortIndex and sortAscending properties to not sort twice?

eclipse-scout-core/src/table/columns/DateColumn.ts
```diff
@@ -99,4 +129,41 @@ export class DateColumn extends Column<Date> implements DateColumnModel {
...
+    }
+    if (!value2) {
+      return 1;
+    }
+
+    if (this.grouped && this.groupType) {
+      let group = TableMatrix.resolveDateGroup(this.groupType);
+      if (group) {
+        let matrix = new TableMatrix(this.table);
+        let axis = matrix.addAxis(this, group);
```
cguglielmo: Does this have a performance impact on large tables if matrix is created on every compare?

eclipse-scout-core/src/table/TableMatrix.ts
```diff
@@ -89,14 +95,15 @@ export class TableMatrix {
     return dataAxis;
   }
 
-  // add x or y Axis
+  /**
+   * add key axis (x or y)
```
cguglielmo:
```suggestion
   * Adds key axis (x or y)
```

eclipse-scout-core/src/table/columns/DateColumnTableHeaderMenu.ts
```diff
@@ -0,0 +1,363 @@
...
+      let isOwnContextMenu = this._contextMenu.anchor === anchor;
+      this._contextMenu.close();
+      if (isOwnContextMenu) {
+        return; // toggle only -> done
+      }
+    }
+
+    // If button was already selected, just ungroup the column (don't show the context menu)
+    if (anchor instanceof TableHeaderMenuButton && anchor.selected) {
+      this.table.group(this.column, undefined, undefined, true); // ungroup
```
cguglielmo: You can use removeGroupColumn

eclipse-scout-core/src/table/columns/DateColumnModel.ts
```diff
@@ -1,17 +1,18 @@
...
  *
  * SPDX-License-Identifier: EPL-2.0
  */
-import {ColumnModel, DateFormat} from '../../index';
+import {ColumnModel, DateFormat, DateGroupType} from '../../index';
 
 export interface DateColumnModel extends ColumnModel<Date> {
   format?: DateFormat | string;
   groupFormat?: DateFormat | string;
+  groupType?: DateGroupType;
```
cguglielmo, line 43: Add JsDoc and describe what it is, what the difference is to groupFormat and when to use which one. Please also add JsDoc for format and groupFormat if possible.

eclipse-scout-core/src/table/columns/DateColumnTableHeaderMenu.ts
```diff
@@ -0,0 +1,363 @@
...
+      this.close();
+      return;
+    }
+
+    // Create menus
+    let menus = this._createGroupTypeMenus();
+
+    let specialGroupFormat = this.column.groupFormat.pattern;
+    let hasSpecialGroupFormat = specialGroupFormat !== 'yyyy';
```
cguglielmo: If the specialGroupFormat is a pattern used by an existing group type, shouldn't this groupType be used instead of creating a new entry? For example, I think it is confusing if there is an additional entry with a pattern 'dd.MM.yyyy' even though it is actually the same as the date groupType

eclipse-scout-core/src/prefs/TableUiPreferences.ts
bschwarzent: OK. Added test in TableUiPreferencesSpec.

eclipse-scout-core/src/table/columns/DateColumn.ts
bschwarzent: Explicit test added to DateColumnSpec.

eclipse-scout-core/src/table/columns/DateColumnTableHeaderMenu.ts
bschwarzent: Done.

eclipse-scout-core/src/table/Table.ts
bschwarzent: When grouped columns exist and another columns is defined as the new single-grouped column (multiGroup=false), the sorting should be disabled for the old grouped columns. This aligns the behavior with that of the Scout Classic (Java) tables. Test added in TableGroupingSpec.

eclipse-scout-core/src/table/TableHeaderMenu.ts
bschwarzent: Left as is, because the table API currently does not provide this function.

eclipse-scout-core/src/table/columns/DateColumn.ts
bschwarzent: No, TableMatrix constructor and addAxis don't handle the table data.
Small optimization: when groupType is set, an exist is created and stored in a re-usable member.

cguglielmo (approved):

eclipse-scout-core/src/table/columns/DateColumnTableHeaderMenu.ts
bschwarzent: More matching patterns are now converted to the corresponding enum value -> see DateColumnTableHeaderMenu._getGroupTypeForFormatPattern

PR #2154, opened 2026-04-21
Features/cgu/26.2/446489 generic entity choose step

cguglielmo:


eclipse-scout-core/src/form/fields/pagefield/PageField.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  constructor() {
+    super();
+
+    this._addWidgetProperties(['outline']);
+    this._searchFormHandler = this._onPageSearchFormChange.bind(this);
+    this._detailTableHandler = this._onPageDetailTableChange.bind(this);
+  }
+
+  override _jsonModel(): PageFieldModel<TPage> {
```
fschinkel: protected

eclipse-scout-core/src/form/fields/pagefield/PageField.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  setOutline(outline: ObjectOrChildModel<Outline>) {
+    this.setProperty('outline', outline);
+  }
+
+  protected _setOutline(outline: Outline) {
+    outline = scout.nvl(outline, scout.create(Outline, {parent: this}));
+    if (outline?.nodes.length > 1) {
+      throw new Error('Outline must not have more than one root page');
```
fschinkel, line 86: outline can not be `null` or `undefined` at this point

eclipse-scout-core/src/form/fields/pagefield/PageField.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+}
+
+export interface PageFieldModel<TPage extends Page> extends GroupBoxModel {
+  page?: ObjectType<TPage> | ObjectOrChildModel<TPage>;
+}
+
+export interface PageFieldEventMap<TPage extends Page = Page> extends GroupBoxEventMap {
+  'propertyChange:page': PropertyChangeEvent<TPage>;
+}
```
fschinkel, line 221: `'propertyChange:outline'` is missing

eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
```diff
@@ -224,6 +224,9 @@ export class PageWithTable extends Page implements PageWithTableModel {
   }
 
   protected _resetTableData() {
+    if (!this.detailTable) {
+      return;
```
fschinkel, line 236: What about `this.setSearchFilterCompleted(false)`? Is it not needed when there is no table?

eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
```diff
@@ -479,6 +482,11 @@ export class PageWithTable extends Page implements PageWithTableModel {
    * @param restoreSelectionInfo information needed to restore the selection after table data was loaded
    */
   protected _onLoadTableDataDone(tableData: any, restoreSelectionInfo?: RestoreSelectionInfo) {
+    if (!this.detailTable) {
+      // table has been removed in the meantime
```
fschinkel, line 502: What about the load event? Is it not needed when there is no table?

eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
```diff
@@ -512,6 +520,11 @@ export class PageWithTable extends Page implements PageWithTableModel {
   }
 
   protected _onLoadTableDataFail(error: any, restoreSelectionInfo?: RestoreSelectionInfo) {
+    if (!this.detailTable) {
+      // table has been removed in the meantime
```
fschinkel, line 540: What about the error event? Is it not needed when there is no table?

eclipse-scout-core/test/form/fields/pagefield/PageFieldSpec.ts
```diff
@@ -0,0 +1,234 @@
...
+  });
+
+  describe('search form field', () => {
+    it('shows search form of page', () => {
+      let pageField = scout.create(PageField, {parent: session.desktop});
+      expect(pageField.searchFormField.innerForm).toBe(null);
+
+      pageField.setPage({objectType: SearchFormPage});
+      expect(pageField.searchFormField.innerForm).toBe((pageField.page.detailTable.tableControls[0] as SearchFormTableControl).form);
+
```
fschinkel, line 156: maybe use `table.findTableControl(SearchFormTableControl)`, see also the other occurrences in this spec

eclipse-scout-core/src/form/fields/pagefield/PageField.ts
```diff
@@ -0,0 +1,169 @@
...
+    this._updateSearchForm(event.newValue);
+  }
+
+  protected _onPageDetailTableChange(event: PropertyChangeEvent<Table>) {
+    this._updateDetailTable(event.newValue, event.oldValue);
+  }
+
+  protected _updateDetailTable(detailTable: Table, oldDetailTable?: Table) {
+    let oldSearchFormControl = oldDetailTable?.tableControls.find(control => control instanceof SearchFormTableControl);
+    oldSearchFormControl?.off('propertyChange:form', this._searchFormHandler);
```
fschinkel, line 137: maybe use `table.findTableControl(SearchFormTableControl)`

eclipse-scout-core/src/form/fields/pagefield/PageField.ts
```diff
@@ -0,0 +1,169 @@
...
+    this._setProperty('outline', outline);
+
+    this.outline.setNavigateButtonsVisible(false);
+
+    // Ensure outline is not set to compact and does not embed detail content which would happen if desktop.displayStyle is set to COMPACT.
+    // Otherwise, detail menus and table controls would be modified (TableControlAdapterMenu), but the page field should display the table and search form as they are.
+    this.outline.setCompact(false);
+    this.outline.setEmbedDetailContent(false);
+
+    this._setPage(this.outline.nodes[0] as TPage || this.page);
```
fschinkel, line 100: Is there a way a `Page` can be transferred from one `Outline` to another? If so, the page might already be set.
What about all other events from the `Outline`? E.g. if the `Outline` deletes its child nodes. In this case the `Page` is still set and destroyed. Insert events adding more than the allowed single page are not prevented. Is it really clever to expose the `Outline`?

eclipse-scout-core/src/table/controls/TableControl.ts
```diff
@@ -216,8 +216,20 @@ export class TableControl extends Action implements TableControlModel {
   }
 
   protected _setSelected(selected: boolean) {
-    // Does nothing more than the default but allows for extension by a subclass
     this._setProperty('selected', selected);
+    this._updateSelected();
+  }
+
+  protected _setVisible(visible: boolean) {
+    this._setProperty('visible', visible);
```
fschinkel, line 224: Do we need to add the same for `enabled`?

eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
cguglielmo: Maybe, but it is not party of this change anymore.

eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
cguglielmo: I added this null check because if a page is destroyed while it is still loading, the promises failed because the table is set to null. I don't think a load event is expected if the table is null because no rows could be added.

eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
cguglielmo: see above

eclipse-scout-core/src/form/fields/pagefield/PageField.ts
cguglielmo: page field now handles the nodes deleted and inserted cases

eclipse-scout-core/src/table/controls/TableControl.ts
cguglielmo: You are right, we could do the same for enabled, but we don't really need it now. We needed it for selected because if searchRequired is true, the search is opened even if the table control is invisible.

fschinkel (approved):

PR #2165, opened 2026-04-27
Common mime type convenience method for audio types

seesack:


cguglielmo (commented):
I would rephrase the commit message like so: 

Add convenience method for common audio mime types

cguglielmo (approved):

PR #2171, opened 2026-04-29
Set default JAX-WS connect and read timeout

rogersteblerbsi:
Previously, defaults were infinite, which could cause hanging
connections and excessive resource usage.
Now the Connect timeout is set to 5 minutes
and the read timeout is set to 30 minutes.

382540

### DIFF
--- a/org.eclipse.scout.rt.server.jaxws/src/main/java/org/eclipse/scout/rt/server/jaxws/JaxWsConfigProperties.java
+++ b/org.eclipse.scout.rt.server.jaxws/src/main/java/org/eclipse/scout/rt/server/jaxws/JaxWsConfigProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


org.eclipse.scout.rt.server.jaxws/src/main/java/org/eclipse/scout/rt/server/jaxws/JaxWsConfigProperties.java
```diff
@@ -219,12 +219,12 @@ public String getKey() {
...
     @Override
     public String description() {
-      return "Connect timeout in milliseconds to abort a webservice request, if establishment of the connection takes longer than this timeout. A timeout of null means an infinite timeout. The default value is null.";
+      return "Connect timeout in milliseconds to abort a webservice request, if establishment of the connection takes longer than this timeout. A timeout of null means an infinite timeout. The default value is 30 minutes.";
     }
 
     @Override
     public Integer getDefaultValue() {
-      return null; // infinite timeout
+      return 30 * 60 * 1000; // 30 minutes
```
matthiaso: Not sure if 30 minutes is a quite high value to establish a connection; I would expect that a connection should usually be established in a much shorter timeframe.

matthiaso (approved):

org.eclipse.scout.rt.server.jaxws/src/main/java/org/eclipse/scout/rt/server/jaxws/JaxWsConfigProperties.java
rogersteblerbsi: Reduced to 5 min

PR #2179, opened 2026-05-05
PageWithTable: improve loading

fschinkel:
The PageWithTable marks its detailTable loading while it loads the data. Afterward, this flag is reset. This behaviour was implemented incorrectly as simply calling Table.setLoading does not ensure the table will stay loading during the whole time. The table itself contains a TableUpdateBuffer which holds promises and sets the table to loading=false after the last promise is resolved. Therefore, calling table.setLoading(true) without passing a promise to the TableUpdateBuffer may result in the table being loading=false before the page finished its loading, because in the meantime the last promise of the TableUpdateBuffer was resolved. Therefore, the PageWithTable now adds its load-promise to the TableUpdateBuffer and ensures that the table is marked with loading=true until the page finished loading. This also solves issues where rows and child pages are in an inconsistent state when the table was e.g. sorted. Consider a PageWithTable having at least one SmartColumn which resolves its cell text asynchronously. If this page is reloaded multiple times in a short period of time, the asynchronous cell text update of a former reload may finish while the latest page reload is still running. As the SmartColumn uses the TableUpdateBuffer this resulted in the table being marked loading=false before the latest page reload was finished. In addition, the TableUpdateBuffer ensures that e.g. the order of the rows is correct which will now sort rows of a previous reload (as the latest is not finished yet). If now the PageWithTable is an IJsPage that creates its child pages on the UI Server using the JsPageHelper the child pages created for the latest page reload may already be present. In this case the row order update is passed on to the Outline by the OutlineMediator which determines the new child page order using the rows and their linked pages. As the rows belong to a former page reload they are not linked to any pages and therefore, the OutlineMediator tries to update the child page order using an empty array. But as there are already child pages present the Outline throws an error because the order can not be updated correctly.
These issues are solved as the TableUpdateBuffer now waits until the page is loaded before e.g. updating the row order and if the page is loaded it is ensured that child pages and rows are linked correctly.

456855

### DIFF
--- a/eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
+++ b/eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
@@ -415,9 +415,15 @@ export class PageWithTable extends Page implements PageWithTableModel {
     // Ensure the search form data is in sync with the search filter used to load the table data (e.g. required for getSearchFilterText())
     this._updateSearchData(searchFilter);
 
-    return this._loadTableData(searchFilter)
+    const promise = this._loadTableData(searchFilter)
       .then(data => this._onLoadTableDataDone(data, restoreSelectionInfo))
       .catch(error => this._onLoadTableDataFail(error, restoreSelectionInfo));
+
+    // Ensure loading of the detail table is set to false once all loading tasks have been completed.
+    // This includes loading the main table data but also column lookup calls.
+    this.detailTable.updateBuffer.pushPromise(promise);
+
+    return promise;
   }
 
   protected _updateSearchData(searchFilter: any) {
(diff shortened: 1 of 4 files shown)


eclipse-scout-core/src/desktop/outline/pages/PageWithTable.ts
```diff
@@ -415,9 +415,13 @@ export class PageWithTable extends Page implements PageWithTableModel {
     // Ensure the search form data is in sync with the search filter used to load the table data (e.g. required for getSearchFilterText())
     this._updateSearchData(searchFilter);
 
-    return this._loadTableData(searchFilter)
+    const promise = this._loadTableData(searchFilter)
       .then(data => this._onLoadTableDataDone(data, restoreSelectionInfo))
       .catch(error => this._onLoadTableDataFail(error, restoreSelectionInfo));
+
+    this.detailTable.updateBuffer.pushPromise(promise);
```
cguglielmo, line 424:
```suggestion
  // Ensure loading of the detail table is set to false once all loading tasks have been completed; this includes loading the main table data but also column lookup calls.
  this.detailTable.updateBuffer.pushPromise(promise);
```

cguglielmo (approved):

PR #2181, opened 2026-05-05
Security provider: separate hash creation and hash verification

torstentrompler:
The current implementation of the verifyPasswordHash method is costly
when comparing a password against multiple hashes, because the password
hash will be created multiple times.

If the hash creation and hash verification are separated, the hash only
needs to be created once, when comparing a password against multiple
hashes.

372108

### DIFF
--- a/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/security/SunSecurityProvider.java
+++ b/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/security/SunSecurityProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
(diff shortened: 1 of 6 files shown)


org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/security/SunSecurityProvider.java
```diff
@@ -250,23 +253,32 @@ public byte[] createPasswordHash(char[] password, byte[] salt, int iterations) {
 
   @Override
   public boolean verifyPasswordHash(char[] password, byte[] salt, byte[] expectedHash) {
-    if (Arrays.equals(expectedHash, createPasswordHash(password, salt, MIN_PASSWORD_HASH_ITERATIONS))) {
+    return verifyPasswordHash(password, salt, Set.of(expectedHash));
```
andibur: Throws NPE in case expectedHash is null. false was returned before this change:

```suggestion
    return expectedHash != null && verifyPasswordHash(password, salt, Set.of(expectedHash));
```

PR #2187, opened 2026-05-08
Fix handling of number affixes

annebre:
Make sure that affixes of numbers are detected even with whitespace so that the numbers can be normalized and parsed correctly.

425842

### DIFF
--- a/eclipse-scout-core/src/text/DecimalFormat.ts
+++ b/eclipse-scout-core/src/text/DecimalFormat.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -17,10 +17,12 @@ import {Locale, numbers, RoundingMode, scout, strings} from '../index';
  * - %
  */
 export class DecimalFormat {
+
   positivePrefix: string;
   positiveSuffix: string;
   negativePrefix: string;
   negativeSuffix: string;
+  protected _trimmedAffixReplacements: AffixReplacement[];
   groupingChar: string;
   lenientGroupingChars: string;
   groupLength: number;
@@ -92,6 +94,24 @@ export class DecimalFormat {
         this.negativeSuffix = this.negativeSuffix + this.positiveSuffix;
       }
     }
+
+    // The normalize function needs trimmed affixes.
+    this._trimmedAffixReplacements = [
+      {affix: strings.trim(this.positivePrefix), replacement: '+'},
+      {affix: strings.trim(this.positiveSuffix), replacement: ''},
+      {affix: strings.trim(this.negativePrefix), replacement: '-'},
+      {affix: strings.trim(this.negativeSuffix), replacement: ''}
+    ];
+    // Sort by the affix length (descending), so that shorter affixes possibly contained in longer ones are replaced later.
+    this._trimmedAffixReplacements.sort((a, b) => {
+      const lengthDiff = strings.length(b.affix) - strings.length(a.affix);
+      if (lengthDiff) {
+        return lengthDiff;
+      }
+      // If there are two affixes with the same length, they might be equal.
+      // In that case, handle the one with the replacement first: favor an error caused by an incorrect affix placement (e.g. '-' as a suffix) instead of a silent error.
+      return strings.length(a.replacement) > 0 ? -1 : 1;
+    });
 
     // find group length
     let posDecimalSeparator = this.pattern.indexOf(SYMBOLS.decimalSeparator);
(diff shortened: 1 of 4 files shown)


eclipse-scout-core/src/text/DecimalFormat.ts
```diff
@@ -17,10 +17,18 @@ import {Locale, numbers, RoundingMode, scout, strings} from '../index';
...
 export class DecimalFormat {
+
+  static START_OF_THE_LINE_CHAR = '^';
+  static END_OF_THE_LINE_CHAR = '$';
+
   positivePrefix: string;
   positiveSuffix: string;
   negativePrefix: string;
   negativeSuffix: string;
+  positivePrefixTrimmed: string;
```
fschinkel: Is it necessary to add these properties because trimming is to expensive? Or may we trim these prefixes and suffixes in the `normalize` function when they are used.

eclipse-scout-core/src/text/DecimalFormat.ts
```diff
@@ -241,33 +259,36 @@ export class DecimalFormat {
...
       return numberString;
     }
-    let result = numberString
-      .replace(new RegExp('[' + this.groupingChar + this.lenientGroupingChars + ']', 'g'), '')
-      .replace(new RegExp('[' + this.decimalSeparatorChar + ']', 'g'), '.');
 
-    if (strings.hasText(this.positivePrefix)) {
-      result = result.replace(new RegExp(this.positivePrefix, 'g'), '+');
+    let result = numberString.trim();
+    if (strings.hasText(this.positivePrefixTrimmed)) {
```
fschinkel: Why are you removing only the first and the last occurrence? This breaks e.g. some of the calculation features of the `NumberField`. Consider a format _#.#km_ and the input `2km + 5km`. This will lead to a parsing error as it needs to be written as `2 + 5km`. Also `2km * 4` or `8km / 2` will not work.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/form/fields/numberfield/AbstractNumberField.java
```diff
@@ -524,17 +525,52 @@ protected BigDecimal roundParsedValue(BigDecimal valBeforeRounding) {
...
+      text = StringUtility.trim(text.substring(0, text.length() - trimmedPositiveSuffix.length()));
+    }
+    else if (StringUtility.hasText(trimmedNegativeSuffix) && text.endsWith(trimmedNegativeSuffix)) {
+      text = StringUtility.trim(text.substring(0, text.length() - trimmedNegativeSuffix.length()));
+      isNegative = true;
+    }
+
+    // Add the prefix and suffix.
+    if (isNegative) {
+      text = StringUtility.addPrefix(text, negativePrefix);
```
fschinkel: You can use `StringUtility.box` to add prefix and suffix at once.

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/form/fields/numberfield/AbstractNumberField.java
```diff
@@ -524,17 +525,52 @@ protected BigDecimal roundParsedValue(BigDecimal valBeforeRounding) {
     return valBeforeRounding.round(new MathContext(precision, getRoundingMode()));
   }
 
-  private String ensureSuffix(String text) {
+  /**
+   * Returns the given text with the prefix and suffix in the correct format.
+   * If the text contains a prefix/suffix (possibly with additional/missing whitespace), it is removed.
+   * Adds the correct prefix and suffix.
+   */
+  private String ensurePrefixAndSuffix(String text) {
```
fschinkel, line 533: In Scout JS both negative and positive suffixes/prefixes are removed. Why do we only remove negative or positive ones here?

eclipse-scout-core/src/text/DecimalFormat.ts
```diff
@@ -93,6 +98,27 @@ export class DecimalFormat {
...
+    // The normalize function needs trimmed affixes.
+    this._trimmedAffixReplacements = [
+      {affix: strings.trim(this.positivePrefix), replacement: '+'},
+      {affix: strings.trim(this.positiveSuffix), replacement: ''},
+      {affix: strings.trim(this.negativePrefix), replacement: '-'},
+      {affix: strings.trim(this.negativeSuffix), replacement: ''}
+    ];
+    // Sort by the affix length (descending), so that shorter affixes possibly contained in longer ones are replaced later.
+    this._trimmedAffixReplacements.sort((a, b) => {
+      let lengthA = strings.length(a.affix);
```
fschinkel:
```suggestion
      const lengthDiff = strings.length(b.affix) - strings.length(a.affix);
      if (lengthDiff) {
        return lengthDiff;
      }
```

eclipse-scout-core/src/text/DecimalFormat.ts
```diff
@@ -17,10 +17,15 @@ import {Locale, numbers, RoundingMode, scout, strings} from '../index';
  * - %
  */
 export class DecimalFormat {
+
+  static START_OF_THE_LINE_CHAR = '^';
```
fschinkel: not used

fschinkel (approved):

PR #2193, opened 2026-05-12
Security provider: separate hash creation and hash verification

torstentrompler:
The current implementation of the verifyPasswordHash method is costly
when comparing a password against multiple hashes, because the password
hash will be created multiple times.

If the hash creation and hash verification are separated, the hash only
needs to be created once, when comparing a password against multiple
hashes.

372108

### DIFF
--- a/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/security/PasswordHash.java
+++ b/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/security/PasswordHash.java
@@ -1,0 +1,138 @@
+/*
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.scout.rt.platform.security;
+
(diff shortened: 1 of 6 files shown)


org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/security/PasswordHash.java
```diff
@@ -0,0 +1,138 @@
...
+ * Use {@link SecurityUtility#createPasswordHash(char[], byte[])} to create a {@link PasswordHash} that hashes a password
+ * with respect to the current security standard and all known older variations.
+ */
+public final class PasswordHash {
+
+  private final char[] m_password;
+  private final byte[] m_salt;
+  private final List<MutablePair<BiFunction<char[], byte[], byte[]>, byte[]>> m_hashGeneratorAndHash;
+
+  private PasswordHash(char[] password, byte[] salt, List<BiFunction<char[], byte[], byte[]>> hashGeneratorAndHash) {
```
andibur: The last parameter are just hashGenerators, wihtout hashes:

```suggestion
  private PasswordHash(char[] password, byte[] salt, List<BiFunction<char[], byte[], byte[]>> hashGenerators) {
```

andibur (approved):

andibur (approved):
Please make sure to squash the 3 commits. Thanks.

mvilliger (approved):
Looks good, thank you!
Don't forget to manually pick the change to 26.1 and 26.2 release :)

PR #2194, opened 2026-05-13
Use fetchpriority=high for dynamically loaded scripts

bschwarzent:
Modern browsers assign each request a priority. Scripts that are loaded dynamically by injecting an additional <script> tag into the DOM are assigned a low priority by default [1]. Under certain conditions, this will introduce significant but unnecessary delays (> 15s). This is problematic if the caller waits for the script completion.

Since request priorities are primarily relevant during the initial rendering of a HTML page, we can safely increase the priority hint for all scripts that are dynamically injected later using the central method $.injectScript() by setting the "fetchpriority" attribute [2] to "high".

[1] https://web.dev/articles/fetch-priority
[2] https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/fetchpriority

462323

### DIFF
--- a/eclipse-scout-core/src/jquery/jquery-scout.js
+++ b/eclipse-scout-core/src/jquery/jquery-scout.js
@@ -223,6 +223,7 @@ $.injectScript = (url, options) => {
   let $scriptTag = $(scriptTag)
     .attr('src', url)
     .attr('async', true)
+    .attr('fetchpriority', 'high')
     .on('load error', event => {
       if (options.removeTag) {
         myDocument.head.removeChild(scriptTag);


paolobazzi (approved):
looks good to me

cguglielmo (approved):

cguglielmo:
According to your second link, it looks like the async keyword results in a low priority: _Declaring scripts as async or defer tells the browser to load them asynchronously. However, as shown in the [priority table](https://web.dev/articles/fetch-priority#resource-priority), these scripts are also assigned a "Low" priority._

bschwarzent:
Dynamically injected scripts are always loaded asynchronously (same as other resources, e.g. images). Thus, removing the async keyword does not affect the priority. -> Leave it as is.

PR #2195, opened 2026-05-15
Column: sort rows by cell value if sort codes are equal

bschwarzent:
Table cells can have an optional 'sortCode' property that is considered when comparing table rows. If at least one cell has a sortCode, they are compared by sort code first. However, if both sort codes are equal, the normal comparator should be applied to ensures a predictable, consistent row order.

454232

### DIFF
--- a/eclipse-scout-core/src/table/columns/Column.ts
+++ b/eclipse-scout-core/src/table/columns/Column.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2025 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/src/table/columns/Column.ts
```diff
@@ -1029,11 +1029,14 @@ export class Column<TValue = string> extends PropertyEventEmitter implements Col
...
 
   compare(row1: TableRow, row2: TableRow): number {
-    let cell1 = this.table.cell(this, row1),
-      cell2 = this.table.cell(this, row2);
+    let cell1 = this.table.cell(this, row1);
+    let cell2 = this.table.cell(this, row2);
 
     if (cell1.sortCode !== null || cell2.sortCode !== null) {
-      return comparators.NUMERIC.compare(cell1.sortCode, cell2.sortCode);
+      let c = comparators.NUMERIC.compare(cell1.sortCode, cell2.sortCode);
```
fschinkel, line 1036: maybe call this variable `result`

fschinkel (approved):

eclipse-scout-core/src/table/columns/Column.ts
bschwarzent: The name `c` is commonly used for the result of a compare operation.  I think it is clear enough here.

bschwarzent:
Run tests

PR #2202, opened 2026-05-19
Send UI session id also in REST requests if the session is not completely ready yet

rogersteblerbsi:
462325


paolobazzi (approved):
looks good to me!

PR #2205, opened 2026-05-21
Outline: improve lazy expanding

cguglielmo:
- Set lazyExpandingEnabled to true for Scout JS table pages   Scout JS based table pages should expand lazily by default to make it   consistent with Scout Classic table pages.
- Ensure correct expandedLazy state (Scout JS)   If lazyExpandingEnabled is disabled and the node updated,   expandedLazy state should be set to false as well.
- Don't expand table pages on drill down (Scout Classic)   Currently, a double click on a table row in the detail table of a page   always selects and expands the corresponding child page. This is fine   for node pages, but if a table is expanded that has   lazyExpandingEnabled set to true (default), a + is shown for the page   in the outline but no child page is visible. This is confusing and   also inconsistent to what the NavigateDown button does. This button   does the expanding only for node pages (see Outline.ts#drillDown).   Now, the Java based drillDown does the same as the JS based.
- Moved ITreeUIFacade.setNodeSelectedAndExpandedFromUI to IOutline   and renamed it to drillDown, because it is never called from ui.
- Added IPage.computeExpandOnDrillDown which returns true for node   pages.
- Removed unused property isChildrenVolatile/setChildrenVolatile of   ITreeNode.

337244

### DIFF
--- a/eclipse-scout-core/src/desktop/outline/Outline.ts
+++ b/eclipse-scout-core/src/desktop/outline/Outline.ts
@@ -1281,9 +1281,10 @@ export class Outline extends Tree implements DisplayParent, OutlineModel {
   }
 
   /**
-   * Selects the given page.
+   * Selects the given page and, if necessary, scrolls it into the viewport to make it visible.
+   * Also expands the page if it is a node page and `expandDrillNode` is null or undefined.
    *
-   * @param drillNode The page to select. May be null, in which case nothing happens.
+   * @param drillNode The page to select. If it is null, nothing happens.
    * @param expandDrillNode Whether to automatically expand the drill node after selecting it.
    *        By default, this is `true` for node pages, `false` otherwise.
    */
@@ -1295,6 +1296,11 @@ export class Outline extends Tree implements DisplayParent, OutlineModel {
       // If the target node is a node page, expand it
       if (scout.nvl(expandDrillNode, drillNode.nodeType === Page.NodeType.NODES)) {
         this.expandNode(drillNode);
+      }
+
+      // Ensure selected node is visible
+      if (!this.scrollToSelection) {
+        this.revealSelection();
       }
     }
   }
--- a/eclipse-scout-core/src/tree/Tree.ts
+++ b/eclipse-scout-core/src/tree/Tree.ts
@@ -2453,14 +2453,7 @@ export class Tree extends Widget implements TreeModel, Filterable<TreeNode> {
       let propertiesChanged: boolean;
       let oldNode = this.nodesMap[updatedNode.id];
 
-      // if same instance has been updated we must set the flag always to true
-      // because we cannot compare against an "old" node
-      if (updatedNode === oldNode) {
-        propertiesChanged = true;
-      } else {
-        propertiesChanged = this._applyUpdatedNodeProperties(oldNode, updatedNode);
-      }
-
+      propertiesChanged = this._applyUpdatedNodeProperties(oldNode, updatedNode);
       if (propertiesChanged) {
         this.applyFiltersForNode(oldNode);
         this._updateItemPath(false, oldNode.parentNode);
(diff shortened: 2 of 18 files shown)


eclipse-scout-core/src/desktop/outline/Outline.ts
```diff
@@ -1281,9 +1281,10 @@ export class Outline extends Tree implements DisplayParent, OutlineModel {
   }
 
   /**
-   * Selects the given page.
+   * Selects the given page and, if necessary, scrolls it into the viewport to make it visible.
+   * Also expands the page if it is a node page or if `expandDrillNode`is true.
```
bschwarzent: > Also expands the page if it is a node page or if `expandDrillNode`is true.

The first part of this sentence is only true if `expandDrillNode` is undefined. If `expandDrillNode=false`, the page is never expanded, even when it is a node page.

eclipse-scout-core/src/tree/Tree.ts
```diff
@@ -2490,6 +2483,17 @@ export class Tree extends Widget implements TreeModel, Filterable<TreeNode> {
    *          determine if the node has to be rendered again.
    */
   protected _applyUpdatedNodeProperties(oldNode: TreeNode, updatedNode: ObjectOrModel<TreeNode>): boolean {
+    // Make sure expandedLazy is reset to false when lazyExpanding is disabled (same code as in AbstractTreeNode.setLazyExpandingEnabled)
+    if (!updatedNode.lazyExpandingEnabled || !this.lazyExpandingEnabled) {
+      oldNode.expandedLazy = false;
+    }
+
+    // If same instance has been updated we must set the flag always to true
```
bschwarzent: Maybe rephrase? There is no "flag" involved here.

org.eclipse.scout.rt.client.test/src/test/java/org/eclipse/scout/rt/client/ui/desktop/outline/pages/PageTest.java
```diff
@@ -147,7 +147,7 @@ public void testCollapseLazyNodeOnReload() {
 
     // Drill down to child node -> page should be expanded lazily
 
-    outline.getUIFacade().setNodeSelectedAndExpandedFromUI(page.getChildNode(0));
+    outline.drillDown(page.getChildPage(0));
     assertTrue(page.isExpanded());
```
bschwarzent, line 151: This assertion is no longer true - table pages are not expanded automatically.
The test focuses on a different aspect -> maybe just manually expand it here?

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/outline/AbstractOutline.java
```diff
@@ -1094,4 +1094,28 @@ public void changeNode(ITreeNode node, ICell cell) {
...
+  @Override
+  public void drillDown(IPage page) {
+    if (page == null) {
+      return;
+    }
+
+    try {
+      setTreeChanging(true);
+      if (page.computeExpandOnDrillDown()) {
+        if (page.isChildrenDirty()) {
```
bschwarzent, line 1107: Should this be done in setNodeExpandedInternal() instead? That loads children if they are not loaded yet, maybe it could also reload them if they are dirty?

bschwarzent (approved):
Improvements for commit message:

> - Outline.ts#.drillDown now also scrolls to the selected node like

Use `#` or `.` as separator, not both.

> - Don't expand table pages on drill down (Scout Classic)
>   Currently, a double click on a table row in the detail table of a page
>   always selects and expands the corresponding child page. This is fine
>   for node pages, but if a table is expanded that has

"...but if a table page..."

> - Added IPage.computeExpandOnDrillDown which returns true for node
>   pages.

"...which returns false by default and true for node pages."

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/desktop/outline/AbstractOutline.java
cguglielmo: If I moved this to setNodeExpandedInternal(), it would probably create side-effects, for example AbstractPageWithTable.callMinimalInitializer() sets childrenDirty to true and calls setExpandedInternal(). I'd rather not change the existing behavior if not necessary.

PR #2221, opened 2026-06-02
Remove user id from server session

mvilliger:
Fix missing UserId thread local when using the serverbridge.

423389

### DIFF
--- a/org.eclipse.scout.rt.serverbridge/src/main/java/org/eclipse/scout/rt/serverbridge/BridgeToServerBeanDecorator.java
+++ b/org.eclipse.scout.rt.serverbridge/src/main/java/org/eclipse/scout/rt/serverbridge/BridgeToServerBeanDecorator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.serverbridge/src/main/java/org/eclipse/scout/rt/serverbridge/BridgeToServerBeanDecorator.java
```diff
@@ -65,18 +63,14 @@ protected Object ensureRunInServerContext(final IBeanInvocationContext<T> contex
       return continueCall(context);
     }
 
+    IClientSession clientSession = ClientSessionProvider.currentSession();
+
     // bridge to server scope
     ClientNotificationCollector collector = new ClientNotificationCollector();
-    ServerRunContext bridgeRunContext = ServerRunContexts
-        .copyCurrent()
+    ServerRunContext bridgeRunContext = BEANS.get(ServerRunContextProducer.class).produce(clientSession.getSubject(), clientSession.getId())
```
rogersteblerbsi: Is it guaranteed here that the client session always exists? Theoretically the service call could be done in a different run context without a session.

We could also use Subject.current() and null here:

```suggestion
    ServerRunContext bridgeRunContext = BEANS.get(ServerRunContextProducer.class).produce(Subject.current(), clientSession != null ? clientSession.getId() : null)
```

rogersteblerbsi (approved):

org.eclipse.scout.rt.serverbridge/src/main/java/org/eclipse/scout/rt/serverbridge/BridgeToServerBeanDecorator.java
mvilliger: You are right. It could probably be null. I applied your suggestion.

PR #2229, opened 2026-06-09
TabBox.ts: support mark strategy

cguglielmo:
Also update empty property and parent fields similar to the saveNeeded property. For example: GroupBox.empty will return true if all fields are empty and return false if at least one field has a value.

417769


eclipse-scout-core/src/form/fields/FormField.ts
```diff
@@ -1688,3 +1779,11 @@ export interface VisitFieldsOptions {
    */
   visitSelf?: boolean;
 }
+
+/**
+ * Returns true if the field is considered empty, false if not.
```
mvilliger: maybe use @returns here?

eclipse-scout-core/src/form/fields/FormField.ts
```diff
@@ -1529,18 +1546,88 @@ export class FormField extends Widget implements FormFieldModel {
...
+    if (!this.checkEmpty) {
+      return true;
+    }
+
+    let defaultComputeEmpty = this._computeEmpty.bind(this);
+    let empty = objects.isFunction(this.checkEmpty) ? this.checkEmpty(defaultComputeEmpty) : this._computeEmpty();
+    if (empty === null) {
+      // Check whether any direct child field is empty
+      // It is not necessary to check the whole tree because the empty state of the children already reflect the state of their children
+      // Also don't do it during init to prevent unnecessary children traversal
```
mvilliger: This sentence: where in the code is this asserted?

eclipse-scout-core/src/form/fields/tabbox/TabBox.ts
```diff
@@ -48,8 +51,21 @@ export class TabBox extends CompositeField implements TabBoxModel {
...
     this._tabBoxHeaderPropertyChangeHandler = this._onTabBoxHeaderPropertyChange.bind(this);
+    this._tabItemEmptyChangeHandler = this._onTabItemEmptyChange.bind(this);
+    this._tabItemSaveNeededChangeHandler = this._onTabItemSaveNeededChange.bind(this);
   }
 
+  static MarkStrategy = {
+    /**
+     * Sets {@link TabItem.marked} of a tab item to true if it is not {@link empty}.
+     */
+    NOT_EMPTY: 'notEmpty',
```
mvilliger, line 62: we could also call it HAS_VALUE instead of a negation? Would be more readable, on the other side it does not clearly imply that it depends on the 'empy' property. what do you think?

eclipse-scout-core/src/form/fields/tabbox/TabBox.ts
```diff
@@ -228,6 +251,59 @@ export class TabBox extends CompositeField implements TabBoxModel {
...
+      if (this.markStrategy === TabBox.MarkStrategy.SAVE_NEEDED) {
+        tabItem.off('propertyChange:saveNeeded', this._tabItemSaveNeededChangeHandler);
+      } else if (this.markStrategy === TabBox.MarkStrategy.NOT_EMPTY) {
+        tabItem.off('propertyChange:empty', this._tabItemEmptyChangeHandler);
+      }
+    }
+  }
+
+  protected _addMarkStrategyHandlers(tabItems: TabItem[]) {
+    for (const tabItem of tabItems) {
```
mvilliger, line 284: maybe also early return if (!this.markStrategy) like in the _remove method?

mvilliger (approved):

eclipse-scout-core/src/form/fields/FormField.ts
cguglielmo: It was an old comment, removed

eclipse-scout-core/src/form/fields/tabbox/TabBox.ts
cguglielmo: I would not change it because it does not always check the value (e.g. tablefield).

PR #2231, opened 2026-06-10
Increase Http2 Stream Idle Timeout from 30s (jetty default) -> 40s

work-lto:
Update jetty to 12.1.10 and introduce property to configure stream idle timeout.

After update to 12.1.9, we experienced
infrequent ERR_HTTP2_PROTOCOL_ERROR
which might be due to race conditions introduced by inconvenient timeout defaults from both browser and server.

Tests suggest that increasing the stream idle timeout helped eliminate the errors.

453441

### DIFF
--- a/org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ApplicationProperties.java
+++ b/org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ApplicationProperties.java
@@ -670,4 +670,22 @@ public final class ApplicationProperties {
       return "The server stop timeout in milliseconds to allow a proper shutdown. Default value is " + getDefaultValue() + ".";
     }
   }
+
+  public static class ScoutApplicationStreamIdleTimeoutProperty extends AbstractPositiveLongConfigProperty {
+
+    @Override
+    public String getKey() {
+      return "scout.app.jetty.stream.idletimeout";
+    }
+
+    @Override
+    public Long getDefaultValue() {
+      return TimeUnit.SECONDS.toMillis(40);
+    }
+
+    @Override
+    public String description() {
+      return "The HTTP2 stream idle timeout in milliseconds. Default value is " + getDefaultValue() + ".";
+    }
+  }
 }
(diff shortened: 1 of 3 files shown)


org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ApplicationProperties.java
```diff
@@ -670,4 +670,22 @@ public String description() {
...
+  public static class ScoutApplicationStreamIdleTimeoutProperty extends AbstractPositiveLongConfigProperty {
+
+    @Override
+    public String getKey() {
+      return "scout.app.jetty.stream.idletimeout";
+    }
+
+    @Override
+    public Long getDefaultValue() {
+      return TimeUnit.SECONDS.toMillis(40);
```
matthiaso, line 683: Did you choose 40 on purpose/is there any idea behind this number? Maybe we should add a comment why 40 was choosen.

matthiaso (approved):

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ApplicationProperties.java
```diff
@@ -670,4 +670,22 @@ public String description() {
       return "The server stop timeout in milliseconds to allow a proper shutdown. Default value is " + getDefaultValue() + ".";
     }
   }
+
+  public static class ScoutApplicationStreamIdleTimeoutProperty extends AbstractPositiveLongConfigProperty {
+
+    @Override
+    public String getKey() {
+      return "scout.app.jetty.stream.idletimeout";
```
matthiaso:
```suggestion
      return "scout.app.jetty.stream.idleTimeout";   
```

Maybe use upper-case for timeout? At least the property before this one uses uppercase.

org.eclipse.scout.rt.app/src/main/java/org/eclipse/scout/rt/app/ApplicationProperties.java
work-lto: it's a magic number - our best guess for the occuring ERR_HTTP2_PROTOCOL_ERROR errors is that the timeout from the browser (60s) and the streaming timeout (30s) during polling get into a racing condition where probably almost at the same time (60s and 2x30s) requests are send. By increasing the value slightly the error's didn't occur anymore in first tests.

PR #2232, opened 2026-06-11
Add getter to systems.ts to retrieve systems without creating them

tgygax:
443690


eclipse-scout-core/src/system/systems.ts
```diff
@@ -39,6 +39,10 @@ export const systems = {
     return system;
   },
 
+  get(name?: string): System {
```
fschinkel, line 46: Please add js doc.

fschinkel (approved):

fschinkel (approved):

PR #2240, opened 2026-06-22
Lock transitive Node dependencies

mvilliger:
359390


eclipse-scout-releng/bin/releng-scripts.ts
```diff
@@ -0,0 +1,159 @@
...
+    yargs => yargs
+      .option('apikey', {description: 'API Key for authentication', type: 'string'})
+      .option('url', {description: 'URL of the artifactory (including trailing slash).', type: 'string'})
+      .option('user', {description: 'Username', type: 'string'})
+      .option('pwd', {description: 'Password', type: 'string'})
+      .option('reponame', {description: 'Name of the repository', type: 'string'})
+      .option('keep', {description: 'Number of Artifacts to keep', type: 'number'}),
+    cleanupSnapshots
+  )
+  .command('install', 'Install all dependencies using pnpm and update the Scout overrides if required.',
```
cguglielmo: Install -> Installs, update > updates

eclipse-scout-releng/bin/releng-scripts.ts
```diff
@@ -0,0 +1,159 @@
...
+    yargs => yargs
+      .option('updateMode', {
+        description: 'Specifies which dependencies should be updated. ' +
+          `Use "${updateMode.REQUIRED}" to update packages that do not fulfil the specifier in the package.json. ` +
+          `Use "${updateMode.POSSIBLE}" to update all packages for which a newer version exists. ` +
+          `Use "${updateMode.SNAPSHOTS}" to update snapshots only. ` +
+          `Default is "${updateMode.SNAPSHOTS}".`, type: 'string', default: updateMode.SNAPSHOTS
```
cguglielmo: i think REQUIRED should be the default as SNAPSHOTS is only used by the ci build

eclipse-scout-releng/scripts/install/dependency/tree/DependenciesCache.ts
```diff
@@ -0,0 +1,162 @@
...
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+import {type PackageNode} from './PackageNode.ts';
+import {serializeTreeNodeId, type TreeNodeId} from './TreeNodeId.ts';
+
+// Inspired by https://github.com/pnpm/pnpm/blob/v10.26.1/reviewing/dependencies-hierarchy/src/DependenciesCache.ts
```
cguglielmo, line 14: Is this an exact copy or what are the modifications? Maybe mark the modifications so an update is easier or make clear that there are no modifications

eclipse-scout-releng/scripts/install/scoutInstall.ts
```diff
@@ -0,0 +1,104 @@
...
+    '--no-lockfile',
+    '--ignore-scripts',
+    '--config.link-workspace-packages=true',
+    '--config.prefer-workspace-packages=true'
+  ];
+
+  // 'pnpm update' updates all non overridden dependencies to the newest version available that fulfills the required version
+  // if scout overrides are present, 'pnpm update' will only update dependencies that are not part of the scout overrides, i.e. snapshot dependencies
+  // -> disable scout overrides depending on the updateMode
+  if (options.updateMode === updateMode.POSSIBLE) {
```
cguglielmo: I am not sure if I like the name POSSIBLE. Maybe full, all, force?

eclipse-scout-releng/scripts/install/dependency/tree/DependenciesCache.ts
mvilliger: partly exact copy, partly with changes. As newer versions of pnpm completely changed these files, it's not worth marking the modified parts. This would not be helpful in the next update iteration anyway.

eclipse-scout-releng/scripts/install/scoutInstall.ts
mvilliger: We can call it 'all'.

PR #2243, opened 2026-06-25
Busy handling improvements

mvilliger:


eclipse-scout-core/src/lifecycle/Lifecycle.ts
```diff
@@ -122,12 +122,16 @@ export abstract class Lifecycle<TValidationResult extends { errorStatus?: Status
...
       return this._showYesNoCancelMessageBox(
         this.askIfNeedSaveText,
-        this.ok.bind(this),
+        this._okAfterAskIfSaveNeeded.bind(this),
         this.close.bind(this));
     }
     return this.close();
   }
 
+  protected _okAfterAskIfSaveNeeded(): JQuery.Promise<void> {
```
cguglielmo, line 131: Maybe create a SpecLifecycle extending Lifecycle that overrides ok() and triggers and event instead of creating a method only for specs? The lifecycle can easily be replaced on the form like so: 

let form = scout.create(Form, {
  parent: session.desktop,
  lifecycle: SpecLifecycle
});

PR #2246, opened 2026-06-29
Improve extensibility of TreeBox

annebre:
320471

### DIFF
--- a/eclipse-scout-core/src/form/fields/treebox/TreeBox.ts
+++ b/eclipse-scout-core/src/form/fields/treebox/TreeBox.ts
@@ -63,12 +63,16 @@ export class TreeBox<TValue> extends LookupBox<TValue> implements TreeBoxModel<T
       return;
     }
     this._valueSyncing = true;
-    let valueArray = objects.values(this.tree.nodesMap)
-      .filter(node => node.checked)
-      .map((node: TreeBoxTreeNode<TValue>) => node.lookupRow.key);
+    let valueArray = this._getValueFromTree();
 
     this.setValue(valueArray);
     this._valueSyncing = false;
+  }
+
+  protected _getValueFromTree(): TValue[] {
+    return objects.values(this.tree.nodesMap)
+      .filter(node => node.checked)
+      .map((node: TreeBoxTreeNode<TValue>) => node.lookupRow.key);
   }
 
   protected override _valueChanged() {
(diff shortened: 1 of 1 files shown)


eclipse-scout-core/src/form/fields/treebox/TreeBox.ts
```diff
@@ -60,14 +60,18 @@ export class TreeBox<TValue> extends LookupBox<TValue> implements TreeBoxModel<T
...
-    let valueArray = objects.values(this.tree.nodesMap)
-      .filter(node => node.checked)
-      .map((node: TreeBoxTreeNode<TValue>) => node.lookupRow.key);
+    let valueArray = this._getValueFromTree();
 
     this.setValue(valueArray);
     this._valueSyncing = false;
   }
 
+  protected _getValueFromTree() {
```
bschwarzent: Declare return type.

eclipse-scout-core/src/form/fields/treebox/TreeBox.ts
```diff
@@ -147,6 +151,10 @@ export class TreeBox<TValue> extends LookupBox<TValue> implements TreeBoxModel<T
     }, this);
```
bschwarzent: Take the opportunity to change this bound function to an arrow function.

eclipse-scout-core/src/form/fields/treebox/TreeBox.ts
csehlke: missing the return type

eclipse-scout-core/src/form/fields/treebox/TreeBox.ts
```diff
@@ -147,6 +151,10 @@ export class TreeBox<TValue> extends LookupBox<TValue> implements TreeBoxModel<T
...
   }
 
+  protected _isParentKey(rowParentKey: TValue, parentKey: TValue) {
```
csehlke:
```suggestion
  protected _isParentKey(rowParentKey: TValue, parentKey: TValue): boolean {
```

same here

bschwarzent: What exactly is the semantics of this method? It seems to just compare two keys, so maybe remove the word  "parent"? Or maybe the first argument should be a lookup row, the second a key? -> Needs jsdoc, and maybe a better name.

Also: declare return type.

bschwarzent (approved):

PR #2248, opened 2026-06-30
Allow setting CssClass on TableRow without setting it on all Cells

tobias-schlatter:
If you want to apply a CSS class to a TableRow, there are cases where you don't want that CSS class to be inherited by all cells. This adjustment makes that possible.

459443

### DIFF
--- a/org.eclipse.scout.rt.client.test/src/test/java/org/eclipse/scout/rt/client/ui/basic/table/internal/InternalTableRowTest.java
+++ b/org.eclipse.scout.rt.client.test/src/test/java/org/eclipse/scout/rt/client/ui/basic/table/internal/InternalTableRowTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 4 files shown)


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/TableRow.java
```diff
@@ -448,9 +448,16 @@ public void setForegroundColor(String c) {
 
   @Override
   public void setCssClass(String cssClass) {
+    setCssClass(cssClass, true);
+  }
+
+  @Override
+  public void setCssClass(String cssClass, boolean setCellCssClass) {
```
egloff: Could we name the property "propagateCssClassToCells"?

org.eclipse.scout.rt.client.test/src/test/java/org/eclipse/scout/rt/client/ui/basic/table/internal/InternalTableRowTest.java
```diff
@@ -61,6 +61,23 @@ public void testSetCssClass() {
...
+    assertEquals("CssClassOnRow", row.getCssClass());
+    // setting a css class will set css class on cell as well
+    assertEquals("CssClassOnRow", row.getCellForUpdate(0).getCssClass());
+    // set css class on the cell to a different one
+    row.getCellForUpdate(0).setCssClass("CssClassOnCell");
+    assertEquals("CssClassOnCell", row.getCellForUpdate(0).getCssClass());
+    // get internal table row and ensure that the css classes for the row and the cell match
+    InternalTableRow ir = new InternalTableRow(table, row);
+    assertEquals("CssClassOnRow", ir.getCssClass());
+    assertEquals("CssClassOnCell", ir.getCellForUpdate(0).getCssClass());
```
egloff: Could you add a testcase where the new flag ist set to `false`?

egloff (changes requested):

org.eclipse.scout.rt.client.test/src/test/java/org/eclipse/scout/rt/client/ui/basic/table/internal/InternalTableRowTest.java
tobias-schlatter: Good point. I've expanded the test to cover the scenario with the flag set to `false` as well

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/TableRow.java
tobias-schlatter: Sure. I've changed the name.

egloff (approved):

PR #2253, opened 2026-07-03
Table: add support for dragging table rows by mouse

bschwarzent:
354650

### DIFF
--- a/eclipse-scout-core/src/table/Table.ts
+++ b/eclipse-scout-core/src/table/Table.ts
@@ -9,14 +9,14 @@
  */
 import {
   Action, AggregateTableControl, Alignment, AppLinkKeyStroke, aria, arrays, BooleanColumn, Cell, CellEditorPopup, clipboard, Column, ColumnModel, CompactColumn, Comparator, ContextMenuKeyStroke, ContextMenuPopup, dataObjects, Desktop,
-  DesktopPopupOpenEvent, Device, DisplayViewId, DoubleClickSupport, dragAndDrop, DragAndDropHandler, DropType, EnumObject, ErrorHandler, EventHandler, EventModel, events, Filter, Filterable, FilterOrFunction, FilterResult, FilterSupport,
-  FullModelOf, graphics, GridAriaRules, HtmlComponent, IconColumn, InitModelOf, Insets, IUserFilterStateDo, keys, KeyStrokeContext, LimitedResultTableStatus, LoadingSupport, Menu, MenuBar, MenuDestinations, MenuItemsOrder,
-  menus as menuUtil, menus, NumberColumn, NumberColumnAggregationFunction, NumberColumnBackgroundEffect, ObjectOrChildModel, ObjectOrModel, objects, Predicate, PropertyChangeEvent, Range, scout, scrollbars, ScrollToAlignment,
-  ScrollToOptions, Status, StatusOrModel, strings, styles, TabbableCoordinator, TableClientUiPreferenceProfileDo, TableCompactHandler, TableControl, TableCopyKeyStroke, TableCustomizer, TableDefaultRowActionKeyStroke, TableEventMap,
-  TableFooter, TableGroupEvent, TableHeader, TableLayout, TableModel, TableNavigationCollapseKeyStroke, TableNavigationDownKeyStroke, TableNavigationEndKeyStroke, TableNavigationExpandKeyStroke, TableNavigationHomeKeyStroke,
-  TableNavigationPageDownKeyStroke, TableNavigationPageUpKeyStroke, TableNavigationUpKeyStroke, TableOrganizer, TableRefreshKeyStroke, TableRow, TableRowModel, TableSelectAllKeyStroke, TableSelectionHandler, TableSelectKeyStroke,
-  TableStartCellEditKeyStroke, TableTextUserFilter, TableTileGridMediator, TableToggleRowKeyStroke, TableTooltip, TableUiPreferences, tableUiPreferences, TableUpdateBuffer, TableUserFilter, TableUserFilterModel, Tile, TileTableHeaderBox,
-  tooltips, TooltipSupport, TreeGridAriaRules, UiPreferences, UpdateFilteredElementsOptions, UserFilterStateMappers, ValueField, Widget
+  DesktopPopupOpenEvent, Device, DisplayViewId, DoubleClickSupport, dragAndDrop, DragAndDropHandler, DraggableTableRowElement, DropType, EnumObject, ErrorHandler, EventHandler, EventModel, events, Filter, Filterable, FilterOrFunction,
+  FilterResult, FilterSupport, FullModelOf, graphics, GridAriaRules, HtmlComponent, IconColumn, InitModelOf, Insets, IUserFilterStateDo, keys, KeyStrokeContext, LimitedResultTableStatus, LoadingSupport, Menu, MenuBar, MenuDestinations,
+  MenuItemsOrder, menus as menuUtil, menus, NumberColumn, NumberColumnAggregationFunction, NumberColumnBackgroundEffect, ObjectOrChildModel, ObjectOrModel, objects, Predicate, PropertyChangeEvent, Range, scout, scrollbars,
+  ScrollToAlignment, ScrollToOptions, Status, StatusOrModel, strings, styles, TabbableCoordinator, TableClientUiPreferenceProfileDo, TableCompactHandler, TableControl, TableCopyKeyStroke, TableCustomizer, TableDefaultRowActionKeyStroke,
+  TableEventMap, TableFooter, TableGroupEvent, TableHeader, TableLayout, TableModel, TableMoveSupport, TableNavigationCollapseKeyStroke, TableNavigationDownKeyStroke, TableNavigationEndKeyStroke, TableNavigationExpandKeyStroke,
+  TableNavigationHomeKeyStroke, TableNavigationPageDownKeyStroke, TableNavigationPageUpKeyStroke, TableNavigationUpKeyStroke, TableOrganizer, TableRefreshKeyStroke, TableRow, TableRowDropPosition, TableRowModel, TableSelectAllKeyStroke,
+  TableSelectionHandler, TableSelectKeyStroke, TableStartCellEditKeyStroke, TableTextUserFilter, TableTileGridMediator, TableToggleRowKeyStroke, TableTooltip, TableUiPreferences, tableUiPreferences, TableUpdateBuffer, TableUserFilter,
+  TableUserFilterModel, Tile, TileTableHeaderBox, tooltips, TooltipSupport, TreeGridAriaRules, UiPreferences, UpdateFilteredElementsOptions, UserFilterStateMappers, ValueField, Widget
 } from '../index';
 import $ from 'jquery';
 
@@ -136,6 +136,7 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
   defaultRowAction: Action;
   userPreferenceContext: string;
   uiPreferencesEnabled: boolean;
+  rowsDraggable = false;
 
   $data: JQuery;
   $emptyData: JQuery;
@@ -170,6 +171,7 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
   protected _mouseDownRowId: string;
   protected _mouseDownColumn: Column<any>;
   protected _initialUiPreferences: TableClientUiPreferenceProfileDo;
+  protected _rowsDraggableComputed: boolean;
 
   constructor() {
     super();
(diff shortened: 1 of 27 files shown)


eclipse-scout-core/src/table/Table.ts
```diff
@@ -6744,6 +6755,138 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
     }
   }
 
+  setRowsDraggable(rowsDraggable: boolean) {
```
cguglielmo, line 6760: Add /** @see TableModel.rowsDraggable */

eclipse-scout-core/src/table/TableMoveSupport.ts
```diff
@@ -0,0 +1,564 @@
...
+    // Retrieve the allowed drop types for the current target row. By default, the drop types are taken
+    // from the row's custom value "dropTypes". The table also emits an event, so consumers can compute
+    // the drop types dynamically.
+    if (this._moveData.targetRow !== targetRow) {
+      let dropTypes: TableRowDropTypesDo;
+      if (targetRow) {
+        dropTypes = this.table.getAcceptedRowDropTypes(event, this._moveData.sourceRow, targetRow);
+      }
+      this._moveData.targetRowDropTypes = scout.create(TableRowDropTypesDo, {
+        before: dropTypes ? (dropTypes.before || TableRowDropType.ALLOWED) : TableRowDropType.NONE,
```
cguglielmo, line 145: Is it necessary to apply the default drop types again? getAcceptedRowDropTypes already applied them

eclipse-scout-core/src/table/Table.ts
```diff
@@ -6744,6 +6755,138 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+      dropTypes: defaultDropTypes
+    });
+    this.trigger('acceptRowDrop', acceptRowDropEvent);
+    return acceptRowDropEvent.dropTypes;
+  }
+
+  async dropRow(event: JQuery.MouseUpEvent, sourceRow: TableRow, targetRow: TableRow, position: TableRowDropPosition): Promise<void> {
+    let rowDropEvent = new TableRowDropEvent({
+      mouseEvent: event,
+      sourceRow: sourceRow,
```
cguglielmo: maybe use short notation?

eclipse-scout-core/src/table/Table.ts
```diff
@@ -6744,6 +6755,138 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+    }
+  }
+
+  protected _dropRow(sourceRow: TableRow, targetRow: TableRow, position: TableRowDropPosition) {
+    let sourceIndex = this.rows.indexOf(sourceRow);
+    let targetIndex = this.rows.indexOf(targetRow);
+
+    // Check if source row gets a new parent row
+    let parentRow = targetRow.parentRow;
+    if (position === 'first-child' || position === 'last-child') {
```
cguglielmo: i would use TableRowDropPosition instead of strings

eclipse-scout-core/src/table/Table.ts
```diff
@@ -6818,6 +6961,20 @@ export type AggregateTableRow = {
   height?: number;
 };
 
+/**
+ * Interface for rows with custom values "draggable" and "dropTypes" (see JsonTable#tableRowToJson).
+ */
+export type DraggableTableRow = TableRow & {
```
cguglielmo: Maybe better add the properties to the TableRow directly in JavaScript?

eclipse-scout-core/src/table/TableModel.ts
```diff
@@ -308,4 +308,12 @@ export interface TableModel extends WidgetModel {
    * set the `uuid` property explicitly.
    */
   uiPreferencesEnabled?: boolean;
+  /**
+   * Specifies whether rows can be moved by dragging them with the mouse. Only has an effect if the table is enabled.
+   * Default is `false`.
```
cguglielmo: Add an empty line before Default is false

eclipse-scout-core/src/table/TableMoveSupport.ts
```diff
@@ -0,0 +1,564 @@
...
+    // Wait a short moment before making the cursor backdrop visible. That way, the mouse pointer stays unchanged
+    // when the user just selects a row. This pause is also necessary for editable cells. If the backdrop was
+    // shown immediately, it would capture all mouse events, including the 'mouseup' event needed to start the
+    // cell editor (see Table#_onRowMouseUp).
+    this._moveData.$cursorBackdrop.setVisible(false);
+    this._cursorBackdropTimeoutId = setTimeout(() => {
+      this._moveData.$cursorBackdrop.setVisible(true);
+      // Force reflow to update cursor
+      void this._moveData.$cursorBackdrop[0].offsetWidth;
+      // After another short delay, automatically start dragging
```
cguglielmo, line 56: why is this necessary? to show the correct cursor?

eclipse-scout-core/src/table/TableMoveSupport.ts
```diff
@@ -0,0 +1,564 @@
...
+  protected _computeRowText(row: TableRow): string {
+    let summaryColumns = this.table.summaryColumns();
+    if (arrays.hasElements(summaryColumns)) {
+      return strings.join(' ', ...summaryColumns.map(summaryColumn => summaryColumn.cellText(row)));
+    }
+    for (let column of this.table.visibleColumns()) {
+      if (!column.nodeColumnCandidate) {
+        continue; // ignore special gui-only columns
+      }
+      // Explicitly cast text to String, because callText() might return other types (see Column#_formatValue)-
```
cguglielmo, line 439: when does cellText not return a string? This sounds like a bug

eclipse-scout-core/src/table/TableMoveSupport.ts
```diff
@@ -0,0 +1,564 @@
...
+
+    // Compute the potential target position from the current mouse position. This position is later validated
+    // and matched with the allowed drop types. The result is called "placement" and describes the position
+    // of the drop indicator. When the mouse cursor is released, the current placement is converted into
+    // a more convenient enum value called "drop position".
+    if (rowBelowCursor) {
+      targetRow = rowBelowCursor;
+
+      let rowBounds = graphics.offsetBounds($rowBelowCursor);
+      let rowMiddleY = rowBounds.y + rowBounds.height * 0.5;
```
cguglielmo: or rowBounds.center().y

eclipse-scout-core/src/table/TableMoveSupport.ts
```diff
@@ -0,0 +1,564 @@
...
+ * @see "TableRowDropType.java"
+ */
+export enum TableRowDropType {
+  ALLOWED = 'allowed',
+  FORBIDDEN = 'forbidden',
+  NONE = 'none'
+}
+
+/**
+ * @see "TableRowDropPosition.java"
```
cguglielmo: I would not use JsDoc to link to a Java element as it is not relevant for the consumer of the api. If you think linking is necessary, I would just use a regular comment

eclipse-scout-core/src/table/Table.ts
```diff
@@ -6744,6 +6755,138 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+    this._updateRowsDraggableComputed();
+  }
+
+  override recomputeEnabled(parentEnabled?: boolean) {
+    super.recomputeEnabled(parentEnabled);
+    this._updateRowsDraggableComputed();
+  }
+
+  isRowDraggable(row: DraggableTableRow): boolean {
+    return this._rowsDraggableComputed && scout.nvl(row.draggable, true);
```
cguglielmo: I think a disabled row should not be draggable by default, what do you think? The question is how can it be adjusted if it SHOULD be draggable in some cases. Do we need an acceptRowDrag event?

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/AbstractTable.java
```diff
@@ -4422,6 +4445,21 @@ private void fireRowAction(ITableRow row) {
     }
   }
 
+  private void fireRowDrop(ITableRow sourceRow, ITableRow targetRow, TableRowDropPosition position) {
+    try {
+      // FIXME bsh [dnd-table]: Add interceptRowDrop()
```
cguglielmo: I would say yes

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/AbstractTable.java
```diff
@@ -4422,6 +4445,21 @@ private void fireRowAction(ITableRow row) {
...
+      // FIXME bsh [dnd-table]: fireRowDrop vs. fireRowDropAction()?
```
cguglielmo, line 4451: We need to somehow clean this up. 
1. Rename fireRowDropAction to fireDrop including events
2. Or rename execRowDrop to execMoveRowsDrop
3. Or use the existing execDrop and don't create a new method. TransferObject would then contain the source row (or rows) and the position
4. Or use the existing execDrop and delegate to a new execRowsDrop with a simpler api

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/AbstractTable.java
```diff
@@ -4422,6 +4445,21 @@ private void fireRowAction(ITableRow row) {
...
+      execRowDrop(sourceRow, targetRow, position);
+    }
+    catch (Exception ex) {
+      BEANS.get(ExceptionHandler.class).handle(ex);
+    }
+  }
+
+  protected void execRowDrop(ITableRow sourceRow, ITableRow targetRow, TableRowDropPosition position) {
+    // FIXME bsh [dnd-table]: Should there be a default implementation, analog to Table#_rowDrop in JS?
```
cguglielmo: Probably, but we can implement it later

eclipse-scout-core/src/table/Table.ts
```diff
@@ -6744,6 +6755,138 @@ export class Table extends Widget implements TableModel, Filterable<TableRow> {
...
+  }
+
+  async dropRow(event: JQuery.MouseUpEvent, sourceRow: TableRow, targetRow: TableRow, position: TableRowDropPosition): Promise<void> {
+    let rowDropEvent = new TableRowDropEvent({
+      mouseEvent: event,
+      sourceRow: sourceRow,
+      targetRow: targetRow,
+      position: position
+    });
+    this.trigger('rowDrop', rowDropEvent);
```
cguglielmo: Should we prepare the api for multi row drop? I think it is something we would like to have.
-> rowsDrop, dropRows, sourceRows: TableRow[], 
Due to lack of time we can also add it later and write in the release notes that it is an experimental api and will likely be changed

eclipse-scout-core/src/table/TableAdapter.ts
```diff
@@ -367,6 +367,28 @@ export class TableAdapter extends ModelAdapter {
     event.preventDefault();
   }
 
+  protected _onWidgetRowDrop(event: TableRowDropEvent) {
+    this._sendRowDrop(event.sourceRow, event.targetRow, event.position);
+    event.promise = new Promise((resolve, reject) => {
```
cguglielmo: Why do we need a promise at all? Can't we just call event.preventDefault()? Why do we have to wait for the server response?

eclipse-scout-core/src/table/TableMoveSupport.ts
bschwarzent: The event listener could set the values to null, therefore the defaults have to be applied again.

eclipse-scout-core/src/table/Table.ts
bschwarzent: 2 of 4 properties cannot use short notation, therefore I prefer to use long notation for all properties.

eclipse-scout-core/src/table/TableMoveSupport.ts
bschwarzent: It transitions between "grab" and "grabbing", giving the user a nice indication what is happening.

eclipse-scout-core/src/table/Table.ts
bschwarzent: Changed to `scout.nvl(row.draggable, row.enabled)` -> row can explicitly set "draggable" if desired

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/ui/basic/table/AbstractTable.java
bschwarzent: The AbstractTable API currently does not support changing the parent row without reloading the entire table (or only very complicated). Therefore, the execRowDrop() method stays empty for the moment.

PR #2257, opened 2026-07-13
Add dataObjectVisitors

fschinkel:
Data objects can now be visited in Scout JS using dataObjectVisitors. The visitor will traverse sets, arrays, maps, objects and BaseDoEntities and will be executed on all of these elements and their children.

418622

### DIFF
--- a/eclipse-scout-core/src/dataobject/dataObjectVisitors.ts
+++ b/eclipse-scout-core/src/dataobject/dataObjectVisitors.ts
@@ -1,0 +1,225 @@
+/*
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+import {arrays, BaseDoEntity, Constructor, objects, scout, strings, TreeVisitResult} from '../index';
+
+export const dataObjectVisitors = {
+
+  /**
+   * Visits all nodes and calls on each element matching the given type the provided consumer.
+   * If a node matches, child nodes of this node are not visited.
+   */
+  forEach<T>(root: any, type: Constructor<T>, consumer: (node: T) => void) {
+    dataObjectVisitors.forEachIf(root, (type ? c => c instanceof type : c => true) as (candidate) => candidate is T, consumer);
+  },
+
+  /**
+   * Visits all nodes and calls on each element matching the given type the provided consumer.
+   * If a node matches, child nodes of this node are also visited.
+   */
+  forEachRec<T>(root: any, type: Constructor<T>, consumer: (node: T) => void) {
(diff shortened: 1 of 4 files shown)


eclipse-scout-core/src/dataobject/dataObjectVisitors.ts
```diff
@@ -0,0 +1,164 @@
...
+import {BaseDoEntity, objects, scout, strings, TreeVisitor, TreeVisitResult} from '../index';
+
+export const dataObjectVisitors = {
+
+  /**
+   * Visits all nodes and calls on each element matching the given type predicate the provided consumer.
+   * If a node matches, child nodes of this node are not visited.
+   */
+  forEach<T>(root: any, typePredicate: (candidate) => candidate is T, consumer: (node: T) => void) {
+    scout.create(DataObjectVisitor).visit(root, node => {
```
mvilliger: Could this function use forEachRecIf internally instead of copying?

eclipse-scout-core/src/dataobject/dataObjectVisitors.ts
```diff
@@ -0,0 +1,164 @@
...
+      return TreeVisitResult.CONTINUE;
+    });
+  },
+
+  /**
+   * Visits all nodes and calls on each element matching the given type predicate the provided consumer.
+   * If a node matches, child nodes of this node are also visited.
+   */
+  forEachRec<T>(root: any, typePredicate: (candidate) => candidate is T, consumer: (node: T) => void) {
+    scout.create(DataObjectVisitor).visit(root, node => {
```
mvilliger: Could this function use forEachRecIf internally instead of copying?

eclipse-scout-core/src/dataobject/dataObjectVisitors.ts
```diff
@@ -0,0 +1,164 @@
...
+  forEach<T>(root: any, typePredicate: (candidate) => candidate is T, consumer: (node: T) => void) {
+    scout.create(DataObjectVisitor).visit(root, node => {
+      if (typePredicate(node)) {
+        consumer(node);
+        return TreeVisitResult.SKIP_SUBTREE;
+      }
+      return TreeVisitResult.CONTINUE;
+    });
+  },
+
```
mvilliger, line 49: maybe add convenience functions for a Type-Reference (constructor) as typePredicate?
So I can write `forEach(node, MyDo, n => {})` instead of `forEach(node, c => c instanceof MyDo, n => {})`?

eclipse-scout-core/src/dataobject/dataObjectVisitors.ts
```diff
@@ -0,0 +1,164 @@
...
+  /**
```
mvilliger, line 50: Maybe also add a visit() function to BaseDoEntity?

eclipse-scout-core/src/dataobject/dataObjectVisitors.ts
fschinkel: I would like to, but I do not see how.
At runtime a constructor is simply a function so I can not distinguish whether the caller passed a constructor or a predicate. The only way is a whole new set of functions.

PR #2258, opened 2026-07-13
scoutInstall: enable scout overrides again after pnpm install/update

fschinkel:
If scoutInstall fails e.g. because of and error while updating the overrides the scout overrides must not stay disabled because otherwise re-running scoutInstall will result in a full update of all dependencies. Therefore, re-enable the scout overrides again after the pnpm calls they have been disabled for.

359390

Assisted-by: Sonnet 4.6, Haiku 4.5 (Claude Code)


eclipse-scout-releng/scripts/util/PnpmWorkspaceYaml.ts
mvilliger: create a constant for this and reuse in addScoutOverrides

mvilliger (approved):

mvilliger (approved):

PR #2260, opened 2026-07-17
Improve batched Column performance

mvilliger:
SmartColumns with batched LookupCall already collected all keys and then updated the cell text as soon as the async LookupCall completed. Therefore, there is only one promise for the LookupCall execution returning the texts for all keys.
Unfortunately the SmartColumn then splits this promise into one promise for each cell containing only the value for the specific cell. This leads to thousands of promises on large tables.
These promises are collected in the TableUpdateBuffer and removed as they resolve. This operation was O(n^2) as the buffer internally used an array to store the promises.

Processing lots of such promises together with the inefficient handling of the buffer leads to unnecessary computation time on the client. Because of some async overhead, executing a promise for each cell is significantly slower than executing one promise handling all cells. On large tables with thousands of rows and multiple SmartColumns this may sum up to several seconds just to process these promises.

This commit changes the following:
- Introduce a helper class BatchCall to collect keys and using an async batch callback executed for all the keys.
- Column uses this new BatchCall in formatValue() if available. This allows each column to support batch processing more easily. Implementors only need to set the _batchFormat member defining the batch callback.
- The deferred Cell text update in batch mode always updates all texts of the column instead of only one cell. Accordingly, _formatValue and formatValue() in the async case now return all texts instead of only the one for a single cell.
- StaticLookupCall & LookupCallColumn now supports batching to benefit from the new features.

410165, 399939


eclipse-scout-core/src/util/BatchCall.ts
```diff
@@ -0,0 +1,135 @@
...
+        existing = [];
+        keyMap.set(key, existing);
+      }
+      existing.push(keys[index]);
+    });
+    return keyMap;
+  }
+
+  protected _serializeKey(key: TKey): string {
+    return dataObjects.stringify(key);
```
fschinkel, line 124: Maybe deserialize the key first. Consider a `SmartColumn<SomeDo>` using the `RestLookupCall` to lookup its values. Depending on whether `SomeDo` is a `BaseDoEntity` and/or how the data of the table is loaded the value of the cell might be a `BaseDoEntity`. But as the resulting key of the `RestLookupCall` will always be a pojo simply using `dataObjects.stringify` will result in different values if the pojo has e.g. a `_typeVersion` set.

fschinkel (dismissed):

eclipse-scout-core/src/table/columns/LookupColumn.ts
```diff
@@ -23,13 +25,38 @@ export class LookupColumn<TValue> extends LookupCallColumn<TValue[], TValue> imp
     this.setProperty('distinct', distinct);
   }
 
+  protected override _init(model: InitModelOf<this>) {
+    super._init(model);
+    this._batchFormat = new BatchCall<TValue[], string>(this._batchFormatValues.bind(this));
+  }
+
+  protected _batchFormatValues(keys: TValue[][]): JQuery.Promise<Map<TValue[], string>> {
+    const allKeys = [...new Set(keys.flat())];
```
fschinkel, line 34: Does this work for complex objects?

eclipse-scout-core/src/table/columns/Column.ts
```diff
@@ -147,7 +148,7 @@ export class Column<TValue = string> extends PropertyEventEmitter implements Col
 
     this._tableColumnsChangedHandler = this._onTableColumnsChanged.bind(this);
     this._realWidth = null;
-
+    this._batchFormat = null;
```
bschwarzent, line 151: Initialization to null not necessary for this internal property (or can be merged with the property declaration).

eclipse-scout-core/src/table/columns/Column.ts
```diff
@@ -398,12 +406,10 @@ export class Column<TValue = string> extends PropertyEventEmitter implements Col
       // If every cell of a row is empty the row would collapse, using nbsp makes sure the row is as height as the others even if it is empty
       content = '&nbsp;';
       cssClass = strings.join(' ', cssClass, 'empty');
+    } else if (cell.flowsLeft) {
```
bschwarzent, line 409: Event shorter:
```
} else {
  content = cell.flowsLeft ? text + icon : icon + text;
}
```

eclipse-scout-core/src/table/columns/Column.ts
```diff
@@ -781,26 +787,47 @@ export class Column<TValue = string> extends PropertyEventEmitter implements Col
...
-  setCellTextDeferred(promise: JQuery.Promise<string>, row: TableRow, cell: Cell<TValue>) {
-    promise
-      .done(text => this.setCellText(row, text, cell))
-      .fail(error => {
-        this.setCellText(row, '', cell);
-        $.log.error('Could not resolve cell text for value ' + cell.value, error);
-      });
-
+  setCellTextDeferred(promise: JQuery.Promise<BatchCallResult<TValue, unknown>>) {
     // (then) promises always resolve asynchronously which means the text will always be set later after row is initialized and will generate an update row event.
```
bschwarzent, line 791: What does `(then)` at the start of this sentence mean?

eclipse-scout-core/src/table/columns/Column.ts
```diff
@@ -345,15 +346,22 @@ export class Column<TValue = string> extends PropertyEventEmitter implements Col
...
-  protected _formatValue(value: TValue, row?: TableRow): string | JQuery.Promise<string> {
+  protected _formatValue(value: TValue, row?: TableRow): string | JQuery.Promise<BatchCallResult<TValue, unknown>> {
+    if (objects.isNullOrUndefined(value)) {
+      return '';
+    }
+    if (this._batchFormat) {
+      this._batchFormat.addKey(value);
+      return this._batchFormat.promise();
+    }
     return scout.nvl(value, '');
```
bschwarzent, line 365: If value is not a string, the result will also not be a string, which contradicts the declared return type of this method. Maybe just convert it to String, same as in _batchFormatResultToCellText()?

eclipse-scout-core/src/table/columns/Column.ts
```diff
@@ -781,26 +787,47 @@ export class Column<TValue = string> extends PropertyEventEmitter implements Col
...
+  setCellTextDeferred(promise: JQuery.Promise<BatchCallResult<TValue, unknown>>) {
     // (then) promises always resolve asynchronously which means the text will always be set later after row is initialized and will generate an update row event.
     // To make sure not every cell update will render the viewport (which is an expensive operation), the update is buffered and done as soon as all promises resolve.
-    this.table.updateBuffer.pushPromise(promise);
+    const added = this.table.updateBuffer.pushPromise(promise);
+    if (!added) {
+      return;
+    }
+
+    promise
```
bschwarzent: Since we are not returning the result, it should be safe to use then/catch instead of done/fail.

eclipse-scout-core/src/table/columns/Column.ts
```diff
@@ -781,26 +787,47 @@ export class Column<TValue = string> extends PropertyEventEmitter implements Col
...
+      .done(result => this._onSetCellTextDeferredDone(result))
+      .fail(error => this._onSetCellTextDeferredFail(error));
+  }
+
+  protected _onSetCellTextDeferredDone(result: BatchCallResult<TValue, unknown>) {
+    for (const row of this.parent.rows) {
```
bschwarzent, line 805: Use `this.table` instead of `this.parent` (same value, but better readable).
Also in _onSetCellTextDeferredFail().

eclipse-scout-core/src/table/columns/Column.ts
```diff
@@ -781,26 +787,47 @@ export class Column<TValue = string> extends PropertyEventEmitter implements Col
...
-        this.setCellText(row, '', cell);
-        $.log.error('Could not resolve cell text for value ' + cell.value, error);
-      });
-
+  setCellTextDeferred(promise: JQuery.Promise<BatchCallResult<TValue, unknown>>) {
     // (then) promises always resolve asynchronously which means the text will always be set later after row is initialized and will generate an update row event.
     // To make sure not every cell update will render the viewport (which is an expensive operation), the update is buffered and done as soon as all promises resolve.
-    this.table.updateBuffer.pushPromise(promise);
+    const added = this.table.updateBuffer.pushPromise(promise);
+    if (!added) {
```
bschwarzent: The comment does not really explain why the update buffer would reject the promise, and neither does the pushPromise() method itself. I think this is an important aspect of the entire feature, therefore it should be explained.

eclipse-scout-core/src/table/columns/Column.ts
```diff
@@ -781,26 +787,47 @@ export class Column<TValue = string> extends PropertyEventEmitter implements Col
...
+    }
+
+    promise
+      .done(result => this._onSetCellTextDeferredDone(result))
+      .fail(error => this._onSetCellTextDeferredFail(error));
+  }
+
+  protected _onSetCellTextDeferredDone(result: BatchCallResult<TValue, unknown>) {
+    for (const row of this.parent.rows) {
+      const cell = this.parent.cell(this, row);
```
bschwarzent, line 806: Use `this.cell(row)` instead.

eclipse-scout-core/src/table/columns/Column.ts
```diff
@@ -781,26 +787,47 @@ export class Column<TValue = string> extends PropertyEventEmitter implements Col
...
+        // only update cells for which a mapping exists
+        const cellText = this._batchFormatResultToCellText(result.get(cell.value), row, cell);
+        this.setCellText(row, cellText, cell);
+      }
+    }
+  }
+
+  protected _onSetCellTextDeferredFail(error: any) {
+    for (const row of this.parent.rows) {
+      const cell = this.parent.cell(this, row);
```
bschwarzent, line 817: Could be shortened to:
```
this.setCellText(row, '');
```

eclipse-scout-core/src/table/columns/LookupCallColumn.ts
```diff
@@ -40,13 +40,21 @@ export class LookupCallColumn<TValue, TKey = TValue> extends Column<TValue> impl
...
-      this.lookupCall = scout.create(CodeLookupCall<TKey>, {
+      this._setLookupCall(scout.create(CodeLookupCall<TKey>, {
         session: this.session,
         codeType
-      });
+      }));
     }
   }
 
+  protected _prepareAndExecuteLookupCallCloneAsync(lookupCall: LookupCall<any>, row?: TableRow): string {
```
bschwarzent, line 50: Why does this method name contain the word "clone"? Neither does the implementation clone anything nor is it relevant for its execution.
Also, ist the word "async" necessary? The method itself is not async, and executing a lookup call is always async by definition.

eclipse-scout-core/src/table/columns/LookupColumn.ts
bschwarzent: Should be ok, because key-identity is handled by BatchCall. Correct?

eclipse-scout-core/src/table/columns/SmartColumn.ts
```diff
@@ -65,6 +67,17 @@ export class SmartColumn<TValue> extends LookupCallColumn<TValue> implements Sma
     }
   }
 
+  protected _batchFormatValues(keys: TValue[]): JQuery.Promise<Map<TValue, string>> {
+    const lookupCall = this.lookupCall.cloneForKeys(keys);
+    this.trigger('prepareLookupCall', {lookupCall});
```
bschwarzent, line 72: The previous implementation in _formatValue() triggered this event including the `row` property. Because we are now in a batch context, this can no longer be set. However, listeners still may rely on the property. Isn't this a problem?

eclipse-scout-core/src/util/BatchCall.ts
```diff
@@ -0,0 +1,125 @@
+/*
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {scout} from '../scout';
```
bschwarzent, line 10: Import from index, not from individual files.

eclipse-scout-core/src/util/BatchCall.ts
```diff
@@ -0,0 +1,125 @@
...
+      originalKeys?.forEach(originalKey => result.set(originalKey, value));
+    });
+    return result;
+  }
+
+  protected _doBatchCall(keys: TKey[]): JQuery.Promise<Map<TKey, TValue>> {
+    if (arrays.empty(keys)) {
+      return $.resolvedPromise();
+    }
+
```
bschwarzent, line 70:
```suggestion
```

eclipse-scout-core/src/util/BatchCall.ts
```diff
@@ -0,0 +1,125 @@
...
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {scout} from '../scout';
+import {arrays} from './arrays';
+import {dataObjects} from '../dataobject/dataObjects';
+import $ from 'jquery';
+
+export class BatchCall<TKey, TValue> {
```
bschwarzent, line 15: Add some documentation? (see old `BatchLookup` implementation).

eclipse-scout-core/src/util/BatchCall.ts
```diff
@@ -0,0 +1,125 @@
...
+    this._batchCall = scout.assertValue(batchCall, 'Missing batch call handler.');
+  }
+
+  addKey(key: TKey) {
+    this._ensureBatchCallReady();
+    this._keySet.add(key);
+  }
+
+  /**
+   * Same key might point to the same values! Do not modify (read only!)
```
bschwarzent, line 35: What is the purpose of this documentation? It does not seem to be related to this method. Modifying the content of keys is never a good idea, is an explicit warning really needed?

eclipse-scout-core/src/table/columns/LookupColumn.ts
```diff
@@ -23,13 +25,38 @@ export class LookupColumn<TValue> extends LookupCallColumn<TValue[], TValue> imp
...
+    return LookupEditor.formatValues(allKeys, this.lookupCall, lookupCall => this.trigger('prepareLookupCall', {lookupCall})).then(rows => {
+      const rowsByKey = new Map<TValue, LookupRow<TValue>>(rows.map(r => [r.key, r]));
```
cguglielmo: Generics for Map are not necessary, can be inferred by TS

eclipse-scout-core/src/table/columns/SmartColumn.ts
mvilliger: This will be described in the migration notes and all current usages have been migrated.

eclipse-scout-core/src/table/columns/LookupColumn.ts
mvilliger: Exactly

eclipse-scout-core/src/util/BatchCall.ts
mvilliger: This should no longer be necessary in the near future as the RestLookupCall is migrated to BaseDoEntity as well.

eclipse-scout-core/src/table/columns/Column.ts
mvilliger: Yes, JsDoc is still missing in this version and will be added.

eclipse-scout-core/src/util/BatchCall.ts
mvilliger: JsDoc is still missing in this version and will be added

PR #2261, opened 2026-07-21
Form: don't scroll to a focused widget when rendering

cguglielmo:
The form already tried to prevent scrolling by resetting scroll positions to 0. This does not work anymore since 26.1 if the widget to be focused is a very large table, because the table data now has the focus instead of the table, which means the scrollParent() returns the table data instead of the group box body. So scroll positions were reset for the wrong object.
Instead of only considering the actual parents (scrollParent() returns the element itself if it is scrollable, which is not expected here), we use the preventScroll option because it is now widely supported by the browsers.

453283

### DIFF
--- a/eclipse-scout-core/src/focus/FocusManager.ts
+++ b/eclipse-scout-core/src/focus/FocusManager.ts
@@ -291,22 +291,23 @@ export class FocusManager implements FocusManagerOptions {
     }
   }
 
-  requestFocusIfReady(element: HTMLElement | JQuery, filter?: () => boolean): boolean {
-    return this.requestFocus(element, filter, {onlyIfReady: true});
+  /**
+   * @see requestFocus
+   * @see RequestFocusOptions.onlyIfReady
+   */
+  requestFocusIfReady(element: HTMLElement | JQuery, options?: FocusContextFocusOptions): boolean {
+    options = options || {};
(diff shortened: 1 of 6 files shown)


eclipse-scout-core/src/focus/FocusManager.ts
```diff
@@ -291,8 +291,13 @@ export class FocusManager implements FocusManagerOptions {
     }
   }
 
-  requestFocusIfReady(element: HTMLElement | JQuery, filter?: () => boolean): boolean {
-    return this.requestFocus(element, filter, {onlyIfReady: true});
+  /**
+   * @see requestFocus
```
bschwarzent, line 295: It's ok to use the same method signature as requestFocus().

But is requestFocus() really correct?

- The JsDoc of the "element" argument is incorrect. Passing null will **not** use the context's first focusable element. Instead, the method returns with false. The docs seems to have been copied from FocusContext.validateAndSetFocus().
- What exactly is the purpose of the "filter" argument? If an explicit element is passed, what additional filtering is there to be done? Also, I can't see any usage of this argument, not even in tests. This (presumably unnecessary) arguments makes the method cumbersome to use if the caller wants to set focus options.

cguglielmo: nothing new, but good input, thanks

bschwarzent (approved):

PR #2262, opened 2026-07-21
Improve permission code calculation by hashing the ID signature

rogersteblerbsi:
Uses IdCodec to generate a signature for the permission's ID and hashes that signature, providing stronger security than hashing the raw ID

469416


paolobazzi (approved):
looks good

PR #2263, opened 2026-07-22
Client notification registry: Separate queue creation/lastAccess time

paolobazzi:
Offer separate methods to get queue creation timestamp and queue
last consume access. The former implementation implied that the
queue's creation timestamp was also the initial last consume access
timestamp, which was wrong as long as a queue was never consumed.

470449


rogersteblerbsi (approved):

paolobazzi:
Already merged.

PR #2267, opened 2026-07-24
Add UI node id header for requests from UI

rogersteblerbsi:
Additionally, for requests proxied to the backend, the corresponding client node ID gets sent to the backend via HTTP header

379091

### DIFF
--- a/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/rest/NodeIdClientRequestFilter.java
+++ b/org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/rest/NodeIdClientRequestFilter.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.scout.rt.client.rest;
+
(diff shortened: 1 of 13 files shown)


org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/rest/NodeIdClientRequestFilter.java
```diff
@@ -0,0 +1,32 @@
...
+@ApplicationScoped
+public class NodeIdClientRequestFilter implements ClientRequestFilter {
+
+  @Override
+  public void filter(ClientRequestContext requestContext) throws IOException {
+    addNodeIdHeader(requestContext);
+  }
+
+  protected void addNodeIdHeader(ClientRequestContext requestContext) {
+    requestContext.getHeaders().add(NodeId.HTTP_HEADER_NAME, IIds.toString(NodeId.current()));
```
matthiaso: I presume we do not want a null-check here on purpose as we assume `NodeId.current()` never-ever returns `null` (which should be true according to `org.eclipse.scout.rt.platform.context.NodeIdentifier#compute`).

Otherwise if `null` could be returned it would be possible that we send a `String` containing `"null"`.

matthiaso (approved):

org.eclipse.scout.rt.client/src/main/java/org/eclipse/scout/rt/client/rest/NodeIdClientRequestFilter.java
rogersteblerbsi: Added a null check just to be sure, in case someone replaced the way how the nodeId is computed.

PR #2268, opened 2026-07-24
Add eml mime magic bytes for more message types

dwl-bsi:
Added bytes with the following headers: Forward to, #! rnews, N#! rnews,
 Pipe to, Relay-Version, Return-Path, Subject, Date, Message-ID

461844

### DIFF
--- a/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
+++ b/org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
@@ -47,7 +47,22 @@ public interface IMimeMagic {
   IMimeMagic RAR = createMagic("526172211A0700", "526172211A070100");
   IMimeMagic CLASS = createMagic("CAFEBABE");
   IMimeMagic OGG = createMagic("4F676753");
-  IMimeMagic EML = createMagic("52656365697665643A", "46726F6D3A", "582D556E73656E743A");
+  IMimeMagic EML = createMagic(
+      // magic bytes of EML files are simply ascii bytes of the first header
+      "52656365697665643A", // Received:
+      "46726F6D3A", // From:
+      "582D556E73656E743A", // X-Unsent:
+      "466F727761726420746F", // Forward to
+      "232120726E657773", // #! rnews
+      "4E232120726E657773", // N#! rnews
+      "5069706520746F", // Pipe to
+      "52656C61792D56657273696F6E3A", // Relay-Version:
+      "52657475726E2D506174683A", // Return-Path:
+      "52657475726E2D706174683A", // Return-path:
+      "5375626A6563743A20", // Subject:
+      "446174653A", // Date:
+      "4D6573736167652D49443A" // Message-ID:
+  );
   IMimeMagic FLV = createMagic("464C56");
   IMimeMagic MPG = createMagic("47", "000001BA", "000001B3");
   IMimeMagic FLAC = createMagic("664C6143");


org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
```diff
@@ -47,7 +47,22 @@ public interface IMimeMagic {
...
+      // magic bytes of EML files are simply ascii bytes of the first header
+      "52656365697665643A", // Received:
+      "46726F6D3A", // From:
+      "582D556E73656E743A", // X-Unsent:
+      "466F727761726420746F", // Forward to
+      "232120726E657773", // #! rnews
+      "4E232120726E657773", // N#! rnews
+      "5069706520746F", // Pipe to
+      "52656C61792D56657273696F6E3A", // Relay-Version:
+      "52657475726E2D506174683A", // Return-Path:
```
Knicksaas, line 60: Puremagic additionally uses here another magic `52657475726e2d50` -> `Return-P`

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
```diff
@@ -47,7 +47,22 @@ public interface IMimeMagic {
   IMimeMagic RAR = createMagic("526172211A0700", "526172211A070100");
   IMimeMagic CLASS = createMagic("CAFEBABE");
   IMimeMagic OGG = createMagic("4F676753");
-  IMimeMagic EML = createMagic("52656365697665643A", "46726F6D3A", "582D556E73656E743A");
+  IMimeMagic EML = createMagic(
+      // magic bytes of EML files are simply ascii bytes of the first header
+      "52656365697665643A", // Received:
+      "46726F6D3A", // From:
```
Knicksaas, line 53: Puremagic uses here another magic `46726f6d` -> `From` (without colon)

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
```diff
@@ -47,7 +47,22 @@ public interface IMimeMagic {
...
+      "232120726E657773", // #! rnews
+      "4E232120726E657773", // N#! rnews
+      "5069706520746F", // Pipe to
+      "52656C61792D56657273696F6E3A", // Relay-Version:
+      "52657475726E2D506174683A", // Return-Path:
+      "52657475726E2D706174683A", // Return-path:
+      "5375626A6563743A20", // Subject:
+      "446174653A", // Date:
+      "4D6573736167652D49443A" // Message-ID:
+  );
```
Knicksaas, line 65: Puremagic classifies the magic `582d` -> `X-` as "Exchange E-Mail". This is probably a little bit too generic. What do you think?

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
dwl-bsi: I'm not sure why they use this prefix. As far as I can tell, only "Return-Path:" is allowed.
Maybe someone added the prefix "just in case"? I would not add it to our list for now

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
dwl-bsi: The e-mail "From:" header should always contain the colon. We used these magic bytes successfully before, too. I would not change them for now

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
dwl-bsi: That seems very strange indeed. I would not use that.
I found another open source project for file detection via metadata which agrees with the magic bytes defined in this PR: https://raw.githubusercontent.com/apache/tika/master/tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml (look for message/x-emlx)

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
Knicksaas: I agree

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
Knicksaas: Ok.

org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/resource/IMimeMagic.java
Knicksaas: That seems to be a better source than Puremagic. I agree—it's probably best if we leave it at that.

Knicksaas (approved):

PR #2272, opened 2026-07-27
Table: update tree structure when adding/updating/removing rows

bschwarzent:
When adding, updating or removing rows in a table, its "hierarchical" property might change. Unless the table was non-hierarchical before the change and is non-hierarchical after the change, the internal data structures managing the row hierarchy have to be fully re-created.

Additionally, if the table switches from hierarchical to non-hierarchical mode or vice versa, all rows in the viewport have to be re-rendered to adjust for the changed indentation of the tableNodeColumn. For example, if there are two rows A and B where B is a child row of A and row B is deleted, row A still has to be re-rendered, despite not being directly altered.

471043


bschwarzent:
Run tests

cguglielmo (approved):

PR #2274, opened 2026-07-28
Fix UiNotifications in cluster

fschinkel:
The UiNotification holds all notifications known on a cluster node and each notification has information from which node it originated. When a listener subscribed on a topic the registry sends the last known notification for each cluster node. These last known notifications are part of the polling mechanism and will be used to filter the response so that only unknown notifications are sent to the UI and will be updated each time the poller processes a notification.
If a notification is added to the registry as a result of an event that is synced over the cluster the notification needs to be added to the registry using the clusterSync-flag set to true, and it must only be added on the cluster node that initially created the event. If the notification is added on each node using clusterSync=false, there are different notifications on each node and all of these notifications are only known to one node. In this case the response for the subscription or the poller will only contain one of these notifications. So if the poller reaches any other node in the future it will consider these notifications as unknown and therefore, processes them resulting in the listener being executed multiple times for each original event. In addition, the time when the listener is executed is no longer deterministic for the subsequent executions as this depends on the poller being routed to different cluster nodes which may happen at any time.
In order to fix this issue, all ui notifications added to the registry as a result of a cluster sync event are add using clusterSync=true but only on the node the event initially occurs.

470605


paolobazzi (approved):
Looks good to me.

paolobazzi (approved):

PR #2279, opened 2026-07-29
Restrict client notification collector to service tunnel

rogersteblerbsi:
REST invocations to a Scout backend are also executed within a
ServerRunContext but do not have the possibility to include client
notification as piggyback in the invocation response. Before this
change client notifications added piggyback during a REST invocation
where later discarded and not dispatched to the calling UI node.

471219


paolobazzi (approved):
looks good, added unit test

PR #2282, opened 2026-08-03
Inject relevant style sheets into PopupWindow

fschinkel:
The PopupWindow may display content that depends on dynamically injected style sheets (e.g. by jquery-scout). These style sheets are not reused from the main window or injected dynamically by the browser. The PopupWindow now ensures that all relevant style sheets (marked with 'relevantForPopupWindow') are injected into the popup before any data is rendered. Additionally, all relevant style sheets that are injected into the main window while the popup is opened are injected into the popup window as well.

463702

### DIFF
--- a/eclipse-scout-core/src/jquery/jquery-scout.js
+++ b/eclipse-scout-core/src/jquery/jquery-scout.js
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2024 BSI Business Systems Integration AG
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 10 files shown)


eclipse-scout-core/src/jquery/jquery-scout.js
```diff
@@ -252,10 +253,17 @@ $.injectStyleSheet = (url, options) => {
...
+    for (const [key, value] of Object.entries(options.data)) {
+      $linkTag.data(key, value);
+    }
+  }
+
   // Use raw JS function to append the <script> tag, because jQuery handles
   // script tags specially (see "domManip" function) and uses eval() which
   // is not CSP-safe.
   myDocument.head.appendChild(linkTag);
+  $(myDocument.head).trigger('injectStyleSheet', {$linkTag});
```
cguglielmo: Does $linkTag need to be wrapped in an object?

cguglielmo (approved):

PR #2284, opened 2026-08-03
PageWithTable: align loading indicators in a Scout Classic application

fschinkel:
A PageWithTable sets its table loading while the data is loaded. This leads to a loading indicator being displayed in the table while the rest of the application, e.g. the search form, is still enabled. The behavior differs from a Scout Classic IPageWithTable, where the table is not loading but the desktop is busy and shows a modal busy indicator. To not confuse users in Scout Classic applications having Scout JS PageWithTable, this behavior is aligned.
A new property asyncLoading was introduced on the Scout JS table which determines whether the table shows a loading indicator or sets the desktop busy while loading its data. By default, this property is set to true, which means a loading indicator is shown.
The OutlineAdapter disables asyncLoading on all tables that belong to a PageWithTable, which means the loading indicator is not rendered while the data is loaded. Instead, the desktop is set busy during this time.

471868

### DIFF
--- a/eclipse-scout-core/src/table/TableLoadingSupport.ts
+++ b/eclipse-scout-core/src/table/TableLoadingSupport.ts
@@ -1,0 +1,80 @@
+/*
+ * Copyright (c) 2010, 2026 BSI Business Systems Integration AG
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+import {LoadingSupport, LoadingSupportOptions, PropertyChangeEvent, Table} from '../index';
+
+export class TableLoadingSupport extends LoadingSupport {
+
+  declare widget: Table;
+
+  protected _busy = false;
+  protected _tableAsyncLoadingChangeHandler = this._onTableAsyncLoadingChange.bind(this);
+  protected _tableRemoveHandler = this._onTableRemove.bind(this);
+
+  constructor(options: LoadingSupportOptions) {
+    super(options);
+
+    this.widget.on('propertyChange:asyncLoading', this._tableAsyncLoadingChangeHandler);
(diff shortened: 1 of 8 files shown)


eclipse-scout-core/src/table/TableLoadingSupport.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  protected _onTableAsyncLoadingChange(event: PropertyChangeEvent<boolean>) {
+    // remove loading indicator if there is one and render a new one
+    this._removeLoadingIndicator();
+    this.renderLoading(true);
+  }
+
+  protected _onTableRemove() {
+    // remove busy indicator, this can not be done in remove() as it is also asynchronously triggered by __removeLoadingIndicator()
```
cguglielmo:
```suggestion
    // remove busy indicator, this can not be done in remove() as it is also asynchronously triggered by _removeLoadingIndicator()
```

eclipse-scout-core/src/table/TableModel.ts
```diff
@@ -318,4 +318,12 @@ export interface TableModel extends WidgetModel {
    * @see TableRow.dropTypes
    */
   rowsDraggable?: boolean;
+  /**
+   * Specifies whether the table loads its data asynchronously.
```
cguglielmo, line 322: Add empty line after first sentence to great a paragraph.

eclipse-scout-core/src/table/TableUpdateBuffer.ts
```diff
@@ -28,12 +28,12 @@ export class TableUpdateBuffer extends EventEmitter {
    * The buffer is active if it contains at least one promise. When all promises resolve the buffer will be processed.
    */
   pushPromise(promise: JQuery.Promise<any>) {
+    this._setTableLoading(true);
```
cguglielmo: Is the change in this class still necessary?

cguglielmo (commented):
Please add a spec for the asyncLoading property.

cguglielmo (approved):

PR #2291, opened 2026-08-12
SearchOutline: retrigger search when adding new pages

bschwarzent:
When a new page is added to the SearchOutline with a pending search state and the user previously entered a valid search query, the search is now automatically retriggered to populate the newly inserted page. Otherwise, populated and unpopulated pages would be mixed until the search query is changed by the user. This also coverts the case where the entire outline is reloaded.

473062

### DIFF
--- a/eclipse-scout-core/src/desktop/outline/SearchOutline.ts
+++ b/eclipse-scout-core/src/desktop/outline/SearchOutline.ts
@@ -336,6 +336,12 @@ export class SearchOutline extends Outline implements SearchOutlineModel {
     });
 
     this._updateSearchStatus();
+
+    // If a valid search query is present and a new search state was added (e.g. new page inserted)
+    // retrigger the search. This also covers the case where the entire outline is reloaded with F5.
+    if (this._searchQueryValid && searchStatesToAdd.size) {
+      this.search();
+    }
   }
 
   protected _installSearchStateListeners(searchState: SearchState) {
--- a/eclipse-scout-core/test/desktop/outline/SearchOutlineSpec.ts
+++ b/eclipse-scout-core/test/desktop/outline/SearchOutlineSpec.ts
@@ -7,7 +7,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  */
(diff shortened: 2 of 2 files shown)


eclipse-scout-core/src/desktop/outline/SearchOutline.ts
```diff
@@ -336,6 +336,12 @@ export class SearchOutline extends Outline implements SearchOutlineModel {
     });
 
     this._updateSearchStatus();
+
+    // If a search query is present and a new, unpopulated search state was added (e.g. new page inserted)
+    // retrigger the search. This also covers the case where the entire outline is reloaded with F5.
+    if (this.searchQuery && [...searchStatesToAdd].some(searchState => searchState.pending)) {
```
fschinkel: Is the check for `this.searchQuery` necessary? `search()` will validate the query.
If a new page is added containing a search result that does not match the searchQuery, it needs to be reset. I think it is sufficient to only check if there are any search states added.

eclipse-scout-core/test/desktop/outline/SearchOutlineSpec.ts
```diff
@@ -39,6 +39,84 @@ describe('SearchOutline', () => {
 
       expect(outline._searchStates).toEqual(new Set([searchPage0.searchState, searchPage1.searchState, searchPage2.searchState]));
     });
+
+    it('does not retrigger search when search query is absent', () => {
+      const outline = createSearchOutline();
```
fschinkel, line 44: This spec does not check if search was triggered or do I miss something? Maybe add a spy for `triggerSearch` or a listener for `search`?

eclipse-scout-core/src/desktop/outline/SearchOutline.ts
bschwarzent: Changed to this._searchQueryValid. Specs adjusted.

eclipse-scout-core/test/desktop/outline/SearchOutlineSpec.ts
bschwarzent: Added event collector.

fschinkel (approved):

PR #2309, opened 2026-08-21
Add RemoveChildPagesHybridAction

cguglielmo:
The action is necessary if a js page uses searchRequired. In that case PageWithTable._resetTableData has to remove the remote child pages first before any table row can be deleted. Why? Because deleting the rows will also delete the child pages, but the ui is not allowed to delete pages managed by the server. There may still be a response pending containing events for these remote pages which would result in an error if they cannot be found.

474004

### DIFF
--- a/eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
+++ b/eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
@@ -219,7 +219,7 @@ export class JsPageHelper implements JsPageHelperModel, ObjectWithType {
   }
 
   /**
-   * Calls loadChildPages on the UI server for the given ids or {@link PageParamDo}s. The pages are created by the Java equivalent of the JsPage (see AbstractJsPage.createChildPage).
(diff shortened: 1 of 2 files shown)


eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
```diff
@@ -269,6 +269,22 @@ export class JsPageHelper implements JsPageHelperModel, ObjectWithType {
     this._addChildPagesToIdMap();
   }
 
+  /**
+   * Calls `removeChildPages` on the UI server.
+   * The returned {@link Promise} is resolved when all child pages are removed on the UI server.
+   */
+  async callRemoveChildPages(): Promise<void> {
+    if (this.page.leaf) {
```
bschwarzent, line 277: Can we also skip the hybrid action if `page.childNodes` is empty?

bschwarzent (approved):

eclipse-scout-core/src/desktop/outline/pages/js/JsPageHelper.ts
cguglielmo: yes, should be possible

PR #2317, opened 2026-08-26
TypeParameterBeanRegistry: fix race condition on bean unregister

rogersteblerbsi:
Synchronize bean removal performed by the registration handle using the registry write lock.

Without the lock, concurrent dispose() and getBeans() calls could corrupt the underlying IndexedStore and lead to intermittent failures.

474166


work-lto (approved):
Thank you for fixing :)

PR #2322, opened 2026-08-31
Property support: Use static unmodifiable Maps for common defaults

matthiaso:
Goal: Decrease memory usage when lots of property supports are in use.

449281


org.eclipse.scout.rt.platform/src/main/java/org/eclipse/scout/rt/platform/reflect/DefaultValueMap.java
```diff
@@ -0,0 +1,309 @@
...
+   *     potential default values for this {@link Map}
+   * @param startEmpty
+   *     whether {@link Map} is initially empty (true) or should be pre-filled with the default values
+   */
+  public DefaultValueMap(Map<K, V> defaultValues, boolean startEmpty) {
+    m_defaultValues = defaultValues;
+    m_values = new HashMap<>();
+
+    if (startEmpty) {
+      // mark all default values as removed in the initial values
```
paolobazzi:
```suggestion
      // mark all default values as removed in the values map
```

paolobazzi (approved):
looks good!

PR #2325, opened 2026-09-02
Application: Add HTTP connection idle timeout, improve stream/async TO

paolobazzi:
- Added new low-level HTTP connection timeout for HTTP/HTTPS default 6m
- Improved default value for HTTP2 stream timeout, new value 6 min
- Improved default value for HTTP async context timeout, new value 6 min

475330


work-lto (approved):
Local TEST OK

Before patch i ran into 503, with patch i didn't

PR #2331, opened 2026-09-08
ChunkedDataWriter: fetch and iterate within the same transaction

andibur:
Chunked data transfer is used for large amounts of data, such as those loaded from a database. In this case, it is absolutely necessary that the execution of the SQL statement and the iteration over the ResultSet take place within the same transaction.

This change modifies the signature of toResponse from an Iterator<T> to a Supplier<Iterator<T>>, allowing to create and iterate in the same RunContext.

474697

### DIFF
--- a/org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/chunked/IChunkedDataWriter.java
+++ b/org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/chunked/IChunkedDataWriter.java
@@ -12,12 +12,16 @@ package org.eclipse.scout.rt.rest.chunked;
 import java.io.Closeable;
 import java.io.IOException;
 import java.util.Iterator;
+import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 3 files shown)


org.eclipse.scout.rt.rest/src/main/java/org/eclipse/scout/rt/rest/chunked/IChunkedDataWriter.java
```diff
@@ -104,4 +110,54 @@ default Response toResponse(Iterator<T> iterator) {
     });
     return Response.ok(toEntity()).build();
   }
+
+  /**
+   * Schedules job that invokes given supplier and that writes its values, and returns a response created by {@link #toEntity()}.
+   * The iterator supplier as well as the iteration loop are performed within the same job in order to use
+   * the same transaction. This method blocks until the supplier completes and re-throes any unchecked exception.
```
ralph-steiner: typo 're-throes'

ralph-steiner (approved):

PR #2332, opened 2026-09-08
REST: Add logLevel to ErrorDo

rogersteblerbsi:
Add an optional logLevel property to REST error responses.

This allows server-side code to explicitly control how errors are logged on the client instead of relying solely on the severity-to-log-level mapping.

Veto exceptions are now reported with log level INFO.

474040


rogersteblerbsi:
PR 26.2: https://github.com/eclipse-scout/scout.rt/pull/2312

PR #2292, opened 2026-08-13
Add per-cache statistics for InvalidateCacheNotification

rogersteblerbsi:
Store cluster synchronization status information using a message type and optional identifier key. For cache invalidation messages, the cache ID is used as identifier, allowing status information to be queried per cache via getInvalidationStatusInfo(String).

214038

PR #2333, opened 2026-09-08
Add per-cache statistics for InvalidateCacheNotification

rogersteblerbsi:
Store cluster synchronization status information using a message type and optional identifier key. For cache invalidation messages, the cache ID is used as identifier, allowing status information to be queried per cache via getInvalidationStatusInfo(String).

214038


rogersteblerbsi:
PR 26.2: https://github.com/eclipse-scout/scout.rt/pull/2292

PR #2340, opened 2026-09-10
Password policy: Increase default min password length to 14

rogersteblerbsi:
476392

### DIFF
--- a/org.eclipse.scout.rt.server.test/src/test/java/org/eclipse/scout/rt/server/services/common/pwd/DefaultPasswordPolicyTest.java
+++ b/org.eclipse.scout.rt.server.test/src/test/java/org/eclipse/scout/rt/server/services/common/pwd/DefaultPasswordPolicyTest.java
@@ -9,24 +9,18 @@
  */
 package org.eclipse.scout.rt.server.services.common.pwd;
 
-import org.eclipse.scout.rt.platform.BEANS;
-import org.eclipse.scout.rt.platform.IBean;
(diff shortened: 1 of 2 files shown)


org.eclipse.scout.rt.server.test/src/test/java/org/eclipse/scout/rt/server/services/common/pwd/DefaultPasswordPolicyTest.java
```diff
@@ -9,24 +9,18 @@
...
 public class DefaultPasswordPolicyTest {
 
-  protected IBean<?> m_minLengthPropertyMockBean;
-
-  @After
-  public void after() {
-    BEANS.get(BeanTestingHelper.class).unregisterBean(m_minLengthPropertyMockBean);
-  }
+  @Rule
+  public MockConfigPropertyRule<Integer> m_passwordPolicyMinLengthPropertyRule = new MockConfigPropertyRule<>(PasswordPolicyMinLengthProperty.class, 12);
```
work-lto, line 23: Why 12?

work-lto: Talked in person and the reasoning is twofold: 1) The test should work for whatever default is set 2) it's more pragmatic to not change the entire test

work-lto (approved):

PR #2357, opened 2026-09-21
OWASP dependency check

paolobazzi:
453821


paolobazzi:
Review see https://github.com/eclipse-scout/scout.rt/pull/2353/

PR #2358, opened 2026-09-21
DoNode: ensure hashCode consistency for array values

rogersteblerbsi:
Add ObjectUtility.hashCode() and use it in DoNode to keep hashCode() consistent with the existing array-aware equals() implementation.

476611


paolobazzi (approved):
looks good, added some additional tests as separate (AI-assisted) commit

PR #2361, opened 2026-09-22
DoNode: ensure hashCode consistency for array values

rogersteblerbsi:
Add ObjectUtility.hashCode() and use it in DoNode to keep hashCode() consistent with the existing array-aware equals() implementation.

476611


rogersteblerbsi:
Cherry-Pick of https://github.com/eclipse-scout/scout.rt/pull/2358

